PR #5, opened 2019-12-04
(DOCSP-7599): find multiple documents usage example

nathan-contino-mongo:
JIRA: https://jira.mongodb.org/browse/DOCSP-7599

Staging: https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker/DOCSP-7599/usage-examples/find


source/code-snippets/usage-examples/find.js
```diff
@@ -0,0 +1,28 @@
...
+async function run() {
+   try {
+      await client.connect();
+
+      const database = client.db('sample_mflix');
+      const collection = database.collection('movies');
+
+      // create a query document for a range match for a movie less than 15 minutes long
+      const query = { runtime: { $lt : 15 } };
+      const cursor = collection.find(query);
```
mongomoe: should this be: `await collection.find(query);`

nathan-contino-mongo: It works either way -- I suspect since the cursor is lazy, the fact that we use an await below makes it unnecessary. I'll add an await here though because I suspect we're "better safe than sorry" when it comes to new users who might play around with/edit these examples to better understand what's going on (and then use the resulting code in personal projects.

source/usage-examples/find.txt
```diff
@@ -0,0 +1,143 @@
...
+
+.. code-block:: javascript
+
+   { "_id": 4, "name": "avocados", "qty": 3, "rating": 5 }
+   { "_id": 1, "name": "apples", "qty": 5, "rating": 3 }
+   { "_id": 3, "name": "oranges", "qty": 6, "rating": 2 }
+   { "_id": 2, "name": "bananas", "qty": 7, "rating": 1 }
+
+
+If we rewrote this sort to use a value of
```
terakilobyte: Active voice and simplification. As an example `Rewriting this sort to use a value of **1** instead of **-1** will sort in ascending order.`

source/usage-examples/find.txt
```diff
@@ -0,0 +1,143 @@
...
+``1`` for the ``rating`` key instead of a ``-1``, ``find`` would sort
+our fruit in *ascending* order by rating, and the lowest rated fruit
+would come first. If you do not specify a sort order, documents are
+returned in
+`natural sort order <https://docs.mongodb.com/manual/reference/glossary/#term-natural-order>`_.
+
+Projection
```
terakilobyte: Specify Which Fields to Return may be more naturally descriptive.

source/usage-examples/find.txt
```diff
@@ -0,0 +1,143 @@
...
+=======================
+
+.. default-domain:: mongodb
+
+Overview
+--------
+
+`collection.find() <https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#find>`_
+lets you query for multiple documents in MongoDB. You can pass ``find``
+an (optional) query document and an options object. Use the options
```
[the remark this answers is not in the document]
nathan-contino-mongo: Agreed, especially since the purpose of this is sort of a jumping-off point for new users/users who have forgotten the concept but may have learned it before.

source/usage-examples/find.txt
```diff
@@ -0,0 +1,143 @@
...
+object to specify a sort order and a projection, both of which
+are optional. Consult the `collection.find() reference documentation <https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#find>`_
```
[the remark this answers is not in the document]
nathan-contino-mongo: Not really, but I felt it was appropriate to link explicitly to the reference docs once the basics were explained. I'm going to convert the first link to simple monospace instead for this reason.

source/code-snippets/usage-examples/findOne.js
```diff
@@ -13,7 +13,7 @@ async function run() {
       const database = client.db('sample_mflix');
       const collection = database.collection('movies');
 
-      // create a query document to look up an exact match for a move with this title
+      // create a query document to look up an exact match for a movie with this title
```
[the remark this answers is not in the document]
nathan-contino-mongo: Changing from ``an`` -> ``any`` to make this more explicit

source/usage-examples/find.txt
```diff
@@ -0,0 +1,143 @@
...
+to learn about other things you can accomplish with options and
+``find``.
+
+Query Document
+~~~~~~~~~~~~~~
+
+You can use the query document to filter the collection down from all
```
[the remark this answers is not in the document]
nathan-contino-mongo: Clarifying this line -- I think it's important to give the visual of "filtering" in the sense that this is how mongo decides which documents to return, but I don't have the full sentence figured out yet. Tune in after lunch for the exciting conclusion.

source/usage-examples/find.txt
```diff
@@ -0,0 +1,143 @@
...
+   const query = { qty: { $gt : 5 } };
+
+``find`` returns the following documents for such a query:
+
+.. code-block:: javascript
+
+   { "_id": 2, "name": "bananas", "qty": 7, "rating": 1 }
+   { "_id": 3, "name": "oranges", "qty": 6, "rating": 2 }
+
+Sort Order
```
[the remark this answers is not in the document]
nathan-contino-mongo: I don't think I see a big advantage on either side here. I like that the titles here actually match the terms I used above, so it's easier to tell what these sections are describing... but I see how turning the titles into an action give a sense of what this term actually does in ``find``. I'm going to leave it as is but if you feel strongly I can change it.

source/usage-examples/find.txt
```diff
@@ -0,0 +1,143 @@
...
+
+`collection.find() <https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#find>`_
+lets you query for multiple documents in MongoDB. You can pass ``find``
+an (optional) query document and an options object. Use the options
+object to specify a sort order and a projection, both of which
+are optional. Consult the `collection.find() reference documentation <https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#find>`_
+to learn about other things you can accomplish with options and
+``find``.
+
+Query Document
```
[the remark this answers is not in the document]
nathan-contino-mongo: Same as sort comment

source/usage-examples/find.txt
```diff
@@ -0,0 +1,143 @@
...
+.. code-block:: javascript
+
+   { "_id": 2, "name": "bananas", "qty": 7, "rating": 1 }
+   { "_id": 3, "name": "oranges", "qty": 6, "rating": 2 }
+
+Sort Order
+~~~~~~~~~~
+
+You can use the sort order to change the order in which documents are
+returned by ``find``. You could pass the following sort order to
```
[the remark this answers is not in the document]
nathan-contino-mongo: Agreed, changing.

source/usage-examples/find.txt
```diff
@@ -0,0 +1,143 @@
...
+You could pass the following query to the ``find`` method to search for
+only fruit documents listing a quantity of greater than 5:
+
+.. code-block:: javascript
+
+   const query = { qty: { $gt : 5 } };
+
+``find`` returns the following documents for such a query:
+
+.. code-block:: javascript
```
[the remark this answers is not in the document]
nathan-contino-mongo: Adding this to each example, in addition to a couple lines building the options object. If you can think of a smoother way to demonstrate this, I'm happy to polish it.

Agreed that this builds familiarity with how you actually construct a ``find`` call and the ``options`` object.

source/usage-examples/find.txt
```diff
@@ -0,0 +1,143 @@
...
+
+   const sort = { rating: -1 }
+
+In this case, the number ``-1`` tells ``find`` to sort the fruit in
+descending order by rating, so the highest rated fruit comes first and
+the lowest rated fruit comes last. ``find`` returns the following
+documents when this sort is used with no query:
+
+.. code-block:: javascript
+
```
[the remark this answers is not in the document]
nathan-contino-mongo: added

source/usage-examples/find.txt
nathan-contino-mongo: same as sort/query comments

source/usage-examples/find.txt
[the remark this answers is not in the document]
nlarew: My inclination is to use verb phrases here (i.e. N[0]'s suggestion). To me, noun headings convey "conceptual/reference" content and verb headings convey "guided/use case"  content.

We want these examples to be a guide, not an exhaustive reference, so I'd go with verb phrases here. Separately, we probably _should_ have reference content elsewhere that more exhaustively defines each concept and have noun headings.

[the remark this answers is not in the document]
nlarew: One thing we discussed in the office today is that these sub-sections could also be first class example pages themselves. Given that parallelism, it seems even more appropriate to use verb phrases here.

source/usage-examples/find.txt
```diff
@@ -0,0 +1,164 @@
...
+   options = {};
+   options.sort = sort;
```
nlarew: - Should have a `const`
- Can this just be one statement? i.e.
```js
const options = {
  sort: { rating: -1 }
}
```

source/usage-examples/find.txt
```diff
@@ -0,0 +1,164 @@
...
+returned documents. Pass the following projection to only return the
+names of the fruits:
+
+.. code-block:: javascript
+
+   const projection = { name: 1 }
+
+``name`` has a value of ``1`` here to indicate that ``find`` should
+*include* the ``name`` field of each document returned. This implicitly
+*excludes* any fields that aren't explicitly included, so remember to
```
nlarew: included _in the projection_

source/usage-examples/find.txt
```diff
@@ -0,0 +1,164 @@
...
+``_id`` field is a unique identifier for each document, and if you
+want to refer to these documents again, you'd create a query document
+matching by ``_id`` in case there are duplicate ``name`` values in your
+collection. MongoDB anticipates that you might want a unique identifier
+for any documents you've found, so ``_id`` is always included unless you
+explicitly exclude it.
+
+You can explicitly *exclude* fields in your projections using a
+value of ``0``. This works just like include: if you explicitly exclude,
+all other fields are implicitly *included*.
```
nlarew: Probably should note that you can't mix these two approaches

source/usage-examples/find.txt
```diff
@@ -0,0 +1,164 @@
...
+   options.projection = projection;
+   const cursor = collection.find({}, options);
+
+   { "_id": 1, "name": "apples" }
+   { "_id": 2, "name": "bananas" }
+   { "_id": 3, "name": "oranges" }
+   { "_id": 4, "name": "avocados" }
+
+Despite the fact that this query only explicitly included the ``name``
+field, the ``_id`` field is returned as well! This happens because the
```
nlarew: the query returns...

source/usage-examples/find.txt
```diff
@@ -0,0 +1,164 @@
...
+   options = {};
+   options.projection = projection;
+   const cursor = collection.find({}, options);
+
+   { "_id": 1, "name": "apples" }
+   { "_id": 2, "name": "bananas" }
+   { "_id": 3, "name": "oranges" }
+   { "_id": 4, "name": "avocados" }
+
+Despite the fact that this query only explicitly included the ``name``
```
nlarew: this query -> the projection

source/usage-examples/find.txt
```diff
@@ -0,0 +1,164 @@
...
+field, the ``_id`` field is returned as well! This happens because the
+``_id`` field is a unique identifier for each document, and if you
+want to refer to these documents again, you'd create a query document
+matching by ``_id`` in case there are duplicate ``name`` values in your
+collection. MongoDB anticipates that you might want a unique identifier
+for any documents you've found, so ``_id`` is always included unless you
+explicitly exclude it.
```
nlarew: This paragraph is a little confusing. I'd emphasize first that `_id` is a special case and then give context as to why.

nlarew: i.e. this happens because "projections treat the `_id` field differently than other fields", not just because it's a unique ID.

source/usage-examples/find.txt
```diff
@@ -0,0 +1,164 @@
...
+all other fields are implicitly *included*.
+
+Return Value
+~~~~~~~~~~~~
+
+``find`` returns a
+`Cursor <https://mongodb.github.io/node-mongodb-native/3.3/api/Cursor.html>`_
+that contains the results of your query. If no document matches the
+query, the returned Cursor contains an empty collection of
+documents.
```
nlarew: **Return Value** is reference-oriented and doesn't really map to a use case. Maybe we could just mention this in the general description of how find works in the intro section?

source/usage-examples/find.txt
```diff
@@ -0,0 +1,164 @@
...
+
+Example
+-------
+
+The following snippet finds documents from the ``movies``
+collection. It uses the following:
+
+- A **query document** that filters returned documents down to only
+  movies with a runtime of less than 15 minutes.
```
nlarew: - It doesn't filter the returned documents, it determines which documents the database includes in the result set. e.g. "A **query document** that configures the query to return only movies with a runtime of less than 15 minutes."

source/usage-examples/find.txt
```diff
@@ -0,0 +1,164 @@
...
+
+- A **sort order** that organizes returned documents by ascending title
+  (A->Z)
```
nlarew: - Maybe just **sort** rather than **sort order**?
- "by ascending title" -> "in ascending order by title"

source/usage-examples/find.txt
```diff
@@ -0,0 +1,164 @@
...
+
+- A **projection** that explicitly excludes the ``_id`` field from
+  returned documents and explicitly includes only the title and imdb
+  object, which contains ratings, votes, and an id.
```
nlarew: Monospace field names

source/usage-examples/find.txt
```diff
@@ -0,0 +1,164 @@
...
+=======================
+
+.. default-domain:: mongodb
+
+Overview
+--------
+
+``collection.find()`` lets you
+`query <https://docs.mongodb.com/manual/tutorial/query-documents/#read-operations-query-argument>`_
+for multiple documents in MongoDB. You can pass ``find``
```
nlarew: nit: find is a function, so I'd notate it as `find()` rather than `find`

source/code-snippets/usage-examples/find.js
```diff
@@ -0,0 +1,41 @@
...
+);
+
+async function run() {
+   try {
+      await client.connect();
+
+      const database = client.db('sample_mflix');
+      const collection = database.collection('movies');
+
+      // create a query document for a range match for a movie less than 15 minutes long
```
nlarew: "range match" seems a little jargon-y IMO. Maybe:

`// Query for movies that have a runtime less than 15 minutes`

source/code-snippets/usage-examples/find.js
```diff
@@ -0,0 +1,41 @@
...
+      // initialize an empty options object
+      options = {}
+
+      // sort our returned documents by ascending title
+      const sort = { title: 1 };
+
+      // returned documents should only contain title and the imdb object, no _id
+      const projection = { title : 1, _id: 0, imdb: 1 }
+      options.sort = sort;
+      options.projection = projection;
```
nlarew: I think we could condense this a bit without losing clarity if we combine all the definitions into the `options` declaration.
```js
const options = {
  // Sort the returned documents in ascending order by title
  sort: { title: 1 },
  // Include only the `title` and `imdb` fields in each returned document
  projection: { _id: 0, title: 1, imdb: 1 }
}
```

source/code-snippets/usage-examples/find.js
```diff
@@ -0,0 +1,41 @@
...
+
+      // pass our query and constructed options
+      const cursor = collection.find(query, options);
+
+      // iterate with await so the session doesn't end while printing
+      for await (const movie of cursor) {
+         console.dir(movie);
+      }
```
nlarew: What if the query does not match any documents or throws an error?

nlarew: Why do we need to iterate with `await`? (Because the Cursor is just a pointer to documents in the result set, so we need to actually go and fetch each one)

source/code-snippets/usage-examples/findOne.js
```diff
@@ -13,7 +13,7 @@ async function run() {
       const database = client.db('sample_mflix');
       const collection = database.collection('movies');
 
-      // create a query document to look up an exact match for a move with this title
+      // create a query document to look up any exact match for a movie with this title
```
nlarew: "exact match" is a bit confusing because I could interpret that to mean that the document is literally identical to the query (no additional fields, etc). Maybe:

`// Query for a movie that has the title "The Room"`

source/usage-examples/find.txt
nlarew: I could see readers misinterpreting "the cursor contains" to imply that the cursor object keeps the result set in memory. This isn't true - a cursor is a reference pointer to a documents in the result set, but doesn't itself contain the data.

You need to iterate through the cursor using a cursor method like `next()`, `toArray()`, or `forEach()` to actually fetch the documents.

source/usage-examples/find.txt
```diff
@@ -0,0 +1,164 @@
+=======================
+Find Multiple Documents
+=======================
+
+.. default-domain:: mongodb
+
+Overview
+--------
+
+``collection.find()`` lets you
```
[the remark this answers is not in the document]
nathan-contino-mongo: I think I prefer it this way (see my conversation with Nathan). Trouble is that there are a LOT of options and only one place to link to, and I'd like to give users a brief summary of what's going on before I throw them at the reference material. I really like the "call to action" style at the end, too.

While I would normally place the link at the beginning, links like that are largely a waste. I could link in both places if you feel really strongly that this should be a link.

source/usage-examples/find.txt
nathan-contino-mongo: The reference material certainly makes it seem that way, but you can always exclude ``_id`` so... I'm really not sure what I can add here that won't require at least one caveat and potentially confuse readers. Thoughts?

source/usage-examples/find.txt
nathan-contino-mongo: also the title of my MongoDB movie sequel

source/usage-examples/find.txt
nathan-contino-mongo: Fair, I was trying to do the opposite: point out why _id is especially useful, then mention that mongodb treats it differently as a result. Since it sounds like mine can be a bit confusing, I'll change mine to the opposite style like you suggest to make it clearer.

source/usage-examples/find.txt
nathan-contino-mongo: I don't have a strong opinion on this, but ``find`` occurs a lot on this page, as you might expect. I think it feels a little disruptive to see the extra parens, but that's definitely a personal feeling. Thoughts, @nlarew ?

source/code-snippets/usage-examples/find.js
nathan-contino-mongo: - if the query does not match any documents, it doesn't fail but it also iterates over zero items, so nothing really happens. Not sure if there's a better way (maybe check length, if 0 print "no matches found?")
- if the query throws an error, it's caught in the catch() at the bottom of the file
- i agree with your last point, but it's a little long to include in a comment. I think I'd prefer to explain that part of the Cursor object in the text and then keep it simple (this plus a little reference to that?) in the comments on the code itself. Thoughts? @nlarew

source/usage-examples/find.txt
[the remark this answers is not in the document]
nathan-contino-mongo: Agreed. This makes a lot more sense when you consider that one day there will be a more comprehensive reference page for this too.

source/usage-examples/find.txt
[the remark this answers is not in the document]
nathan-contino-mongo: Thoughts, @nlarew ?

source/usage-examples/find.txt
nathan-contino-mongo: Agreed -- removed the "contains" reference, added notes about iteration to actually get contents, and moved to the intro section.

PR #8, opened 2019-12-09
TEMPORARY - DO NOT REVIEW: Docsp 7603 usage example update a doc

mongomoe:
**Hey, potential reviewers of this. Currently, putting this PR in a 'do not review' state while I address the other PR comments on this PR. If you have some thoughts on this PR feel free to reach out to me on Slack or via email.** 

Jira
-----
* https://jira.mongodb.org/browse/DOCSP-7603

Staging
--------
* https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker/DOCSP-7603-usage-example-update-a-doc/usage-examples/updateOne
* https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker/DOCSP-7603-usage-example-update-a-doc/usage-examples/findOneAndUpdate

### DIFF
--- a/source/code-snippets/usage-examples/updateOne.js
+++ b/source/code-snippets/usage-examples/updateOne.js
@@ -1,0 +1,37 @@
+// ignored first line
+const { MongoClient } = require('mongodb');
+
+const client = new MongoClient(
+  'mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&w=majority',
+);
+
+async function run() {
+  try {
+    await client.connect();
+
+    const database = client.db('sample_mflix');
+    const collection = database.collection('movies');
+
+    // create a query for a movie to update
+    const filter = { title: 'Blacksmith Scene' };
+    const options = {};
+    options.upsert = true; // create a document if no documents match the filter
+    options.w = 1; // request acknowledgment from MongoDB when the update has been performed on the standalone mongod instance
+
+    const result = await collection.updateOne(
+      filter,
+      {
+        $set: {
+          plot: 'Blacksmith Scene is a silent film directed by William K.L. Dickson',
+        },
+      },
+      options,
+    );
+    console.log(
+      `${result.matchedCount} document matched the query and ${result.modifiedCount} document was updated`,
+    );
+  } finally {
+    await client.close();
+  }
+}
+run().catch(console.dir);
--- a/source/usage-examples/findOneAndUpdate.txt
+++ b/source/usage-examples/findOneAndUpdate.txt
@@ -8,8 +8,7 @@ A single document can be updated and returned in one single atomic
 operation using the `findOneAndUpdate()
 <https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#findOneAndUpdate>`_
 method.  ``findOneAndUpdate()`` takes a filter object, updating the
-first document that matches it. An empty filter updates the first
-document in a collection. 
+first document that matches it. 
 
 .. note::
     While ``findOneAndUpdate()`` is in execution, any attempts to modify
--- a/source/usage-examples/findOneAndUpdate.txt
+++ b/source/usage-examples/findOneAndUpdate.txt
@@ -31,6 +31,11 @@ is specified in the Options object, and info about the success of the
 operation and the ``operationTime``.
 
 
+Specifying incorrect options can cause errors to be thrown. Defining
+both fields to include and to exclude ( ``{ field: 1, otherField: 0 }``)
+will cause the error: "MongoError: Projection cannot have a mix of
+inclusion and exclusion."
+
 .. literalinclude:: /code-snippets/usage-examples/findOneAndUpdate.js
   :language: javascript
   :linenos:
--- a/source/usage-examples/findOneAndUpdate.txt
+++ b/source/usage-examples/findOneAndUpdate.txt
@@ -31,10 +31,10 @@ is specified in the Options object, and info about the success of the
 operation and the ``operationTime``.
 
 
-Specifying incorrect options can cause errors to be thrown. Defining
-both fields to include and to exclude ( ``{ field: 1, otherField: 0 }``)
-will cause the error: "MongoError: Projection cannot have a mix of
-inclusion and exclusion."
+Specifying incorrect options can cause errors to be thrown. For
+instance, defining both fields to include and to exclude ( ``{ field: 1,
+otherField: 0 }``) will cause the error: "MongoError: Projection cannot
+have a mix of inclusion and exclusion."
 
 .. literalinclude:: /code-snippets/usage-examples/findOneAndUpdate.js
   :language: javascript
--- a/source/usage-examples/findOneAndUpdate.txt
+++ b/source/usage-examples/findOneAndUpdate.txt
@@ -1,0 +1,36 @@
+============================
+Update a Document Atomically
+============================
+
+.. default-domain:: mongodb
+
+A single document can be updated and returned in one single atomic
+operation using the `findOneAndUpdate()
+<https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#findOneAndUpdate>`_
+method.  ``findOneAndUpdate()`` takes a filter object, updating the
+first document that matches it. An empty filter updates the first
+document in a collection. 
+
+.. note::
+    While ``findOneAndUpdate()`` is in execution, any attempts to modify
+    the same document by another client will have an error conflict. The
+    concurrent    operation to alter the document will automatically
+    rerun after the original ``findOneAndUpdate()`` is done. 
+
+Create an Options object to specify additional options. Specify the
+fields to return by providing specific fields to retrieve via the
+``projection`` key. Specify whether the filter should be ascending or
+descending to update a document based on a particular order by using the
+``sort`` key. The ``upsert`` option creates a new document if no
+documents match the filter. Specify to return the updated document by
+setting ``returnNewDocument`` to true. 
+
+The ``findOneAndUpdate()`` method returns an object containing the
+original document, or the updated document if ``returnNewDocument:true``
+is specified in the Options object, and info about the success of the
+operation and the ``operationTime``.
+
+
+.. literalinclude:: /code-snippets/usage-examples/findOneAndUpdate.js
+  :language: javascript
+  :linenos:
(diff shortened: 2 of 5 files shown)


source/usage-examples/findOneAndUpdate.txt
```diff
@@ -0,0 +1,42 @@
...
+documents match the filter. Specify to return the updated document by
+setting ``returnNewDocument`` to true. 
+
+The ``findOneAndUpdate()`` method returns an object containing the
+original document, or the updated document if ``returnNewDocument:true``
+is specified in the Options object, and info about the success of the
+operation and the ``operationTime``.
+
+Specifying incorrect parameters for your ``findOneAndUpdate()`` operation can
+cause errors to be thrown. One common error is the attempt to update an
```
terakilobyte: The jump in tone and voice here is dramatic. I wonder if this could be cleaned up with simplification? My take would be

Incorrect or malformed paramters will cause an error. Other errors include attempting to update the ``_id`` field or attempting to update a value that would cuase a unique constraint violation on a unique index.

source/usage-examples/findOneAndUpdate.txt
```diff
@@ -0,0 +1,42 @@
...
+immutable field, causing the error message: "Performing an update on the
+path '_id' would modify the immutable field '_id'". Also, attempting to
+update a field with a unique index would throw the error, "E11000
```
terakilobyte: Saying it in this manner could be interpreted as incorrect.

source/usage-examples/updateOne.txt
```diff
@@ -0,0 +1,32 @@
+=================
+Update a Document
+=================
+
+.. default-domain:: mongodb
+
+You can update a single document using the `updateOne() <https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#updateOne>`_ method. 
+``updateOne()`` takes a filter for a document, and updates the first document
```
terakilobyte: Perhaps we should use the same terminalogy across all of our methods in the docs. What do we say the other find methods in this position for the action performed?

source/code-snippets/usage-examples/updateOne.js
```diff
@@ -0,0 +1,37 @@
...
+
+async function run() {
+  try {
+    await client.connect();
+
+    const database = client.db('sample_mflix');
+    const collection = database.collection('movies');
+
+    // create a query for a movie to update
+    const filter = { title: 'Blacksmith Scene' };
```
terakilobyte: you are using `query` as the variable name in findOneAndUpdate

source/usage-examples/updateOne.txt
```diff
@@ -0,0 +1,32 @@
...
+
+Create an Options object to specify additional options. Specify the
+``upsert`` option to create a new document if no documents match the
+filter.  To receive acknowledgment in the response from MongoDB that the
+write operation was performed on the specified number of :manual:`mongod
+</reference/program/mongod>` servers, set the ``w`` option. 
+
+The ``updateOne()`` method returns a document containing information such
+as the number of documents that match the filter, and the number of
+documents that were modified.
```
terakilobyte: I hope only one documents was modified!

source/usage-examples/updateOne.txt
```diff
@@ -0,0 +1,32 @@
...
+
+Specifying incorrect parameters for your ``updateOne()`` operation can
+cause errors to be thrown. One common error is the attempt to update an
+immutable field, causing the error message: "Performing an update on the
+path '_id' would modify the immutable field '_id'". Also, attempting to
+update a field with a unique index would throw the error, "E11000
```
terakilobyte: Same comments as in the other method's text

source/usage-examples/findOneAndUpdate.txt
```diff
@@ -0,0 +1,42 @@
+============================
+Update a Document Atomically
+============================
+
+.. default-domain:: mongodb
+
+A single document can be updated and returned in one single atomic
```
nathan-contino-mongo: I would recommend passing over this whole document with active/passive voice in mind. Especially in an opener I'd probably start with "You can update and return a single document in an atomic operation with..." and then elaborate about what all of those concepts mean.

source/usage-examples/findOneAndUpdate.txt
```diff
@@ -0,0 +1,42 @@
...
+
+The ``findOneAndUpdate()`` method returns an object containing the
+original document, or the updated document if ``returnNewDocument:true``
+is specified in the Options object, and info about the success of the
+operation and the ``operationTime``.
+
+Specifying incorrect parameters for your ``findOneAndUpdate()`` operation can
+cause errors to be thrown. One common error is the attempt to update an
+immutable field, causing the error message: "Performing an update on the
+path '_id' would modify the immutable field '_id'". Also, attempting to
```
nathan-contino-mongo: Monospacing ``_id`` might make it more clear that we're talking about a field here.

source/usage-examples/findOneAndUpdate.txt
```diff
@@ -0,0 +1,42 @@
...
+update a field with a unique index would throw the error, "E11000
+duplicate key error collection: <dbName>.<collectionName> index:
+<indexField> dup key: { : <attempted update value> }"
```
nathan-contino-mongo: This error is a little hard to parse inline; do you think moving it to a code snippet might help?

source/usage-examples/updateOne.txt
```diff
@@ -0,0 +1,32 @@
...
+Update a Document
+=================
+
+.. default-domain:: mongodb
+
+You can update a single document using the `updateOne() <https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#updateOne>`_ method. 
+``updateOne()`` takes a filter for a document, and updates the first document
+that matches the filter.
+
+Create an Options object to specify additional options. Specify the
```
nathan-contino-mongo: This line feels tautological. Could we use it to introduce the names of the options we're going to discuss to give it a bit more purpose?

source/usage-examples/updateOne.txt
```diff
@@ -0,0 +1,32 @@
...
+``upsert`` option to create a new document if no documents match the
+filter.  To receive acknowledgment in the response from MongoDB that the
+write operation was performed on the specified number of :manual:`mongod
+</reference/program/mongod>` servers, set the ``w`` option.
```
nathan-contino-mongo: Reversing this line might make it easier to track -- I know the first half of the sentence was a bit confusing for me to parse.

source/usage-examples/updateOne.txt
```diff
@@ -0,0 +1,32 @@
...
+``updateOne()`` takes a filter for a document, and updates the first document
+that matches the filter.
+
+Create an Options object to specify additional options. Specify the
+``upsert`` option to create a new document if no documents match the
+filter.  To receive acknowledgment in the response from MongoDB that the
+write operation was performed on the specified number of :manual:`mongod
+</reference/program/mongod>` servers, set the ``w`` option. 
+
+The ``updateOne()`` method returns a document containing information such
```
nathan-contino-mongo: I wonder if it might be less confusing to use the term object here, even though this is a document, since using the same term to refer to the operation summary document and the items in the collection might be a little confusing for new users who don't have all the vocabulary down yet. Maybe an example document in a code snippet could help with this?

Also, I assume the method actually returns a Promise that resolves to the operation summary document since you use ``await`` in the code sample -- we should definitely mention that.

source/usage-examples/updateOne.txt
```diff
@@ -0,0 +1,32 @@
...
+as the number of documents that match the filter, and the number of
+documents that were modified.
+
+Specifying incorrect parameters for your ``updateOne()`` operation can
```
nathan-contino-mongo: This sentence sounds like we're going to discuss multiple error cases, but we only talk about one. I feel like we might just not need the first sentence here at all. What do you think?

source/usage-examples/findOneAndUpdate.txt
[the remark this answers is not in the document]
mongomoe: Can you suggest an alternative sentence? 
In my opinion, the suggested sentence ("A single document can be updated and returned in an atomic operation using the findOneAndUpdate() method.") goes against our anti-jargon philosophy of these usage examples because it uses a complex term (atomic) without providing an explanation as to what it is.

source/usage-examples/updateOne.txt
mongomoe: I think you're specifically referring to `query` vs `filter`. I agree that they need to be standard rather than synonyms. Defining which one to use is the challenge because:

* Both the node api docs and manual docs use 'filter' for both updateOne and findOneAndUpdate 
* Both the node api docs and manual docs use 'query' for both find and findOne


Sources:
---------
updateOne:
* https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#updateOne
* https://docs.mongodb.com/manual/reference/method/db.collection.updateOne/

findOneAndUpdate:
* https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#findOneAndUpdate
* https://docs.mongodb.com/manual/reference/method/db.collection.findOneAndUpdate/

find:
* https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#find
* https://docs.mongodb.com/manual/reference/method/db.collection.find/

findOne:
* https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#findOne
* https://docs.mongodb.com/manual/reference/method/db.collection.findOne/

source/usage-examples/updateOne.txt
mongomoe: Note: I'm making a new PR for this, but to address your comment, I'm removing the 'w' opt altogether, we don't want to talk about write concern in these docs.

source/usage-examples/updateOne.txt
mongomoe: Note: I'm making a new PR for this, but to address your comment:

I'm updating the new pr to specify that it returns a promise, and I've changed the wording to `object`

thanks

source/usage-examples/updateOne.txt
mongomoe: I disagree, the first line exists as an introduction to the paragraph that informs the user there can be user errors that occur. It's a bridge from the return paragraph to the error paragraph

nathan-contino-mongo: Good point -- a bridge is definitely necessary here! I think if we want to build a more comprehensive paragraph describing errors it might be a good idea to make sure that this bridge introduces all potential errors, and not just the subset of them that are caused by incorrect parameters (permissions errors, broken connection errors, timeouts, etc. all fall outside of this category). It might be tricky to do this without sounding too vague, but how about something like:
```
``updateOne()`` throws an error in several exceptional
circumstances. For instance...
```

Could even throw a bulleted list under there. Only thing I can't figure out right now is how to explain how the callback version of the method doesn't throw exceptions :)

PR #11, opened 2019-12-11
DOCSP-7602: adds bulkWrite usage example

schmalliso:
jira: https://jira.mongodb.org/browse/DOCSP-7602

staged: https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker/DOCSP-7602/usage-examples/bulkWrite

### DIFF
--- a/conf.py
+++ b/conf.py
@@ -61,7 +61,7 @@ rst_epilog = '\n'.join([
 pygments_style = 'sphinx'
 
 extlinks = {
-    'manual': ('http://docs.mongodb.com/manual%s', ''),
+    'manual': ('https://docs.mongodb.com/manual%s', ''),
 }
 
 ## add `extlinks` for each published version.
--- a/source/usage-examples/bulkWrite.txt
+++ b/source/usage-examples/bulkWrite.txt
@@ -43,7 +43,7 @@ documentation for full details.
   Promise resolves to a :node-api:`bulkWriteOpCallback </Collection.html#~bulkWriteOpCallback>` object containing the
   result of the operation.
 
-``collection.bulkWrite()`` returns a ``BulkWriteResult`` object
+``bulkWrite()`` returns a ``BulkWriteResult`` object
 that includes the number of updated or inserted documents, the ``_id`` of
 each document, and any ``writeError`` or ``writeConcernErrors``. Refer to
 the documentation for :node-api:`BulkWriteResult </BulkWriteResult>` for
@@ -69,8 +69,10 @@ Example
 
 The following code sample performs a bulk write operation against the
 ``users`` collection in the ``sample_mflix`` database. Specifically, the
-program formats the data from a ``users.json`` source file to perform
-a series of ``insertOne`` operations with the ``collection.bulkWrite()`` method. Download the dataset here: `users.json <https://raw.githubusercontent.com/mongodb-university/universal-driver-examples/master/users.json>`_
+program formats the data from a ``users.json`` source file to perform a
+series of ``insertOne`` operations with the ``bulkWrite()``collection
+method. Download the dataset here: `users.json
+<https://raw.githubusercontent.com/mongodb-university/universal-driver-examples/master/users.json>`_
 
 ``users.json`` is an array of objects that resemble the following:
 
--- a/source/usage-examples/bulkWrite.txt
+++ b/source/usage-examples/bulkWrite.txt
@@ -70,8 +70,7 @@ Example
 The following code sample performs a bulk write operation against the
 ``users`` collection in the ``sample_mflix`` database. Specifically, the
 program formats the data from a ``users.json`` source file to perform
-a series of ``insertOne`` operations with the :node-api:`collection.bulkWrite()
-<Collection.html#bulkWrite>` method. Download the dataset here: `users.json <https://raw.githubusercontent.com/mongodb-university/universal-driver-examples/master/users.json>`_
+a series of ``insertOne`` operations with the ``collection.bulkWrite()`` method. Download the dataset here: `users.json <https://raw.githubusercontent.com/mongodb-university/universal-driver-examples/master/users.json>`_
 
 ``users.json`` is an array of objects that resemble the following:
 
--- a/source/usage-examples/bulkWrite.txt
+++ b/source/usage-examples/bulkWrite.txt
@@ -25,17 +25,17 @@ documentation for full details.
   of the operation, such as :manual:`write concern
   </reference/write-concern>` and order.
 
-  By default, MongoDB executes bulk operations in the specified order
-  (i.e. serially). When executing operations in order, if an error occurs
-  during the processing MongoDB returns without processing any of the
-  remaining operations in the list. In contrast, when ``ordered`` is
-  ``false``, MongoDB continues to process remaining write operations in
-  the list. Unordered operations are faster as MongoDB can execute the
-  operations in parallel, but the results of the operation may vary. For
-  example, a ``deleteOne`` operation run before an ``updateMany``
-  operation might have a different result from running it after the
-  ``updateMany``. Refer to :ref:`Execution of Operations
-  <bulkwrite-write-operations-executionofoperations>` for more
+  By default, MongoDB executes bulk operations one-by-one in the
+  specified order (i.e. serially). During an ordered bulk write, if
+  an error occurs during the processing of an operation, MongoDB returns without
+  processing the remaining operations in the list. In contrast,
+  when ``ordered`` is ``false``, MongoDB continues to process remaining
+  write operations in the list. Unordered operations are faster as
+  MongoDB can execute the operations in parallel, but the results of the
+  operation may vary. For example, a ``deleteOne`` operation run before
+  an ``updateMany`` operation might have a different result from running
+  it after the ``updateMany``. Refer to :manual:`Execution of Operations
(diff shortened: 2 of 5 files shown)


conf.py
```diff
@@ -61,7 +61,7 @@
 pygments_style = 'sphinx'
 
 extlinks = {
-    'manual': ('http://docs.mongodb.com/manual%s', ''),
+    'manual': ('https://docs.mongodb.com/manual%s', ''),
```
schmalliso: Ignore this -- I'm reverting this change since it doesn't actually do anything anyway.

source/usage-examples/bulkWrite.txt
```diff
@@ -0,0 +1,86 @@
+=======================
+Perform Bulk Operations
+=======================
+
+.. default-domain:: mongodb
+
+Overview
+--------
+
+MongoDB provides clients the ability to perform write operations in bulk.
```
nathan-contino-mongo: Name-dropping the method name in this first chunk of text would align your formatting with our other examples and avoid burying the lede in terms of answering the question "how do i perform bulk operations"

source/usage-examples/bulkWrite.txt
```diff
@@ -0,0 +1,86 @@
...
+Bulk operations affect a *single* collection.
+
+The following code sample performs a bulk write operation against the
```
nathan-contino-mongo: Our other usage examples have tried to flesh out the methodology before bringing up the code samples. For reference, we're using this document: https://docs.google.com/document/d/1s4C1gA0OiNWgLeYzIwTvBsT1D_XcP-7A-Lb5WNUmUT4/edit?folder=14KsXkgiGp4uu14nb0hBc45qwbMePMZCA#heading=h.fqr4s89j4ciy
to track how we're structuring these. It might be a good idea to move the "following code" example description down into the example section and keep the bulkWrite reference up here.

source/usage-examples/bulkWrite.txt
```diff
@@ -0,0 +1,86 @@
...
+``users`` collection in the ``sample_mflix`` database. Specifically, the
+program uses the :node-api:`collection.bulkWrite
+<Collection.html#bulkWrite>` method and ``insertOne`` operations to write
+the user information from a ``users.json`` file to the ``users``
```
nathan-contino-mongo: What kinds of files can I use for this? just json? xml? binary? txt? If there's a reference describing this, I would link to it (and if not, you can add that to our missing-links file in ``source/usage-examples//``

source/usage-examples/bulkWrite.txt
```diff
@@ -0,0 +1,86 @@
...
+collection. Download the dataset here: `users.json <https://raw.githubusercontent.com/mongodb-university/universal-driver-examples/master/users.json>`_
+
+``bulkWrite()`` accepts the following parameters:
+
+- ``operations`` (``Array.<object>``): specifies the bulk operations to
```
nathan-contino-mongo: While I like the fact that you document the type immediately after the parameter name, I think we're trying to keep the tone of these slightly less reference-y so I've omitted this level of detail in my own. I figure that since we're saying the same thing later on ("an object in an array") it's mostly redundant anyway.

source/usage-examples/bulkWrite.txt
```diff
@@ -0,0 +1,86 @@
...
+  during the processing MongoDB returns without processing any of the
+  remaining operations in the list. In contrast, when ``ordered`` is
+  ``false``, MongoDB continues to process remaining write operations in
+  the list. Unordered operations are faster as MongoDB can execute the
+  operations in parallel, but the results of the operation may vary. For
+  example, a ``deleteOne`` operation run before an ``updateMany``
+  operation might have a different result from running it after the
+  ``updateMany``. Refer to :ref:`Execution of Operations
+  <bulkwrite-write-operations-executionofoperations>` for more
+  information.
```
nathan-contino-mongo: Big fan of this example here, I think you nailed our approach of "explain a concept in an approachable way, without too much jargon, and link to some reference doc". Perfect.

source/usage-examples/bulkWrite.txt
```diff
@@ -0,0 +1,93 @@
...
+  perform. Each operation is passed to ``bulkWrite()`` as an object in
+  an array.
+
+- ``options``: *optional* settings that affect the execution
+  of the operation, such as :manual:`write concern
+  </reference/write-concern>` and order.
+
+  By default, MongoDB executes bulk operations in the specified order
+  (i.e. serially). When executing operations in order, if an error occurs
+  during the processing MongoDB returns without processing any of the
```
nathan-contino-mongo: comma after the first processing here? my brain got all loopy when i read this the first time and had to figure out where to pause.

source/usage-examples/bulkWrite.txt
```diff
@@ -0,0 +1,93 @@
...
+- ``operations``: specifies the bulk operations to
+  perform. Each operation is passed to ``bulkWrite()`` as an object in
+  an array.
+
+- ``options``: *optional* settings that affect the execution
+  of the operation, such as :manual:`write concern
+  </reference/write-concern>` and order.
+
+  By default, MongoDB executes bulk operations in the specified order
+  (i.e. serially). When executing operations in order, if an error occurs
```
nathan-contino-mongo: I think serially might actually be too jargon-y here, but I can't deny it's correct and keeps things brief. If you can think of a simpler way to explain this, it might be easier for less-experienced users. Something like:
 "executes bulk operations in the order they are provided"?
"in order, so the first item in the provided collection completes before the second begins"?
but hopefully less awful and wordy than what I've said could be nice.

source/usage-examples/bulkWrite.txt
```diff
@@ -0,0 +1,93 @@
...
+  Promise resolves to a :node-api:`bulkWriteOpCallback </Collection.html#~bulkWriteOpCallback>` object containing the
+  result of the operation.
+
+``bulkWrite()`` returns a ``BulkWriteResult`` object
+that includes the number of updated or inserted documents, the ``_id`` of
+each document, and any ``writeError`` or ``writeConcernErrors``. Refer to
+the documentation for :node-api:`BulkWriteResult </BulkWriteResult>` for
+more information.
+
+If you create an index with a :ref:`unique constraint
```
nathan-contino-mongo: the link here is broken, sadly. I also think "If you create an index" might give the wrong impression -- this is really talking about "If you perform bulk operations on a collection with a unique constraint", so that might be a clearer way of expressing this.

nathan-contino-mongo: (I think the ref link destination doesn't exist here?)

source/usage-examples/bulkWrite.txt
```diff
@@ -0,0 +1,93 @@
...
+encounter warnings or errors related to the formatting of inserted or modified
+documents.
+
+Example
+-------
+
+The following code sample performs a bulk write operation against the
+``users`` collection in the ``sample_mflix`` database. Specifically, the
+program formats the data from a ``users.json`` source file to perform a
+series of ``insertOne`` operations with the ``bulkWrite()``collection
```
nathan-contino-mongo: I think the missing space before "collection" is messing with the formatting here

source/usage-examples/bulkWrite.txt
```diff
@@ -0,0 +1,93 @@
...
+method. Download the dataset here: `users.json
+<https://raw.githubusercontent.com/mongodb-university/universal-driver-examples/master/users.json>`_
+
+``users.json`` is an array of objects that resemble the following:
```
nathan-contino-mongo: is -> contains sounds slightly better to me, use your best judgment

source/usage-examples/bulkWrite.txt
```diff
@@ -0,0 +1,93 @@
...
+.. code-block:: javascript
+
+   { 
+     "email" : "marieconrad@example.com",
+     "password" : "450f6704710dcf8033157978f7fbdfde",
+     "name" : "Marie Conrad"
+   }
+
+The existing documents in the ``users`` collection also contain
+``email``, ``password``, and ``name`` fields.
```
nathan-contino-mongo: This sentence seems kind of lonely on its own between two code samples. Maybe we could combine it with the sentence before the example document by just mentioning that the ``users.json`` documents have the same structure as the documents already in the ``users`` collection?

source/usage-examples/bulkWrite.txt
schmalliso: Historically, we've been able to use refs from the manual in other projects. I'm not sure if that's something nextgen supports yet, so I'll just link to the manual page directly. I've also rephrased the paragraph.

source/usage-examples/bulkWrite.txt
schmalliso: Rephrased for newbies, though I kept 'i.e. serially' for folks who know what that means.

source/usage-examples/bulkWrite.txt
schmalliso: that was not a good sentence. ugh. rephrased

source/usage-examples/bulkWrite.txt
schmalliso: sgtm

source/usage-examples/bulkWrite.txt
schmalliso: Fair. Tweaked the intro to the sample doc.

PR #13, opened 2019-12-12
Docsp 7603 update a document usage

mongomoe:
ticket:

https://jira.mongodb.org/browse/DOCSP-7603
stage:

https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker/DOCSP-7603-Update-A-Document-Usage/usage-examples/updateOne


source/usage-examples/updateOne.txt
```diff
@@ -0,0 +1,41 @@
...
+
+.. default-domain:: mongodb
+
+You can update a single document using the `updateOne() <https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#updateOne>`_ method. 
+``updateOne()`` takes a filter for a document, and updates the first document
+that matches the filter.
+
+Create an `object
+<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object>`_
+to specify additional options. Set the ``upsert`` option to create a new
```
terakilobyte: set it to what?

source/usage-examples/updateOne.txt
```diff
@@ -0,0 +1,41 @@
...
+document if no documents match the filter.
+
+The ``updateOne()`` method returns `Promise
```
terakilobyte: returns a/the/?

source/usage-examples/updateOne.txt
```diff
@@ -0,0 +1,41 @@
...
+<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise>`_
+that resolves to an object that contains information, such as the number
+of documents that match the filter, and the number of documents that
```
terakilobyte: could updateOne ever modify more than 1 document?

source/usage-examples/updateOne.txt
```diff
@@ -0,0 +1,41 @@
...
+were modified.
+
+Specifying incorrect parameters for your ``updateOne()`` operation can
```
terakilobyte: We spoke about this paragraph offline.

mongomoe: Yup, fixed I did change one word from the agreed upon paragraph.

From: 

"Specifying incorrect parameters for your updateOne() operation can cause
errors"

to:

"Specifying incorrect parameters for your updateOne() operation can cause
problems"

source/usage-examples/updateOne.txt
mongomoe: No, it can only return 0 or 1, hence the original statement "number of documents".

I've updated the fragment to "...such as the number
of documents that match the filter, and if a document was updated"

source/usage-examples/updateOne.txt
mongomoe: Fixed

source/usage-examples/updateOne.txt
mongomoe: Fixed

source/usage-examples/updateOne.txt
```diff
@@ -0,0 +1,37 @@
+=================
+Update a Document
+=================
+
+.. default-domain:: mongodb
+
+You can update a single document using the `updateOne() <https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#updateOne>`_ method. 
+``updateOne()`` takes a filter for a document, and updates the first document
```
nathan-contino-mongo: We should specify that ``updateOne()`` also takes an update document -- otherwise, it is unclear how you define the mutation that is applied.

source/usage-examples/updateOne.txt
```diff
@@ -0,0 +1,37 @@
...
+that matches the filter.
+
+Create an `object
+<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object>`_
+to specify additional options. Set the ``upsert`` option to ``true`` to create a new
+document if no documents match the filter.
+
+The updateOne() method returns a `Promise
```
nathan-contino-mongo: updateOne() should be monospaced here.

source/usage-examples/updateOne.txt
```diff
@@ -0,0 +1,37 @@
...
+<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise>`_
+that resolves to an object. The ``modifiedCount`` field of this object
+has a value of 0 if no documents were updated, and a value of 1 if a
```
nathan-contino-mongo: 1 and 0 should be monospaced here for consistency with our other example pages.

source/usage-examples/updateOne.txt
mongomoe: Fixed

source/usage-examples/updateOne.txt
mongomoe: Fixed

source/code-snippets/usage-examples/updateOne.js
```diff
@@ -0,0 +1,37 @@
...
+    options.upsert = true; // create a document if no documents match the filter
+
+    const result = await collection.updateOne(
+      filter,
+      {
+        $set: {
+          plot:
+            "Blacksmith Scene is a silent film directed by William K.L. Dickson",
+        },
+      },
```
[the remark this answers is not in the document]
mongomoe: Fixed

source/code-snippets/usage-examples/updateOne.js
```diff
@@ -0,0 +1,37 @@
+// ignored first line
+const { MongoClient } = require("mongodb");
+
+const client = new MongoClient(
```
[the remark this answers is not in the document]
mongomoe: fixed

source/usage-examples/updateOne.txt
mongomoe: fixed

source/code-snippets/usage-examples/updateOne.js
```diff
@@ -0,0 +1,39 @@
...
+    const options = {};
+    options.upsert = true; // create a document if no documents match the filter
+    // create a document that sets the plot of the movie
+    const updateDoc = {
+      $set: {
+        plot:
+          "Blacksmith Scene is a silent film directed by William K.L. Dickson",
+      },
+    };
+    const result = await collection.updateOne(filter, updateDoc, options);
```
nathan-contino-mongo: Instead of a try around the entire body of this function, can we just use .next() and .catch() to handle the promise/errors generated by updateDoc? I'll probably create a PR updating the other methods too if you can get it to work here -- it's much closer to best practice than callbacks.

We'll want to leave the .catch() in the last line to handle connection-level errors though.

mongomoe: +1 @terakilobyte on his thoughts for this

nathan-contino-mongo: As a compromise, could we leave the ``try``/``finally`` in here but add a

```
.then (console.log())
.catch(error) 
```
We need the finally in there anyway, so this might let us keep both models.

source/usage-examples/updateOne.txt
```diff
@@ -0,0 +1,41 @@
...
+Update a Document
+=================
+
+.. default-domain:: mongodb
+
+You can update a single document using the `updateOne()
+<https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#updateOne>`_
+method. ``updateOne()`` takes a filter for a document, and updates the
+first document that matches the filter using an update document. The
+update document uses the :manual:`$set operator
```
terakilobyte: This statement isn't entirely [accurate](https://docs.mongodb.com/manual/reference/operator/update/#id1). I'd suggest rephrasing it to state that it accepts an available update operator to perform the desired update.

source/code-snippets/usage-examples/findOneAndUpdate.js
```diff
@@ -0,0 +1,50 @@
...
+const { MongoClient } = require("mongodb");
+
+// Replace the following with your MongoDB deployment's connection string.
+const uri =
+  "mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&w=majority";
+
+const client = new MongoClient(uri, {
+  useNewUrlParser: true,
+  useUnifiedTopology: true,
+});
```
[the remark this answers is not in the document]
mongomoe: Fixed

source/code-snippets/usage-examples/updateOne.js
```diff
@@ -0,0 +1,39 @@
...
+const { MongoClient } = require("mongodb");
+
+// Replace the following with your MongoDB deployment's connection string.
+const uri =
+  "mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&w=majority";
+
+const client = new MongoClient(uri, {
+  useNewUrlParser: true,
+  useUnifiedTopology: true,
+});
```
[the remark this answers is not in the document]
mongomoe: Fixed

source/usage-examples/updateOne.txt
mongomoe: Fixed:

* Changed the sentence to: "The update document uses the $set operator to set the value of a field in a document"

source/code-snippets/usage-examples/updateOne.js
nathan-contino-mongo: I disagree -- I feel that handling error messages differently for different levels of abstraction demonstrates explicit error handling to users. Specifically I think it's useful to separate connection level errors from logic/internal errors related to the ``updateOne`` call itself; in a small example like this it doesn't make a huge difference, but it does matter for operations that use multiple methods in sequence where the result of one method affects the next call and so on. I also think it would be useful to explicitly print useful fields of the result object instead of just printing the entire object.

That being said, all of my thoughts here are really subjective. So if y'all feel like it's fine as is I think that's acceptable too -- just trying to make this as useful as possible for end users :smiling_imp:

source/usage-examples/updateOne.txt
terakilobyte: That's what I mean though, `$set` is just one of many update operators available.

PR #15, opened 2019-12-13
(DOCSP-8056): CRUD tutorial/guide

nathan-contino-mongo:
JIRA: https://jira.mongodb.org/browse/DOCSP-8056

Stage: https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker/crud-guide/crud


source/crud.txt
```diff
@@ -0,0 +1,94 @@
...
+  navigate matched documents.
+
+- If you want to transform materialized views of documents in a
+  collection, you can use ``aggregate()``. This method accepts a
+  pipeline of aggregation commands that are run in sequence.
+
+- If you want to monitor a collection for new documents that match a
+  certain criteria, you can use ``watch()``. This method takes a
+  pipeline of aggregation commands that are run in sequence on new
+  data whenever write operations are run on the collection.
```
nlarew: - Let's make sure that these are distinct from the text in the paragraph above
- Maybe set these off with a "the driver includes methods for each type of read operation" or some other phrase that transitions into talking about specific methods?

source/crud.txt
```diff
@@ -0,0 +1,94 @@
...
+Write Operations
+----------------
+
+CRUD write operations help you get data into MongoDB. Sometimes this
+means putting entirely new data into a collection, but other times it
+involves changing data that already exists, or even removing it
+entirely. There are three distinct operations: create, update, and
+delete. Write operations allow you to handle return values with
+Promises and throw exceptions if an error occurs during operation
+execution.
```
nlarew: - Again, not "help" - it's just the way you do it
- We should try to lead with a solid definition. It's more than just "get data into MongoDB". It's "add, modify, and remove data". Can we emphasize that in this first sentence?

- three distinct _write_ operations
- Info about promises and error handling feels a bit too low-level to include in this section IMO. Maybe mention the underlying reason instead?
  - CRUD operations are all asynchronous (reads as well as writes)

source/crud.txt
nlarew: Can we reformat these bullet points to use subheading like we do under Write Operations?

e.g. 
- Read Operations
  - Find
  - Aggregate
  - Watch (or Subscribe)
- Write Operations
  - Create
  - Update
  - Delete

source/crud.txt
```diff
@@ -0,0 +1,94 @@
...
+  :doc:`find() </usage-examples/find>` and
+  :doc:`findOne() </usage-examples/findOne>`. These methods accept a
+  query document that describes the documents you would like to view.
+  While ``findOne()`` returns a single document, ``find()`` returns a
+  :doc:`cursor </crud/read-operations/cursor>` that you can use to
+  navigate matched documents.
+
+- If you want to transform materialized views of documents in a
+  collection, you can use ``aggregate()``. This method accepts a
+  pipeline of aggregation commands that are run in sequence.
```
nlarew: Agg is _way_ more than materialized views.tTose only arrived in the most recent version (4.2) with $merge.

- [Aggregation](https://en.wikipedia.org/wiki/Aggregate_function) fundamentally means functions that create summary data for groups of rows/documents (e.g. sum, average, maximum, range, etc.). 
  - That definition is a little loose because you don't need to use a $group or otherwise group documents (though at that point why are you using agg?)
- MongoDB also lets you filter and manipulate your working set. (e.g. $filter, $project, $addFields, etc.)
- You define MongoDB aggregations in a pipeline of stages that each operate on the result set of the previous stage in the pipeline

source/crud.txt
```diff
@@ -0,0 +1,94 @@
...
+
+- If you want to monitor a collection for new documents that match a
+  certain criteria, you can use ``watch()``. This method takes a
```
nlarew: **New** documents, yes, but also **updated**, **replaced**, and **deleted** documents

source/crud.txt
```diff
@@ -0,0 +1,94 @@
...
+
+Create
+~~~~~~
+
+If you want to add new documents to a collection, you can use
+:doc:`insertOne() </usage-examples/insertOne>` or
+:doc:`insertMany() </usage-examples/insertMany>`. These methods accept a
+single document or a collection of documents respectively. Note that the
+driver will generate a unique ``_id`` field automatically for any
+inserted documents that lack such a field.
```
nlarew: Good info in this paragraph, but it's not clear to me how this is distinct from the `insertOne` example page. Should it be distinct? Should we link somewhere for interested readers to learn more?

source/crud.txt
```diff
@@ -0,0 +1,94 @@
...
+~~~~~~
+
+If you want to alter existing documents in a collection, you can use
+:doc:`updateOne() </usage-examples/updateOne>`,
+:doc:`replaceOne() </usage-examples/replaceOne>`, or
+``updateMany()``. These methods
+accept a query document that describes the documents you would like to
+change and an update document that describes the changes you would
+like to apply to matched documents. In the case of ``replaceOne()``,
+a replacement document is used instead of an update document.
```
nlarew: - What is a replacement document?
- Where can I learn more about this?

source/crud.txt
```diff
@@ -0,0 +1,94 @@
...
+
+Delete
+~~~~~~
+
+If you want to remove existing documents from a collection, you can
+use :doc:`deleteOne() </usage-examples/deleteOne>` or
+:doc:`deleteMany() </usage-examples/deleteMany>`. These methods accept a
+query document that describes the documents you would like to delete.
```
nlarew: Or, more strongly, the operations delete a/any documents that match the query filter you define.

source/crud.txt
```diff
@@ -0,0 +1,94 @@
...
+   /crud/compound-operations
+
+Overview
+--------
+
+MongoDB helps you store data persistently across different instances of
+your applications. CRUD (Create, Read, Update, Delete) operations allow
+you to work with the data stored inside of MongoDB. CRUD operations fall
+into two categories: read operations and write operations. All operation
+methods can be configured with a special ``options`` parameter.
```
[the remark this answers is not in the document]
nathan-contino-mongo: Unfortunately there isn't a good place to link to for your last bullet point. I agree with all of your thoughts here, and have restructured the intro as such.

source/crud.txt
```diff
@@ -0,0 +1,94 @@
...
+---------------
+
+CRUD read operations help you get data out of MongoDB.
+Read operations provide flexibility so you can use them to
+view data in many different ways. While some read operations focus on
+allowing you to view data that exists in the database, other operations
+let you monitor the database for new data that matches certain criteria,
+and still other read operations let you apply a series of
+transformations and filters to data before returning a resulting
+emphemeral collection.
```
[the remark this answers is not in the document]
nathan-contino-mongo: - Agreed on your first point. Reworded as per your suggestion.

- While I see what you're saying for the second point (and even recognize that your suggestion is technically more correct vocabulary), I think the verbiage you suggest is a little intimidating/confusing for new users. I lean more toward my initial wording since it actually conveys the thought more effectively. Happy to discuss better terms though if you still feel like this could use improvement!

- agreed, that's sort of a jargon-y term anyway. Reworded as per your suggestion.

source/crud.txt
nathan-contino-mongo: I think we should circle back to this. Summing up the aggregation framework in a couple of sentences is a nontrivial task... unfortunately I don't think we have a concrete replacement for my too-simplified explanation right now. Going to reword it and see if we can iterate from that.

source/crud.txt
nathan-contino-mongo: Good point, standardized with the format of Write. Added an intro sentence as well. I don't see any full sentences/large phrases that are duplicated between the intro and the subsections, but if you see any I'm happy to change.

source/crud.txt
nathan-contino-mongo: Agreed all around. Fixed.

source/crud.txt
nathan-contino-mongo: Good point -- my intent here is "brief summary." I do link to the usage examples here so I think we offer that "next step" reading but if you think I should link somewhere in the manual/etc. I can do that too.

source/crud.txt
nathan-contino-mongo: I'll describe replacement documents in the "update a document" page (which I'm thinking is probably better named "change a document") which will give some high level explanations of things like $set, $inc, and the concepts of replace/update. Will link here under "replacement document".

nathan-contino-mongo: Also added a reference to upsert.

source/crud.txt
nathan-contino-mongo: Added a clolsing sentence to this effect.

source/crud/compound-operations.txt
```diff
@@ -0,0 +1,52 @@
...
+
+This property can be useful in cases where you want to write to a
+specific document, but you haven't found it yet. If you just perform a
+read for the document's ``_id`` and then try to alter the document you
+just found, it's possible that someone else can alter the document in
+between your read and write operations. This doesn't stop you from doing
+this work, but it can make error handling much more difficult. Compound
+operations help keep your logic straightforward by handling that logic
+entirely inside the database behind a layer of abstraction, so you don't
+have to worry about it.
```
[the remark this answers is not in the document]
nathan-contino-mongo: True... but should we really encourage our users to make use of transactions? I understand that there are a lot of valid use cases, but for your average introductory user transactions will just lead them astray when they could use these methods instead and avoid a lot of potential error cases.

I'd be fine with adding a section toward the bottom that talks about transactions, and how transactions essentially allow you to compose your own versions of these compound operations... but I'd probably want a use case outside of findOneAndUpdate/Replace/Delete to do that. Maybe we could talk about ``findMultipleAndDelete``? Would appreciate any suggestions.

source/crud.txt
```diff
@@ -0,0 +1,118 @@
...
+Find
+~~~~
+
+If you want to view documents already in a collection, you can use
+:doc:`find() </usage-examples/find>` and
+:doc:`findOne() </usage-examples/findOne>`. These methods accept a
+query document that describes the documents you would like to view.
+While ``findOne()`` returns a single document, ``find()`` returns a
+:doc:`cursor </crud/read-operations/cursor>` that you can use to
+navigate matched documents.
```
nlarew: Can we add an example here?

```rst
.. example::
   
   A pizza restaurant wants to find all large pizzas that included pepperoni that were sold yesterday. They run the following ``find()`` query on the ``orders`` collection:
   
   .. code-block:: javascript
      
      await orders.find({
        toppings: "pepperoni",
        date: {
          $gt: new Date(Date.now() - 24*60*60*1000)
        }
      });
```

source/crud.txt
```diff
@@ -0,0 +1,118 @@
...
+   /crud/compound-operations
+
+Overview
+--------
+
+CRUD (Create, Read, Update, Delete) operations allow you to work with
+the data stored inside of MongoDB. CRUD operations fall into two
+categories: read operations and write operations. All CRUD operations
+offer customizable options to tailor behavior for your application's
+specific needs. CRUD operations are all asynchronous.
```
nlarew: These last two sentences seem a little bit tacked on. I think we could expand each into a separate paragraph.

- **Options:** the sentence you have is good. Maybe we could use the space to point to some of the read/write operation sub pages that show different options? (e.g. **Sort Results**)
- **Async:** what does asynchronous mean, why does it matter, and how does it affect driver code? Async code is notorious tough for beginners so we should take care to introduce it well.

source/crud.txt
```diff
@@ -0,0 +1,118 @@
...
+navigate matched documents.
+
+Aggregate
+~~~~~~~~~
+
+If you want to create custom processing pipelines for documents in a
+collection, you can use ``aggregate()``. This method accepts a
+pipeline of aggregation commands that are run in sequence. These
+commands let you filter, summarize, and augment documents in a
+collection into a result set for viewing.
```
nlarew: Can we add an example here?

e.g. A pizza restaurant wants to know for each ingredient the number of pizzas that were sold last week that included the ingredient

source/crud.txt
```diff
@@ -0,0 +1,118 @@
...
+
+Watch / Subscribe
+~~~~~~~~~~~~~~~~~
+
+If you want to monitor a collection for new, update, replace, and
+deleted documents that match a certain criteria, you can use
+``watch()``. This method takes a pipeline of aggregation commands that
+are run in sequence on new data whenever write operations are run on
+the collection.
```
nlarew: Can we add an example here?

e.g. A pizza restaurant wants to get a notification whenever a new pizza order comes in

source/crud.txt
nlarew: I think one way we can make it distinct is to add a small _narrative_ example (as described in other comments)

source/crud.txt
```diff
@@ -0,0 +1,118 @@
...
+:doc:`updateOne() </usage-examples/updateOne>`,
+:doc:`replaceOne() </usage-examples/replaceOne>`, or
+``updateMany()``. These methods
+accept a query document that describes the documents you would like to
+change and an `update document <replacementDocuments>`_ that describes
+the changes you would like to apply to matched documents.
+``replaceOne()`` uses a `replacement document <replacementDocuments>`_
+that describes an entirely new document instead of an update document.
+You can configure any update operation to behave as an
+:doc:`upsert </crud/write-operations/upsert>` instead of an update.
```
nlarew: Can we add an example here?

e.g. A pizza restaurant wants to modify the delivery address for an order that already exists

source/crud.txt
```diff
@@ -0,0 +1,118 @@
...
+Delete
+~~~~~~
+
+If you want to remove existing documents from a collection, you can
+use :doc:`deleteOne() </usage-examples/deleteOne>` or
+:doc:`deleteMany() </usage-examples/deleteMany>`. These methods accept a
+query document that describes the documents you would like to delete.
+Delete operations will delete either one (in the case of
+``deleteOne()``) or all (in the case of ``deleteMany()``) documents that
+match the query.
```
nlarew: Can we add an example here?

e.g. A pizza restaurant wants to delete an order that was made in error

(Not sure if this is great. In reality you'd probably want to make this a boolean `cancelled` field instead of deleting the document.)

source/crud/compound-operations.txt
```diff
@@ -0,0 +1,52 @@
...
+  matches multiple documents to a supplied query and removes the first
+  of those matched documents.
+
+- :node-api:`findOneAndUpdate() </Collection.html#findOneAndUpdate`
+  matches multiple documents to a supplied query and updates the first
+  of those matched documents using the provided update document.
+
+- :node-api:`findOneAndReplace() </Collection.html#findOneAndReplace`
+  matches multiple documents to a supplied query and replaces the first
+  of those matched documents using the provided replacement document.
```
nlarew: - "There are **_two_** major compound operations" but we list three
- These are all just wrappers around `findAndModify()`. Do we want to include that too?

source/crud/compound-operations.txt
```diff
@@ -0,0 +1,52 @@
...
+
+All three methods accept an optional ``options`` object with
+configurable :doc:`sort </crud/read-operations/sort>` and
+:doc:`projection </crud/read-operations/project>` options that work just
+like their read operation equivalents. ``findOneAndUpdate()`` and
+``findOneAndDelete()`` allow the client to configure the
+``returnOriginal``  option, a boolean that determines if the method
+returns the original document or the altered document.
```
nlarew: I feel like we could give just a bit more detail about `returnOriginal`, specifically being clear that it's **pre-update** vs **post-update** versions of the modified document.

source/crud.txt
nathan-contino-mongo: Good point. I've expanded both of these points into subsections of their own right. Right now I feel like the customization section does a decent job of explaining some basic options/pointing you towards node driver doc pages, but the async section is a little tougher. Would appreciate suggestions on:
- section name
- tweaks to how I approach the explanation, which is a bit... hackneyed right now

source/crud.txt
nathan-contino-mongo: Adding all of these examples now

source/crud/compound-operations.txt
nathan-contino-mongo: I am inclined to omit findAndModify as that's what I'd call an "advanced user" concept. For an introduction to these concepts, it's an unnecessary implementation detail -- any user that actually requires an explanation of it beyond the reference material probably doesn't have a real use case for it anyway.

source/crud/compound-operations.txt
```diff
@@ -0,0 +1,52 @@
...
+.. default-domain:: mongodb
+
+Usually client applications only need to read data out of a database or
+write data into a database. However, client applications sometimes need
+to read and write data in a single interaction with the database. While
+you can accomplish this task using separate reads and writes, doing so
+requires the client application to gracefully handle potential errors
+at any stage of the process and in multiple potential error states.
+This increases the complexity of your code and can make your client
+application brittle and difficult to test.
```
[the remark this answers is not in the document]
nathan-contino-mongo: - 10/10 suggestion, starting off with "requests" instead of "client applications" makes my intent far clearer to the reader, since client applications generally *do* need to both read and write to the db.
- Good point about ordering, moved that statement.

source/crud/compound-operations.txt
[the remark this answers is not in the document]
nathan-contino-mongo: Going to break the "compound operations" page into Transactions vs. built-in methods as a starting point. I think "compound operations" actually encapsulates both, so this might be a good way to bring in transactions too without pushing users too far toward them

source/crud/query-document.txt
```diff
@@ -0,0 +1,197 @@
...
+.. default-domain:: mongodb
+
+Specify a Query
+~~~~~~~~~~~~~~~
+
+All CRUD operations apply to a set of matched documents in a collection.
+By default, these operations match **all** documents in a collection.
+Sometimes, however, you want to match only certain items in a
+collection; the query document allows you to make your searches more
+specific. Query documents contain one or more field-level expressions
```
nlarew: I like the info in this paragraph! Can we put a little more emphasis upfront on the idea that "you use query expressions to configure which document(s) a query applies to"? I think it could be more effective if we open with that and then use the "by default" case to contrast.

source/crud/read-operations/cursor.txt
```diff
@@ -0,0 +1,146 @@
...
+Work With Operation Results
+===========================
+
+.. default-domain:: mongodb
+
+Read operations that return multiple documents do not immediately return
+all values matching the query. Instead, these operations rely upon an
+object called a **cursor** that fetches documents in batches. Cursors
+are highly configurable and offer multiple interaction paradigms for
+different use cases.
```
nlarew: Can we give a little intuition as to _why_ we use cursors?

- to work with potentially large result sets that may exceed working memory
- to minimize network bandwidth usage

source/crud/read-operations/cursor.txt
```diff
@@ -0,0 +1,146 @@
+===========================
+Work With Operation Results
+===========================
```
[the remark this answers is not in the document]
nlarew: The section on this page feel like they could use a bit more hierarchy. Maybe we can order them around paradigms and nest accordingly.

source/crud/read-operations/geo.txt
```diff
@@ -0,0 +1,163 @@
+===================
+Search Geospatially
+===================
+
+.. default-domain:: mongodb
+
+You can query against geospatial indexes in several ways with
+`geospatial query operators
+<https://docs.mongodb.com/manual/geospatial-queries/index.html>`_.
```
nlarew: Can we focus this on the idea of geospatial data rather than geospatial indexes? The indexes are obviously very important but I don't think they appropriately motivate the page as part of the introduction.

source/crud/read-operations/geo.txt
```diff
@@ -0,0 +1,163 @@
...
+  index on the ``location.geo`` field of the ``theaters`` collection of
+  the ``sample_mflix`` database, use the following command on the
+  sample_mflix database:
+
+  .. code-block:: javascript
+
+     db.movies.createIndex({location.geo: "2dsphere"})
+
+Query by Proximity
+------------------
```
nlarew: These are examples, right? Let's nest them under the **Examples** heading (i.e. use `~~~~` instead of `----`)

source/crud/read-operations/limit.txt
```diff
@@ -0,0 +1,78 @@
...
+
+.. default-domain:: mongodb
+
+Use limit to cap the number of documents that can be returned from a
+read operation. Limit functions as a cap on the maximum number of
+documents that the operation can return, but the operation can return
+a smaller number of documents if there are not enough documents present
+to reach the limit. If limit is used with the skip method, the skip will
+apply first and the cap will only apply to the documents left over after
+the skip.
```
nlarew: Can we link to skip?

source/crud/read-operations/limit.txt
```diff
@@ -0,0 +1,78 @@
...
+
+Consider a collection containing documents that describe books. To
+insert this data into a collection, run the following operation:
+
+.. code-block:: javascript
+
+   await collection.insertMany([
```
nlarew: nit: collection -> books

source/crud/read-operations/limit.txt
```diff
@@ -0,0 +1,78 @@
...
+     { "_id": 6, "name": "A Dance With Dragons", "author": "Tolkein", "length": 1104 },
+   ]);
+
+Pass the following sort document to a read operation to ensure that the
+operation returns books with longer lengths before books with shorter
+lengths. Applying a limit of 3 to this operation causes this operation
+to return only the 3 longest books:
+
+.. code-block:: javascript
+   :emphasize-lines: 4
```
nlarew: can we highlight the `.sort(sort)` and `limit` lines as well?

source/crud/read-operations/limit.txt
```diff
@@ -0,0 +1,78 @@
...
+
+   // define an empty query document
+   const query = {};
+   // sort in descending (-1) order by length
+   const sort = { length: -1 };
+   const limit = 3;
+   const cursor = collection.find(query).sort(sort).limit(limit);
+   await cursor.forEach(console.dir);
+
+``find()`` returns the following documents:
```
nlarew: ... sorted from longest length to shortest length

source/crud/read-operations/limit.txt
```diff
@@ -0,0 +1,78 @@
...
+
+.. code-block:: javascript
+
+   { "_id": 3, "title": "Atlas Shrugged", "author": "Rand", "length": 1088 }
+   { "_id": 5, "title": "Cryptonomicon", "author": "Stephenson", "length": 918 }
+   { "_id": 1, "title": "The Brothers Karamazov", "author": "Dostoyevsky", "length": 824 }
+
+You can combine skip and limit in this way to implement paging for your
+collection, allowing users to view only small "slices" of the
+collection at once.
```
nlarew: skip/limit paging works but isn't optimal. Maybe we can include a note here and have a different guide down the line?

https://www.codementor.io/@arpitbhayani/fast-and-efficient-pagination-in-mongodb-9095flbqr

source/crud/read-operations/project.txt
```diff
@@ -0,0 +1,94 @@
+==============================
+Specify Which Fields to Return
+==============================
+
+.. default-domain:: mongodb
+
+Use a projection to control which fields appear in the documents
+returned by read operations. You can accomplish this in two
+ways:
```
nlarew: Can we add a bit of motivation?

- You may only want some fields from documents that you query
- Useful for matching your application interfaces
- Useful for minimizing network bandwidth

source/crud/read-operations/project.txt
```diff
@@ -0,0 +1,94 @@
...
+explicitly include fields, you cannot explicitly exclude fields, and
+vice versa.
+
+Consider a collection containing documents that describe varieties of
+fruit. To insert this data into a collection, run the following
+operation:
+
+.. code-block:: javascript
+
+   await collection.insertMany([
```
nlarew: nit: collection -> fruits

source/crud/read-operations/text.txt
```diff
@@ -0,0 +1,260 @@
...
+phrases, instead of matching substrings like
+`regular expresssions
+<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions>`_,
+also known as RegEx. While regex is very powerful, and MongoDB even
+offers a `$regex <>` operator, text searches work particularly well when
+applied to unstructured text, like transcripts, essays, or even web
+pages. Because MongoDB's text indexes are aware of concepts like
+plurality, case sensitivity, stop words, whitespace, and punctuation in
+multiple languages, all you have to do is query with a search term to
+find documents that contain your term.
```
nlarew: Pretty solid!

- What's a "collection of strings"? Is this really "string-valued fields"?
- Maybe we could have a blurb about why `$text` is more performant than `$regex`?
- nit: I'd avoid starting a sentence with "Because" - you can probably invert the sentence from "Because REASON, FOO" to "FOO because REASON".
- nit: "all you have to do" is like "simple" and "easy" - probably better to avoid that type of phrasing altogether

source/crud/read-operations/text.txt
```diff
@@ -0,0 +1,260 @@
...
+
+You can only create *one* text index per collection, so every text
+search queries all fields used by the index for term matches.
```
nlarew: This is important. Can we put it in an important admonition?

source/crud/read-operations/text.txt
```diff
@@ -0,0 +1,260 @@
...
+   { title: 'Star Trek VI: The Undiscovered Country' }
+   { title: 'Star Trek V: The Final Frontier' }
+   { title: 'Star Trek IV: The Voyage Home' }
+   { title: 'Star Trek III: The Search for Spock' }
+   { title: 'Star Trek II: The Wrath of Khan' }
+
+Success! The query found every document in the ``movies`` collection
+with a title including the word "trek". Unfortunately, it looks like the
+search included one unintended item: "Trek Nation," which is a movie
+about Star Trek and not a Star Trek movie.
```
nlarew: Nice example! Can we make this just a little more self-contained by pointing to the next section?

e.g. "To solve this, we can query with a more specific phrase"

source/crud/read-operations/cursor.txt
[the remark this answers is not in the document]
nathan-contino-mongo: I feel like this could be restructured a bit... let's talk about it on monday.

source/crud.txt
```diff
@@ -0,0 +1,205 @@
...
+
+.. note::
+
+   CRUD operations execute **asynchronously** because the driver
+   communicates with MongoDB over the network, which is subject to
+   latency, timeouts, and other complications that can delay execution
+   during synchronous communication. As a result, each CRUD
+   method returns a Promise that resolves to the output of that
+   operation. A common method of Promise resolution uses
+   ``async``/``await`` syntax.
```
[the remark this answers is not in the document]
nathan-contino-mongo: Agreed. Monday conversation.

source/crud.txt
```diff
@@ -0,0 +1,205 @@
...
+      :language: javascript
+      :start-after: start delete crud example
+      :end-before: end delete crud example
+
+Compound operations
+-------------------
+
+Some operations combine aspects of read and write operations. Consult
+:doc:`compound operations </crud/compound-operations>` to learn more
+about these hybrid methods.
```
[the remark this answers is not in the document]
nathan-contino-mongo: We could, but I do feel like it's important to bring up transactions/hybrid operations somewhere in here to surface the idea. Let's talk about it on monday and come to a consensus.

source/crud.txt
```diff
@@ -0,0 +1,205 @@
...
+All CRUD operations offer customizable options to tailor behavior for
+your application's specific needs. You can customize read operations to
+:doc:`skip </crud/read-operations/skip>` a certain number of documents
+at the beginning of the collection of matched documents,
+:doc:`limit </crud/read-operations/limit>` the number of returned
+documents, and :doc:`sort <crud/read-operations/sort>` returned
+documents in a particular order. You can customize write operations to
+:doc:`upsert </crud/write-operations/upsert>` instead of update, time
+out if the method fails to complete within a certain period of time,
+and write with varying levels of resilience in a replica set.
```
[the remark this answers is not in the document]
nathan-contino-mongo: Adding a note to each of these. We'll revisit this later, I'm sure.

source/code-snippets/usage-examples/geofind.js
```diff
@@ -0,0 +1,59 @@
...
+    const collection = database.collection("theaters");
+
+    /*const query = {
+      "location.geo": {
+        $near: {
+          $geometry: { type: "Point", coordinates: [-73.9667, 40.78] },
+          $maxDistance: 10000,
+        },
+      },
+    };*/
```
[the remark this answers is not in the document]
nathan-contino-mongo: I'll modify this when I switch the geo page to literalincludes -- this will go away then :)

source/crud/query-document.txt
```diff
@@ -0,0 +1,197 @@
+===============
+Specify a Query
+===============
+
+.. default-domain:: mongodb
+
+Specify a Query
+~~~~~~~~~~~~~~~
+
+All CRUD operations apply to a set of matched documents in a collection.
```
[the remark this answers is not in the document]
nathan-contino-mongo: All -> most :)

source/crud/query-document.txt
nathan-contino-mongo: I've tried to reword this a couple of times and I can't find a structure I prefer/like as much as the current approach. Unless you can propose a better wording, I think I'd rather keep it the way it is now, but feel free to suggest!

source/crud/read-operations/geo.txt
nathan-contino-mongo: Agreed, reworded to introduce the concept better. Added a second sentence about required indexes though, since that's an important prerequisite.

source/crud/read-operations/limit.txt
nathan-contino-mongo: I would... but I can't do that any more after I move them to literalincludes. Put in a request with the platform team for that :)

source/crud/read-operations/limit.txt
nathan-contino-mongo: Well if your index doesn't cover your query that doesn't make a huge difference and also relies upon non-guaranteed attributes of _id (that may not be true if you generate your own _id values)... so I'd probably avoid it. Still, it is worth mentioning that you can use less than/greater than with any field to accomplish the same thing, so maybe a brief example of that could help here. I mostly just wanted to bring up a simple-to-understand use case, so if we can come up with a better one/decide that isn't really necessary we can just avoid the whole thing.

source/crud/read-operations/text.txt
nathan-contino-mongo: - switched to "string-valued fields"
- I would hesitate to imply that $text is always more performant than regex -- regex is darn fast. Regex on an indexed field is probably in the same league as $text, though obviously it has different use cases. $text offers enough usability advances for certain purposes that I don't think we need to worry about performance differences.
- I tried inverting this sentence and I actually really prefer starting it with "because", though obviously that's subjective. If you feel strongly I can always change it despite that.
- I understand our aversion to "simple" and "easy" (because it can seem condescending to an end user on either a conscious or unconscious level) but I can't think of a better way to structure this thought. Would appreciate any suggestions on this.

source/crud/read-operations/text.txt
nathan-contino-mongo: Agreed.

source/crud/read-operations/text.txt
nathan-contino-mongo: Originally this was just one big section that I broke up into different subheadings, will add this to make it flow better.

nathan-contino-mongo: Also added to other sections as well.

source/crud/read-operations/geo.txt
```diff
@@ -0,0 +1,124 @@
...
+
+The number of coordinates depends upon the object type; a Point requires
+only one coordinate: a longitude and a latitude. A Line requires
+two coordinates: a longitude and a latitude for each end. Polygons
+consist of a list of coordinates in which the first and last coordinate
+are the same, effectively closing the polygon. To learn more about
+the GeoJSON shapes you can use in MongoDB, consult the
+`GeoJSON manual entry
+<https://docs.mongodb.com/manual/reference/geojson/>`_. To geospatially
+query GeoJSON data, you must first add your GeoJSON data to a
```
ccho-mongodb: It seems like this is applicable for all the different types of 2D queries listed here (or at least the first two).
Would this be something you can link to? Or perhaps be added to its own sub-section?

source/crud/read-operations/geo.txt
```diff
@@ -0,0 +1,124 @@
...
+   :start-after: start proximity geo example
+   :end-before: end proximity geo example
+   :dedent: 4
+
+Query Within a Range
+~~~~~~~~~~~~~~~~~~~~
+
+The ``$geoWithin`` operator selects documents with geospatial data that
+exist within a specified shape. The following example searches for
+movie theaters in the new england area:
```
ccho-mongodb: Capitalize "New England" here

source/crud/read-operations/geo.txt
```diff
@@ -0,0 +1,124 @@
...
+Examples
+--------
+
+The following examples use the ``theaters`` collection of the
+``sample_mflix`` sample database available in MongoDB Atlas, with a
+``2dsphere`` index on the ``location.geo`` field.
+
+.. note::
+
+  Geo searches require a geospatial index. To create a ``2dsphere``
```
ccho-mongodb: Should this mention both "2d" and "2dsphere" index types? Otherwise, could it specify which geo search you want to perform?

source/crud/read-operations/cursor.txt
```diff
@@ -0,0 +1,150 @@
...
+   const cursor = collection.find({});
+   cursor.on('data', data => console.log(data));
+
+Fetch All Documents At Once
+---------------------------
+
+For use cases that require all documents matched by a query to be held
+in memory at the same time, use ``toArray()``. Note that large sets of
+matched documents can cause performance issues or even failures due to
+excessive memory requirements.
```
ccho-mongodb: Should this be "exceeding" rather than "excessive"? I think "excessive" does not imply resource or limit-breaking which I'm assuming is the point you're making here.

source/crud/read-operations/cursor.txt
```diff
@@ -0,0 +1,150 @@
...
+
+- ``Collection.aggregate()``
+
+- ``Collection.listIndexes()``
+
+- ``Db.aggregate()``
+
+- ``Db.listCollections()``
+
+Other methods, like ``Collection.findOne()`` and ``Collection.watch()``
```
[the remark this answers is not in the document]
nathan-contino-mongo: Good point, `generate` probably isn't the right word choice here. I'm going to change this to `return` because that is probably a more familiar term to many readers and is technically correct (though it does gloss over the details... but we get to those later anyway). Going to avoid using the term "private" in this case because javascript doesn't actually have a formal notion of private like some other languages and I feel like the current wording is a little less jargon-y.

source/crud/read-operations/cursor.txt
```diff
@@ -0,0 +1,150 @@
...
+use cursors to generate results, but do not directly expose those
+cursors to the method caller.
+
+.. warning::
+
+   Do not combine different cusor paradigms on a single cursor.
```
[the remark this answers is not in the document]
nathan-contino-mongo: This term also bothered me, but I can't think of a better term to explain this. I'm going to bring up paradigms explicitly below here and move this warning down so that this isn't the first/only time readers hear the phrase "cursor paradigm."

source/crud/read-operations/cursor.txt
```diff
@@ -0,0 +1,150 @@
...
+   Cursors are designed under the assumption that users will use only
```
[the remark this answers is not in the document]
nathan-contino-mongo: This is exactly what I was trying to string together, but I couldn't get past my original wording! Thanks so much, using this exactly.

source/crud/read-operations/cursor.txt
```diff
@@ -0,0 +1,150 @@
...
+   **one** of the following paradigms to access data. Failure to use
+   just one of these paradigms will lead to undefined behavior. For
+   instance, calling both ``count()`` and ``toArray()`` or both
+   ``hasNext()`` and ``forEach()`` on the same cursor will likely lead
+   to an error.
+
+.. warning::
```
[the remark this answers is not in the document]
nathan-contino-mongo: Trimmed the length down but leaving two warnings, since they are distinctive/i feel like readers might only take away one of them from a single warning.

source/crud/read-operations/cursor.txt
```diff
@@ -0,0 +1,150 @@
...
+
+   Do not execute multiple async methods on a single cursor in parallel.
+   Only perform one async operation at a time on a single cursor.
+   Attempting multiple async operations in parallel on a single cursor
+   will result in undefined behavior. For example,
+   pushing the result of multiple ``next()`` method calls to a
+   collection of promises and then calling ``Promise.all()`` on that
+   collection of promises will likely lead to an error.
+
+View Cursor Contents One By One
```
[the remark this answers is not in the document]
nathan-contino-mongo: Reworded this entirely now, but very valid point

source/crud/read-operations/cursor.txt
```diff
@@ -0,0 +1,150 @@
...
+-------------------------------
+
+Many use cases only require one document at a time. Cursors are designed
+with this usage in mind, abstracting away network accesses and caching
```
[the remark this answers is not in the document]
nathan-contino-mongo: Hm, this is a really good point -- I already mention this in the intro, but I could get rid of this point entirely in this section. Right now it's there to demonstrate the advantages of handling one document at a time, but that's probably irrelevant once you get to this point.

source/crud/read-operations/cursor.txt
```diff
@@ -0,0 +1,150 @@
...
+~~~~~~~~~
+
+As Readable Streams, Cursors also support the Event API.
+
+.. code-block:: javascript
+
+   const cursor = collection.find({});
+   cursor.on('data', data => console.log(data));
+
+Fetch All Documents At Once
```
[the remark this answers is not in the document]
nathan-contino-mongo: Your suggested structure is closer to what I intended but I messed up the headers. Restructuring entirely now so I'm going to resolve this.

source/crud/read-operations/cursor.txt
```diff
@@ -0,0 +1,150 @@
...
+   Attempting multiple async operations in parallel on a single cursor
+   will result in undefined behavior. For example,
+   pushing the result of multiple ``next()`` method calls to a
+   collection of promises and then calling ``Promise.all()`` on that
+   collection of promises will likely lead to an error.
+
+View Cursor Contents One By One
+-------------------------------
+
+Many use cases only require one document at a time. Cursors are designed
```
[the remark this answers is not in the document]
nathan-contino-mongo: Good point, I've renamed this section to make it more paradigm (though I'm going to avoid using that word) centric and I'm changing the setup to just talk about paradigms in general rather than talking about use cases at all.

source/crud/read-operations/geo.txt
```diff
@@ -0,0 +1,124 @@
+===================
+Search Geospatially
+===================
+
+.. default-domain:: mongodb
+
+You can query data based on geographical location using
+`geospatial query operators
+<https://docs.mongodb.com/manual/geospatial-queries/index.html>`_.
+Geospatial queries require a geospatial index.
```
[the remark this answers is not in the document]
nathan-contino-mongo: Added a link to geospatial index manual explanation

source/crud/read-operations/geo.txt
```diff
@@ -0,0 +1,124 @@
...
+
+Coordinates on an Earth-like Sphere
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For geospatial queries using longitude and latitude coordinates
+on an Earth-like sphere, use the `GeoJSON
+<https://docs.mongodb.com/manual/geospatial-queries/index.html#geospatial-geojson>`_
+format. While GeoJSON has `multiple types
+<https://docs.mongodb.com/manual/reference/geojson/>`_, GeoJSON data
+generally looks like this:
```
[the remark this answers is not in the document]
nathan-contino-mongo: Reworded to "use some form of the following structure" in both places.

source/crud/read-operations/geo.txt
```diff
@@ -0,0 +1,124 @@
...
+   <field> : {
+      type: <GeoJSON type>,
+      coordinates: [
+         [longitude1, latitude1],
+         ...
+         [longitudeN, latitudeN]
+      ]
+   }
+
+The number of coordinates depends upon the object type; a Point requires
```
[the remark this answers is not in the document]
nathan-contino-mongo: Agreed, reversed.

source/crud/read-operations/geo.txt
nathan-contino-mongo: You mean creating the index? GeoJSON data uses a `2dsphere index` whereas coordinate pairs use a `2d` index. I want to keep them separate here because I think they two different kinds of index can be confusing when presented together.

source/crud/read-operations/geo.txt
nathan-contino-mongo: done.

source/crud/read-operations/geo.txt
nathan-contino-mongo: Well since index type is dependent upon data type, this particular collection for the examples *must* use a `2dsphere` index. Not sure if that answers your question.

source/crud/read-operations/cursor.txt
nathan-contino-mongo: True, just using a lot of memory won't cause errors of itself -- it's only when we pass a certain point that that becomes a problem. Rephrased as per your suggestion.

source/crud/read-operations/cursor.txt
```diff
@@ -7,10 +7,11 @@ Access Data From a Cursor
 Read operations that return multiple documents do not immediately return
 all values matching the query. Because a query can potentially match
 very large sets of documents, these operations rely upon an
-object called a **cursor** that fetches documents in batches to reduce
-both memory consumption and network bandwidth usage. Cursors
-are highly configurable and offer multiple interaction paradigms for
-different use cases.
+object called a `cursor
+<https://mongodb.github.io/node-mongodb-native/3.4/api/Cursor.html>`_
```
ccho-mongodb: I think you can use the `node-docs` role here:
https://github.com/mongodb/snooty-parser/blob/master/snooty/rstspec.toml#L530

source/crud/read-operations/cursor.txt
```diff
@@ -7,10 +7,11 @@ Access Data From a Cursor
...
+that fetches documents in batches to reduce both memory consumption and
+network bandwidth usage. Cursors are highly configurable and offer
+multiple interaction paradigms for different use cases.
```
ccho-mongodb: When using the adjectives "multiple ___ for different ____" together, it sounds like you're describing a cross-product. I think you only need one adjective. E.g. "...offer interaction paradigms for different use cases" or "offer multiple interaction paradigms." If it is actually a cross-product, feel free to leave as is.

nathan-contino-mongo: Agreed. I think it does apply as a cross product here -- each use case can traverse a cursor in a few different ways.

source/crud/read-operations/cursor.txt
nathan-contino-mongo: Yeah, I definitely need to update a lot of links with this.

source/crud/read-operations/cursor.txt
```diff
@@ -0,0 +1,174 @@
...
+- ``Collection.aggregate()``
+
+- ``Collection.listIndexes()``
+
+- ``Db.aggregate()``
+
+- ``Db.listCollections()``
+
+Other methods, like ``Collection.findOne()`` and ``Collection.watch()``,
+use cursors to generate results but do not directly expose those
```
ccho-mongodb: From this conversation, I think "generate" -> "return" makes sense.
https://github.com/mongodb/docs-node/pull/15#discussion_r366408004

source/crud/read-operations/cursor.txt
```diff
@@ -0,0 +1,174 @@
...
+cursors to the method caller.
+
+Cursor Paradigms
+----------------
+
+You can work with cursors using a number of **cursor paradigms**.
+Many use cases only require one document at a time, so many of these
+cursor paradigms take advantage of this by abstracting away network
```
ccho-mongodb: Would this still make sense and convey the intended message if you left out "Many use cases only require one document at a time"? I think we should only include general statements like this that are either explained shortly afterwards or otherwise omit them. E.g. sans general statement
"The cursor paradigms allow you to access the query result one document at a time and abstract away the network and caching logic".

source/crud/read-operations/cursor.txt
```diff
@@ -0,0 +1,174 @@
...
+   Do not combine different cursor paradigms on a single cursor.
+   Operations such as ``hasNext()``, ``forEach()``, and ``toArray()``
+   each predictably modify the original cursor. If you mix these calls
+   on a single cursor, you may receive unexpected results.
+
+.. warning::
+
+   Because asynchronous calls directly modify the cursor, executing
+   asynchronous calls on a single cursor simultaneously can also cause
+   undefined behaviour. As a result, always wait for the previous
```
ccho-mongodb: Would this retain its original intent by omitting "As a result" so that the imperative "always wait..." is more immediate? Also, would use "one" in the second mention of "asynchronous operation".

source/crud/read-operations/cursor.txt
```diff
@@ -0,0 +1,174 @@
...
+.. literalinclude:: /code-snippets/crud/cursor.js
+   :language: javascript
+   :start-after: start foreach cursor example
+   :end-before: end foreach cursor example
+   :dedent: 4
+
+Asynchronous Iteration
+~~~~~~~~~~~~~~~~~~~~~~
+
+Cursors implement the AsyncIterator interface, which allows you to
```
ccho-mongodb: Nitpick: Use pre-formatted text for AsyncIterator. Alternatively, link to the API reference doc.

source/crud/read-operations/cursor.txt
```diff
@@ -0,0 +1,174 @@
...
+use Cursors in more traditional ``for``...``await`` loops:
```
ccho-mongodb: Nitpick 1: We should stay consistent with capitalization of "cursor".
Nitpick 2: We can probably leave out adjectives like "traditional" for brevity since it does not add much value for readers that either know or do not know what they are.

source/crud/read-operations/limit.txt
```diff
@@ -0,0 +1,80 @@
...
+
+.. default-domain:: mongodb
+
+Use limit to cap the number of documents that can be returned from a
+read operation. Limit functions as a cap on the maximum number of
+documents that the operation can return, but the operation can return
+a smaller number of documents if there are not enough documents present
+to reach the limit. If limit is used with the
+:doc:`skip </crud/read-operations/skip>` method, the skip applies
+first and the caponly applies to the documents left over after
```
ccho-mongodb: Did you mean "caponly" -> "cap only"

source/crud/read-operations/limit.txt
```diff
@@ -0,0 +1,80 @@
...
+     { "_id": 2, "name": "Les Misérables", "author": "Hugo", "length": 1462 },
+     { "_id": 3, "name": "Atlas Shrugged", "author": "Rand", "length": 1088 },
+     { "_id": 4, "name": "Infinite Jest", "author": "Wallace", "length": 1104 },
+     { "_id": 5, "name": "Cryptonomicon", "author": "Stephenson", "length": 918 },
+     { "_id": 6, "name": "A Dance With Dragons", "author": "Tolkein", "length": 1104 },
+   ]);
+
+Pass the following sort document to a read operation to ensure that the
+operation returns books with longer lengths before books with shorter
+lengths. Applying a limit of 3 to this operation causes this operation
```
ccho-mongodb: Nitpick: maintain the imperative voice here for parallel structure among the sentences.
"Pass the following sort document..."
"Apply a limit of `3` to return only the three longest books..."

source/crud/read-operations/skip.txt
```diff
@@ -0,0 +1,58 @@
+=====================
+Skip Returned Results
+=====================
+
+.. default-domain:: mongodb
+
+
+Use skip to omit documents from the beginning of the list of returned
```
ccho-mongodb: It could be helpful to the user to use an inline literal for the `skip` function here and any mention of skip or sort following this.

ccho-mongodb: Also, I would add a description of the parameter that it takes in this sentence. E.g.: "Use skip to omit a specified number of documents from the beginning of the list of returned..."

source/crud/read-operations/skip.txt
```diff
@@ -0,0 +1,58 @@
...
+documents for a read operation. You can combine skip with sort to omit
+the top (for descending order) or bottom (for ascending order) ``N``
```
ccho-mongodb: Users might not be familiar with using "N" to represent an integer.

source/crud/read-operations/cursor.txt
nathan-contino-mongo: Reworded this section -- take another look and see if you have any suggestions. Incorporated your suggestions here.

source/crud/read-operations/skip.txt
```diff
@@ -0,0 +1,58 @@
...
+results for a given query. Since the order of documents returned is
+:manual:`not guaranteed
```
ccho-mongodb: I wonder if this link would work better around "order of documents" as I would otherwise expect to be linked to a document section called "not guaranteed" which, to me, is both confusing without context and has some negative connotation associated.

source/crud/read-operations/cursor.txt
nathan-contino-mongo: Getting rid of "as a result", but I'm not 100% sold on changing that second mention of "asynchronous operation" to "one". Instead, I'm just ending with "another" -- I think that solves things nicely without resorting to a vague pronoun.

source/crud/read-operations/cursor.txt
nathan-contino-mongo: Linking to the mdn docs for this, good point.

source/crud/read-operations/skip.txt
```diff
@@ -0,0 +1,58 @@
...
+</reference/method/cursor.sort/#result-ordering>` in the absence of a
+sort, using skip without using sort omits arbitrary documents.
+
+If the value of skip exceeds the number of matched documents for your
```
ccho-mongodb: A couple requests for this sentence:
1) Could "value of skip" be described as the parameter passed to skip instead? Otherwise, it sounds like `skip` is a variable and not a method.
2) Would it be possible to replace the second instance of a noun with a pronoun for "your query, your query"? I think "it" works ok.
3) I think there are two tenses here, but would prefer just one: "If x is y" (present perfect?) and "y will z" (future).

source/crud/read-operations/cursor.txt
nathan-contino-mongo: Both good points.
- standardized lowercase cursor
- removed "traditional" (and on the look out for more cases of extraneous adjectives that don't make docs any better)

source/crud/read-operations/skip.txt
```diff
@@ -0,0 +1,58 @@
...
+      sort : { rating: -1 },
+      // omit the first two documents
+      skip : 2,
+   }
+
+   const cursor = collection.find(query, options);
+   await cursor.forEach(console.dir);
+
+In this case, the number ``-1`` tells the read operation to sort the
+fruit in descending order by rating. Skip elides the first two matched
```
ccho-mongodb: Nitpick: I would replace "omits" rather than "elides" as its a more common in both spoken and written English.

source/crud/read-operations/skip.txt
nathan-contino-mongo: I don't do this for limit, sort, project, or sort right now -- I think I'd only make this change if we stayed consistent across pages, unless there is a reason for us to differ for some of these. I do think an inline literal makes sense here, so I'm going to make the change across pages and just revert it if we feel it isn't necessary.

source/crud/read-operations/skip.txt
```diff
@@ -0,0 +1,58 @@
...
+the top (for descending order) or bottom (for ascending order) ``N``
+results for a given query. Since the order of documents returned is
+:manual:`not guaranteed
+</reference/method/cursor.sort/#result-ordering>` in the absence of a
+sort, using skip without using sort omits arbitrary documents.
+
+If the value of skip exceeds the number of matched documents for your
+query, your query will return no documents.
+
+Consider a collection containing documents that describe varieties of
```
ccho-mongodb: You might consider adding a transition sentence here that specifically signals to the reader that they are about to start a multi-step example. One idea is "Follow the instructions in the examples below to insert data in a collection and perform a sort and skip on the query results.

source/crud/read-operations/skip.txt
nathan-contino-mongo: The sentence is pretty clear even without N, so I'm removing this entirely. Thanks!

source/crud/read-operations/skip.txt
ccho-mongodb: Similar comment for the `read-operations/sort.txt file`

source/crud/read-operations/skip.txt
nathan-contino-mongo: Done.

source/crud/read-operations/sort.txt
```diff
@@ -0,0 +1,82 @@
+============
+Sort Results
+============
+
+.. default-domain:: mongodb
+
+Use a sort to change the order in which operations return documents.
```
ccho-mongodb: Nitpick: to stay consistent with `read-operations/skip.txt`, you should consider specifying "...read operations return documents".

source/crud/read-operations/skip.txt
nathan-contino-mongo: 1) Tricky, because depending on the method, skip can be either a field (always) or a cursor method (only sometimes). I think I'd prefer to leave it as is so the sentence applies to either usage.
2) changed to "for a query, that query" instead
3) "will return" -> "returns"

source/crud/read-operations/skip.txt
nathan-contino-mongo: Done.

source/crud/read-operations/skip.txt
ccho-mongodb: Similar comment for the `read-operations/sort.txt` file.

source/crud/read-operations/sort.txt
```diff
@@ -0,0 +1,82 @@
...
+Sorts tell MongoDB to order returned documents by the values of one or
```
ccho-mongodb: Would refer to sort in the singular form, especially since that is the name of the function that you are introducing. E.g. "Sort tells MongoDB...".

source/crud/read-operations/sort.txt
```diff
@@ -0,0 +1,82 @@
...
+.. code-block:: javascript
+
+   { "_id": 2, "title": "Les Misérables", "author": "Hugo", "length": 1462 }
+   { "_id": 6, "title": "A Dance With Dragons", "author": "Martin", "length": 1104 }
+   { "_id": 4, "title": "Infinite Jest", "author": "Wallace", "length": 1104 }
+   { "_id": 3, "title": "Atlas Shrugged", "author": "Rand", "length": 1088 }
+   { "_id": 5, "title": "Cryptonomicon", "author": "Stephenson", "length": 918 }
+   { "_id": 1, "title": "The Brothers Karamazov", "author": "Dostoyevsky", "length": 824 }
+
+To resolve ties in your sorted results, add additional fields to the
```
ccho-mongodb: I wonder if it might be useful to describe a "tie", just to cover cases in which there might be ambiguity in what measure is being used. E.g. "To resolve ties (two or more matching values in the results), add additional fields..."

source/crud/read-operations/sort.txt
```diff
@@ -0,0 +1,82 @@
...
+   :emphasize-lines: 4
+
+   // define an empty query document
+   const query = {};
+   // sort in ascending (1) order by length
+   const sort = { length: 1, author: 1 };
+   const cursor = collection.find(query).sort(sort);
+   await cursor.forEach(console.dir);
+
+In this case, the the read operation sorts matching documents first by
```
ccho-mongodb: https://uploads-ssl.webflow.com/596519d17019ab1f49671113/596519d17019ab1f49671f9a_Porky%20Pig.jpg

source/crud/read-operations/sort.txt
```diff
@@ -0,0 +1,82 @@
...
+``length`` and, in the event of a tie, then by ``author``.
+``find()`` returns the following ordering of documents when this sort is
+used with an empty query:
```
ccho-mongodb: While "empty query document" in the code comment (line 63) seems correct, I'm not sure if "empty query" is the correct term. In this instance, I think you can reword the sentence to avoid using that term:
"In this case, MongoDB sorts documents matching the query by ``length`` and..."

source/crud/read-operations/sort.txt
nathan-contino-mongo: Added a couple of extra sentences clarifying what a tie is and what's going on in the particular case.

source/crud/read-operations/sort.txt
nathan-contino-mongo: the the the the

source/crud/read-operations/sort.txt
nathan-contino-mongo: Added a little more explanation for this paragraph, but used your suggestion of "documents matching the query" instead of "an empty query" because you're quite right. Thanks again!

source/crud/write-operations/arrayFilters.txt
```diff
@@ -0,0 +1,112 @@
+===========================
+Update Arrays in a Document
+===========================
+
+.. default-domain:: mongodb
+
+To modify arrays nested within a document, use the
```
nlarew: nested -> embedded

This is a general MongoDB style guide across all of our messaging

source/crud/write-operations/arrayFilters.txt
```diff
@@ -0,0 +1,112 @@
...
+</crud/write-operations/update-document` to modify positions in an array
+matched by conditions passed to the query via the ``arrayFilters``
+option.
+
+You can think of ``arrayFilters`` as an additional query level applied
+only to documents already matched by the update operation's query filter
+(the first parameter for ``updateOne()`` and ``updateMany()``). This
+query finds all positions in the referenced array that match your
+criteria, and you can use the returned position values to modify only
+those positions.
```
nlarew: - Rather than "positions in an array" should we say "specific elements in an array"? "Position values" and such aren't super clear terms and make this a little more confusing for me.
- The conditions are query expressions. I think it's important to convey that you're essentially running a subquery on the array elements. Thoughts?
- The "additional query level" framing doesn't clarify much IMO.

source/crud/write-operations/arrayFilters.txt
```diff
@@ -0,0 +1,112 @@
...
+
+Example
+~~~~~~~
+
+The following example demonstrates how you can use the filtered
+position operator with ``arrayFilters`` to modify only certain
+array positions for fields within documents matched by the query.
```
nlarew: to modify only specific array elements that match a query filter.

source/crud/write-operations/arrayFilters.txt
nathan-contino-mongo: - reworded
- added mention of subquery
- personally i think the first paragraph, while technically correct, is complete gibberish to anyone who doesn't already know what a query filter is. Or is at least overwhelming. The second paragraph is a lot more comprehensible to me (and when I inevitably forget what an arrayFilter is in 3 months I'll come back to this page and have a much better understanding after that paragraph)

source/crud/write-operations/arrayFilters.txt
```diff
@@ -0,0 +1,112 @@
...
+
+.. default-domain:: mongodb
+
+To modify arrays embedded within a document, use the
+:manual:`filtered positional operator
+</reference/operator/update/positional-filtered/index.html>`. You can
+use this operator in any :doc:`update document
+</crud/write-operations/update-document` to modify specific elements in
+an array matched by conditions passed to the query via the
+``arrayFilters`` option.
```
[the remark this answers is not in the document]
nathan-contino-mongo: done.

nathan-contino-mongo:
Merged in, though I'm sure there are still some things to add to this guide we can always work on those in future branches. Want to make sure this is included in the overview for the meeting tomorrow.

PR #17, opened 2019-12-16
(DOCSP-7600): insertOne init

mongomoe:
Ticket: https://jira.mongodb.org/browse/DOCSP-7600
Stage: https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker/DOCSP-7600-insertOne/usage-examples/insertOne

### DIFF
--- a/source/usage-examples/insertOne.txt
+++ b/source/usage-examples/insertOne.txt
@@ -4,10 +4,9 @@ Create a Document
 
 .. default-domain:: mongodb
 
-To create a document, define an object variable with the fields and values that you want to specify. You can insert this document using the `insertOne()
+To create a document, define an object variable with the fields and values that you want to specify. You can insert this document into a collection using the `insertOne()
 <https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#insertOne>`_
-method. The ``insertOne()`` method inserts a document into the specified
-collection. If the specified collection does not exist, the
+method. If the specified collection does not exist, the
 ``insertOne()`` method creates the collection. 
 
 The ``insertOne()`` method returns a `Promise
--- a/source/usage-examples/insertOne.txt
+++ b/source/usage-examples/insertOne.txt
@@ -10,11 +10,6 @@ method. The ``insertOne()`` method inserts a document into the specified
 collection. If the specified collection does not exist, the
 ``insertOne()`` method creates the collection. 
 
-Create an `Object
-<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object>`_
-to specify additional options. Set the ``bypassDocumentValidation`` option to bypass
-:manual:`schema validation <core/schema-validation/>` rules in MongoDB 3.2 or higher.
-
 The ``insertOne()`` method returns a `Promise
 <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise>`_
 that resolves to an object. The ``insertedId`` field of this object is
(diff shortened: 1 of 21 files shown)


source/usage-examples/insertOne.txt
```diff
@@ -0,0 +1,27 @@
+=================
+Create a Document
+=================
+
+.. default-domain:: mongodb
+
+You can create a single document using the `insertOne()
+<https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#insertOne>`_
+method. The ``insertOne()`` method inserts a document into the specified collection
+using a document object. If the specified collection does not exist, the ``insertOne()``
```
nlarew: What is a document object? I think you mean `insertOne(<Document Object>)` - I'd omit the "object" portion:

"inserts a document in the specified collection"

source/usage-examples/insertOne.txt
```diff
@@ -0,0 +1,27 @@
+=================
+Create a Document
+=================
+
+.. default-domain:: mongodb
+
+You can create a single document using the `insertOne()
```
nlarew: Somewhat pedantic, but this is more about how to **Insert a Document** or **Add a Document**. The insert methods take a document that you already created and stores it in the collection.

source/usage-examples/insertOne.txt
```diff
@@ -0,0 +1,27 @@
...
+<https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#insertOne>`_
+method. The ``insertOne()`` method inserts a document into the specified collection
+using a document object. If the specified collection does not exist, the ``insertOne()``
+method creates the collection. 
+
+Create an `Object
+<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object>`_
+to specify additional options. Set the ``bypassDocumentValidation`` option to bypass
+:manual:`schema validation <core/schema-validation/>` rules in MongoDB 3.2 or higher.
```
nlarew: - This should use "You can..." to maintain the same mood as the rest of the example (potential mood vs imperative mood)
- Is `bypassDocumentValidation` a core option? Not sure which options we want to include/exclude, but if we're only including essential options this seems extraneous

source/usage-examples/insertOne.txt
mongomoe: Sentence replaced to "
To create a document, define an object variable with the fields and values that you want to specify. You can insert this document using the `insertOne()
<https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#insertOne>`_
method."

source/usage-examples/insertOne.txt
mongomoe: Fixed

source/usage-examples/insertOne.txt
mongomoe: I've removed this paragraph altogether since there are no typical/essential options that we would want for a usage-example of insertOne.

I've also removed `bypassDocumentValidation` from insertMany per Nathan's suggestion offline that we should not tell our users to bypass schema validation

source/usage-examples/insertOne.txt
```diff
@@ -0,0 +1,22 @@
...
+Create a Document
+=================
+
+.. default-domain:: mongodb
+
+To create a document, define an object variable with the fields and values that you want to specify. You can insert this document using the `insertOne()
+<https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#insertOne>`_
+method. The ``insertOne()`` method inserts a document into the specified
+collection. If the specified collection does not exist, the
+``insertOne()`` method creates the collection.
```
nlarew: Sentences 2&3 here are more or less the same. Can we combine them?

You can use `insertOne()` to add the document into a collection

mongomoe: Altered sentences 2&3 to be "You can insert this document into a collection. using the insertOne() method."

PR #18, opened 2019-12-17
DOCSP-7609: adds countDocuments and estimatedDocumentCount examples

schmalliso:
Jira: https://jira.mongodb.org/browse/DOCSP-7609

Staging: https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker/DOCSP-7609/usage-examples/count

### DIFF
--- a/source/code-snippets/usage-examples/count.js
+++ b/source/code-snippets/usage-examples/count.js
@@ -20,15 +20,10 @@ async function run() {
 
     // Estimate the total number of documents in the collection
     // and print out the count.
-    collection.estimatedDocumentCount({}, function(error, result) {
-      if (error) {
-        console.log("Error: " + error);
-      } else {
-        console.log(
-          "Estimated number of documents in the movies collection: " + result,
-        );
-      }
-    });
+    var estimate = await collection.estimatedDocumentCount();
+    console.log(
+      "Estimated number of documents in the movies collection: " + estimate,
+    );
 
     // Query for movies from Canada.
     const query = { countries: "Canada" };
@@ -36,13 +31,8 @@ async function run() {
     // Find the number of documents that match the specified
     // query, (i.e. with "Canada" as a value in the "countries" field)
     // and print out the count.
-    collection.countDocuments(query, {}, function(error, result) {
-      if (error) {
-        console.log("Error: " + error);
-      } else {
-        console.log("Number of movies from Canada: " + result);
-      }
-    });
+    var countCanada = await collection.countDocuments(query);
+    console.log("Number of movies from Canada: " + countCanada);
   } finally {
     await client.close();
   }
--- a/source/code-snippets/usage-examples/count.js
+++ b/source/code-snippets/usage-examples/count.js
@@ -1,0 +1,50 @@
+// ignored first line
+const { MongoClient } = require("mongodb");
+
+// Replace the following with your MongoDB deployment's connection
+// string.
+const uri =
+  "mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&w=majority";
+
+const client = new MongoClient(uri, {
+  useNewUrlParser: true,
+  useUnifiedTopology: true,
+});
+
+async function run() {
+  try {
+    await client.connect();
+
+    const database = client.db("sample_mflix");
+    const collection = database.collection("movies");
+
+    // Estimate the total number of documents in the collection
+    // and print out the count.
+    collection.estimatedDocumentCount({}, function(error, result) {
+      if (error) {
+        console.log("Error: " + error);
+      } else {
+        console.log(
+          "Estimated number of documents in the movies collection: " + result,
+        );
+      }
+    });
+
+    // Query for movies from Canada.
+    const query = { countries: "Canada" };
+
+    // Find the number of documents that match the specified
+    // query, (i.e. with "Canada" as a value in the "countries" field)
+    // and print out the count.
+    collection.countDocuments(query, {}, function(error, result) {
+      if (error) {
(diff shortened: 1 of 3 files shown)


source/usage-examples/count.txt
```diff
@@ -0,0 +1,61 @@
...
+
+.. default-domain:: mongodb
+
+Overview
+--------
+
+The Node.js driver provides two methods for counting the number of
+documents in a collection:
+
+- :node-api:`collection.countDocuments()
```
nathan-contino-mongo: Both of these links just drop you at Collection since ``:node-api:`` points at the 3.0 documentation instead of the most recent doc.

Suggested fix:
- change 3.0 to 3.3 in the definition of ``:node-api:``
- better yet, always point at the latest node-api docs... but I'm not 100% sure if this is possible, or even a good idea

source/usage-examples/count.txt
```diff
@@ -0,0 +1,61 @@
...
+  </Collection.html#countDocuments>` returns the number of documents in
+  the collection that match the specified query. If you specify a empty
+  query document, ``countDocuments()`` returns the total number of
+  documents in the collection.
+
+- :node-api:`collection.estimatedDocumentCount
```
nathan-contino-mongo: Same as above comment, just making sure you're aware

source/usage-examples/count.txt
```diff
@@ -0,0 +1,61 @@
...
+
+``countDocuments()`` accepts a :manual:`query
+</tutorial/query-documents>` parameter. ``countDocuments()`` will only
+count the documents that match the specified query.
+
+``countDocuments()`` and ``estimatedDocumentCount()`` support optional
+settings that affect the method's execution. Refer to the reference
+documentation for each method for more information.
+
+``countDocuments()`` and ``estimatedDocumentCount()`` behave in two
```
nathan-contino-mongo: Four paragraphs in a row that begin with a method name instead of an english word *feels* subjectively reference-y to me. Do you think we could spice this up a little bit, maybe with something like 
```
"the ``countDocuments()`` and ``estimatedDocumentCount()`` methods"
```
just to vary it up/make it feel less formal?

Of course there's nothing wrong with it as is, so use your best judgment -- this is a "look and feel" suggestion.

source/usage-examples/count.txt
```diff
@@ -0,0 +1,61 @@
...
+different ways depending on whether you specify a callback method:
+
+- If you do not specify a callback method, ``countDocuments()`` and
+  ``estimatedDocumentCount()`` return a `Promise
+  <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise>`_
+  that resolves to an `Object
+  <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object>`_.
```
nathan-contino-mongo: In this case I think the Promise actually just resolves to a number (If I'm reading the API docs right), so it might be clearer to just say "number"

source/usage-examples/count.txt
```diff
@@ -0,0 +1,61 @@
...
+Count the Number of Documents in a Collection
+=============================================
+
+.. default-domain:: mongodb
+
+Overview
+--------
+
+The Node.js driver provides two methods for counting the number of
+documents in a collection:
```
nathan-contino-mongo: While "the number of documents in a collection" isn't inaccurate, I feel like "counting documents in a collection" better conveys the idea that these methods have different counting use cases, since "the number of documents in a collection" is very similar to the wording used to explain ``estimatedDocumentCount()``

source/code-snippets/usage-examples/count.js
```diff
@@ -0,0 +1,50 @@
...
+      }
+    });
+
+    // Query for movies from Canada.
+    const query = { countries: "Canada" };
+
+    // Find the number of documents that match the specified
+    // query, (i.e. with "Canada" as a value in the "countries" field)
+    // and print out the count.
+    collection.countDocuments(query, {}, function(error, result) {
```
nathan-contino-mongo: Turns out that going forward the team has decided to stick with Promises for the usage examples, so we should probably switch these examples away from callbacks (tl;dr: js newcomers are more familiar with Promises than callbacks, and we're trying to keep things as approachable as possible to them)

As for the existing examples that use callbacks, I'll probably PR an edit that uses Promises instead sometime in the near future. Sorry if you've had conflicting signals on some of this stuff :disappointed:

source/usage-examples/count.txt
schmalliso: I've got a pull in against snooty to update it to 3.3; we can figure out how they do their URLs to see if there's a "latest" or "stable" or something to point to instead.

source/usage-examples/count.txt
schmalliso: Ack.

source/usage-examples/count.txt
schmalliso: Yup, fair! I'm tweaking.

source/usage-examples/count.txt
schmalliso: sgtm.

source/usage-examples/count.txt
schmalliso: Done.

source/code-snippets/usage-examples/count.js
schmalliso: Curses! Works for me 😁

schmalliso:
Thanks! Sue is cutting a build of the autobuilders with the snooty change this afternoon, at which point the links will point to 3.4.

PR #19, opened 2019-12-18
DOCSP-7608: nodejs usage example - collection distinct

ccho-mongodb:
JIRA: https://jira.mongodb.org/browse/DOCSP-7608

Staging: https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker/DOCSP-7608-collection-distinct/usage-examples/distinct


source/usage-examples/distinct.txt
```diff
@@ -0,0 +1,48 @@
...
+The ``distinct()`` method returns a
+`Promise <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise>`_
+that resolves to an
+`Array <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array>`_
+that contains each of the different values. If none of the documents in the
+collection contain the field specified in the method call, it returns an
+empty Array.
+
+If you specify a value for the document field name that is not a String such
+as a document, array, number, or null, the method does not execute and
```
nathan-contino-mongo: I have used monospaced font when I mention ``null`` in these examples, since it technically is a specific value of itself. You could do the same here for consistency.

source/usage-examples/distinct.txt
```diff
@@ -0,0 +1,48 @@
...
+Distinct Values of a Field in a Collection
+==========================================
+
+.. default-domain:: mongodb
+
+You can retrieve a list of distinct values for a field across a collection
+by using the
+`distinct()
+<https://mongodb.github.io/node-mongodb-native/3.4/api/Collection.html#distinct>`_
+method. The ``distinct()`` method is called on a collection object and
```
nathan-contino-mongo: Passive sentence, you could change this to:
```
Call the ``distinct()`` method on a Collection object with a document field name to produce a list...
```
which also flows nicely into the next sentence, which currently starts by repeating the gist of this sentence.

source/usage-examples/distinct.txt
```diff
@@ -0,0 +1,48 @@
...
+
+The ``distinct()`` method returns a
+`Promise <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise>`_
+that resolves to an
+`Array <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array>`_
+that contains each of the different values. If none of the documents in the
+collection contain the field specified in the method call, it returns an
+empty Array.
+
+If you specify a value for the document field name that is not a String such
```
nathan-contino-mongo: Not sure if we really need any more here, but what do you think about explaining this method's interaction with dot notation? (for instance, using field1.field2.field3 to query for values in nested objects?)
Other thought: arrays. What happens if I call this method on a field that is an array in some documents, but not others? Or a field that is an array in all documents in the collection? Does the type ``array`` show up, or individual values from arrays?

source/usage-examples/distinct.txt
ccho-mongodb: Good point. That made me think about how I phrased this sentence which could be more clear to identify that String, document, etc. are types instead of values. I can update null to be monospaced, though we should probably record that decision somewhere so that we can ensure consistency before release and beyond. There's also the adjective form of null when used as "null value" that we should decide on outside of this PR.

source/usage-examples/distinct.txt
ccho-mongodb: Thanks, will fix.

source/usage-examples/distinct.txt
ccho-mongodb: Thanks for bringing up this point; I should include more detail to clarify these cases. The value supplied to the document field name parameter still needs to be a string. To specify a field subdocument, you use dot notation (e.g. "imdb.rating"). If you call it on an array (e.g. directors) field, distinct() treats each element as a value (rather than the array as a whole), but if you call it on a subdocument (e.g. imdb) field, the fields within the subdocument identify it as a unique value. E.g. 
```
[
  { rating: 6.6, votes: 8921, id: 86619 },
  { rating: 6.6, votes: 11781, id: 102713 },
  { rating: 6.4, votes: 10666, id: 117057 }
]
```

nathan-contino-mongo:
LGTM, I don't think this needs any more polish! @ccho-mongodb feel free to merge whenever you're ready.

PR #20, opened 2019-12-20
(DOCSP-7605): replace a document

nathan-contino-mongo:
JIRA: https://jira.mongodb.org/browse/DOCSP-7605

Stage: https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker/DOCSP-7605/usage-examples/replaceOne

### DIFF
--- a/source/code-snippets/usage-examples/replaceOne.js
+++ b/source/code-snippets/usage-examples/replaceOne.js
@@ -40,7 +40,7 @@ async function run() {
       }
       if (result.upsertedCount === 1) {
         console.log(
-          "Inserted one new document with an _id of " + result.upsertedId._id,
+          "Inserted one new document with an _id of " + result.upsertedId._id
         );
       }
     }
--- a/source/code-snippets/usage-examples/replaceOne.js
+++ b/source/code-snippets/usage-examples/replaceOne.js
@@ -1,0 +1,51 @@
+// ignored first line
+const { MongoClient } = require("mongodb");
+
+// Replace the following with your MongoDB deployment's connection string.
+const uri =
+  "mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&w=majority";
+
+const client = new MongoClient(uri);
+
+async function run() {
+  try {
+    await client.connect();
+
+    const database = client.db("sample_mflix");
+    const collection = database.collection("movies");
+
+    // create a filter for a movie to update
+    const filter = { title: "Blacksmith Scene" };
+    const options = {
+      // create a document if no documents match the filter
+      upsert: true,
+    };
+    // create a new document that will be used to replace the existing document
+    const replacement = {
+      title: "Sandcastles in the Sand",
+      plot:
+        "Robin Sparkles mourns for a relationship with a mall rat at an idyllic beach.",
+    };
+
+    const result = await collection.replaceOne(filter, replacement, options);
+
+    if (result.modifiedCount === 0 && result.upsertedCount === 0) {
+      console.log("No changes made to the collection.");
+    } else {
+      if (result.matchedCount === 1) {
+        console.log("Matched " + result.matchedCount + " documents.");
+      }
+      if (result.modifiedCount === 1) {
+        console.log("Updated one document.");
+      }
+      if (result.upsertedCount === 1) {
+        console.log(
+          "Inserted one new document with an _id of " + result.upsertedId._id,
+        );
+      }
+    }
+  } finally {
+    await client.close();
+  }
+}
+run().catch(console.dir);
--- a/source/code-snippets/usage-examples/replaceOne.js
+++ b/source/code-snippets/usage-examples/replaceOne.js
@@ -40,7 +40,7 @@ async function run() {
       }
       if (result.upsertedCount === 1) {
         console.log(
-          "Inserted one new document with an _id of " + result.upsertedId._id
+          "Inserted one new document with an _id of " + result.upsertedId._id,
         );
       }
     }
--- a/source/code-snippets/usage-examples/replaceOne.js
+++ b/source/code-snippets/usage-examples/replaceOne.js
@@ -14,10 +14,10 @@ async function run() {
     const database = client.db("sample_mflix");
     const collection = database.collection("movies");
 
-    // create a filter for a movie to update
-    const filter = { title: "Blacksmith Scene" };
+    // create a query for a movie to update
+    const query = { title: "Blacksmith Scene" };
     const options = {
-      // create a document if no documents match the filter
+      // create a document if no documents match the query
       upsert: true,
     };
     // create a new document that will be used to replace the existing document
@@ -27,7 +27,7 @@ async function run() {
         "Robin Sparkles mourns for a relationship with a mall rat at an idyllic beach.",
     };
 
-    const result = await collection.replaceOne(filter, replacement, options);
+    const result = await collection.replaceOne(query, replacement, options);
 
     if (result.modifiedCount === 0 && result.upsertedCount === 0) {
       console.log("No changes made to the collection.");
--- a/source/usage-examples/replaceOne.txt
+++ b/source/usage-examples/replaceOne.txt
@@ -7,8 +7,11 @@ Replace a Document
 You can replace a single document using the `replaceOne()
 <https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#replaceOne>`_
 method. ``replaceOne()`` accepts a query document and a replacement
-document, and replaces the first document that matches the filter with
-the provided replacement document.
+document, and fully replaces the first document that matches the query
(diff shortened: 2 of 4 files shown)


source/usage-examples/replaceOne.txt
```diff
@@ -0,0 +1,46 @@
+==================
+Replace a Document
+==================
+
+.. default-domain:: mongodb
+
+You can replace a single document using the `replaceOne()
+<https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#replaceOne>`_
+method. ``replaceOne()`` accepts a query document and a replacement
+document, and replaces the first document that matches the filter using
```
ccho-mongodb: Nitpick: 
I wonder if is less ambiguous: "first document... using the ... replacement document" or "first document... **with** the ... replacement document"?

nathan-contino-mongo: Agreed, changed to your suggested wording

source/usage-examples/replaceOne.txt
```diff
@@ -0,0 +1,46 @@
+==================
+Replace a Document
+==================
+
+.. default-domain:: mongodb
+
+You can replace a single document using the `replaceOne()
+<https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#replaceOne>`_
+method. ``replaceOne()`` accepts a query document and a replacement
+document, and replaces the first document that matches the filter with
```
ccho-mongodb: Nitpick: I wonder if it would be more clear to use either query or filter rather than interchangeably within the same section without explaining that they are equivalent since this could be new terminology to a Jasmine persona. Also make sure the term you choose syncs with the terminology/variable naming in your code snippet.

source/usage-examples/replaceOne.txt
```diff
@@ -0,0 +1,46 @@
...
+
+You can specify additional options, such as ``upsert``, using the
+optional ``options`` parameter. Configuring the ``upsert`` field to
+``true`` in the options object causes the operation to insert a new
+document if no document matches the query.
+
+The ``replaceOne()`` method returns a `Promise
+<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise>`_
+that resolves to an `updateWriteOpResult
+<https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#~updateWriteOpResult>`_.
```
ccho-mongodb: To me, the sentences that follow would be more clear if you mention that the `updateWriteOpResult` is a document that contains the following fields.

source/usage-examples/replaceOne.txt
```diff
@@ -0,0 +1,46 @@
...
+You can use the ``modifiedCount`` field of this
+object to determine whether or not a document was modified. If the
+operation was configured to upsert a document, the ``upsertedCount``
+field indicates the number of newly inserted documents. Use the
+``upsertedId._id`` field if you need a unique identifier for an upserted
+document.
+
+The ``replaceOne()`` method will throw an exception if an error occurs
+during execution. For example, attempting to update a field to a value
+that would violate unique index rules will throw a
```
ccho-mongodb: I think "attempting... would... will" makes this sentence more difficult to read than it needs to be.
Consider simplifying:
"For example, if you specify a value that violates a unique index rule in your ``replaceOne()`` call, a ``duplicate key error`` is thrown.

source/usage-examples/replaceOne.txt
```diff
@@ -0,0 +1,46 @@
...
+``duplicate key error``.
+
+.. note::
+
+  If your application requires the document after updating,
+  consider using the
```
ccho-mongodb: Nitpick: why not "use the" instead of "consider using the"?

source/usage-examples/replaceOne.txt
```diff
@@ -0,0 +1,46 @@
...
+  `collection.findOneAndReplace() <https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#findOneAndReplace>`_.
```
ccho-mongodb: I think there's a spare "." since you continue the sentence on the next line.

source/usage-examples/replaceOne.txt
```diff
@@ -0,0 +1,46 @@
...
+  method, which has a similar interface to ``replaceOne()`` but also
```
ccho-mongodb: This could be easier to read as two sentences, splitting it at "but also" (and then replacing that with something that would fit the new sentence structure). The second sentence could be used to explain further what you mean by "returns the original or updated document" since the reader might want to know more about when it returns one or the other.

source/usage-examples/replaceOne.txt
```diff
@@ -0,0 +1,46 @@
...
+Replace a Document
+==================
+
+.. default-domain:: mongodb
+
+You can replace a single document using the `replaceOne()
+<https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#replaceOne>`_
+method. ``replaceOne()`` accepts a query document and a replacement
+document, and replaces the first document that matches the filter with
+the provided replacement document.
```
ccho-mongodb: My previous comment may not have been the greatest suggestion. If you use "replaces with", the description probably requires some clarification on whether it replaces the entire document or only the matching fields of the replacement document. Personally, without testing it, I'm not sure which one it is!

source/code-snippets/usage-examples/replaceOne.js
```diff
@@ -0,0 +1,51 @@
...
+    } else {
+      if (result.matchedCount === 1) {
+        console.log("Matched " + result.matchedCount + " documents.");
+      }
+      if (result.modifiedCount === 1) {
+        console.log("Updated one document.");
+      }
+      if (result.upsertedCount === 1) {
+        console.log(
+          "Inserted one new document with an _id of " + result.upsertedId._id,
```
ccho-mongodb: Question: Is this comma necessary/expected?

source/usage-examples/replaceOne.txt
nathan-contino-mongo: Since I've stuck with query in the past, I'm going to use that same term everywhere. Good catch.

source/usage-examples/replaceOne.txt
nathan-contino-mongo: Clarified.

source/usage-examples/replaceOne.txt
nathan-contino-mongo: Good point, rephrased.

source/code-snippets/usage-examples/replaceOne.js
nathan-contino-mongo: An odd artifact of our prettier, it seems. Not much I can do about it.

source/usage-examples/replaceOne.txt
nathan-contino-mongo: Added quite a bit of extra information about this -- you're totally right, there's a lot of vagueness here right now. Let me know what you think of my changes.

source/usage-examples/replaceOne.txt
nathan-contino-mongo: Done, definitely more digestable now.

source/code-snippets/usage-examples/replaceOne.js
ccho-mongodb: I wonder if that is an older version. I was able to run eslint with react-app and prettier configs and it did not suggest/add a comma there.

source/code-snippets/usage-examples/replaceOne.js
```diff
@@ -0,0 +1,51 @@
...
+const client = new MongoClient(uri);
+
+async function run() {
+  try {
+    await client.connect();
+
+    const database = client.db("sample_mflix");
+    const collection = database.collection("movies");
+
+    // create a filter for a movie to update
```
ccho-mongodb: Would it make sense to use the same term (query) as the documentation (on this line and the next)?

nathan-contino-mongo: absolutely. doing this now

source/usage-examples/replaceOne.txt
```diff
@@ -0,0 +1,47 @@
...
+The ``replaceOne()`` method will throw an exception if an error occurs
+during execution. For example, if you specify a value that violates a
+unique index rule, ``replaceOne()`` throws a ``duplicate key error``.
+
+.. note::
+
+  If your application requires the document after updating,
+  use the `collection.findOneAndReplace()
+  <https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#findOneAndReplace>`_
+  method, which has a similar interface to ``replaceOne()``.
```
ccho-mongodb: Nitpick: I don't think you need the ","

nathan-contino-mongo: Removed.

PR #21, opened 2019-12-20
Trailing comma setting adjustment

nathan-contino-mongo:
Our current configuration enforces trailing commas on multi-line constructions. I would suggest that we continue to enforce trailing commas on objects, but omit them from function calls. See my second commit for an example of how trailing commas on function calls can have slightly baffling results.

### DIFF
--- a/source/code-snippets/usage-examples/replaceOne.js
+++ b/source/code-snippets/usage-examples/replaceOne.js
@@ -40,7 +40,7 @@ async function run() {
       }
       if (result.upsertedCount === 1) {
         console.log(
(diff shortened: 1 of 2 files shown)


source/code-snippets/usage-examples/replaceOne.js
```diff
@@ -40,7 +40,7 @@ async function run() {
       }
       if (result.upsertedCount === 1) {
         console.log(
-          "Inserted one new document with an _id of " + result.upsertedId._id,
+          "Inserted one new document with an _id of " + result.upsertedId._id
```
mongomoe: String interpolation?

PR #22, opened 2020-01-02
Docsp 7604 update multiple docs

mongomoe:
JIRA Ticket:
------------
 https://jira.mongodb.org/browse/DOCSP-7604

Stage:
-------
https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker/DOCSP-7604-Update-Multiple-Docs/usage-examples/updateMany


source/code-snippets/usage-examples/updateMany.js
```diff
@@ -0,0 +1,35 @@
...
+  try {
+    await client.connect();
+
+    const database = client.db("sample_mflix");
+    const collection = database.collection("movies");
+
+    // create a filter for movies to update
+    const filter = { rated: "G" };
+    const options = {};
+    options.upsert = true; // create a document if no documents match the filter
```
nathan-contino-mongo: I feel like this example is very similar to the updateOne example. Maybe we could leave out the upsert here since in this use case it doesn't really do anything anyway? Including it could be confusing to readers.

source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,26 @@
...
+Create an `Object
+<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object>`_
+to specify additional options. Set the ``upsert`` option to ``true`` to create a new
+document if no documents match the filter.
+
+The ``updateOne()`` method returns a `Promise
+<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise>`_
+that resolves to an object. The ``modifiedCount`` field of this object
+shows how many documents were modified. The ``upsertedId`` field contains the ``_id``
+of the new document if a document was ``upserted``.
```
nathan-contino-mongo: upserted probably shouldn't be monospaced here because it's not referencing a value in an object or a variable in code directly, and instead just referring to the concept of upserting.

source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,26 @@
...
+
+.. default-domain:: mongodb
+
+You can update multiple documents using the ``updateMany()`` method. 
+``updateMany()`` accepts a filter object, and updates the first document
+that matches the filter using a provided update document. The update
+document requires an :manual:`Update Operator
+</reference/operator/update>` to modify a field in a document.
+
+Create an `Object
```
nathan-contino-mongo: "Create an Object" might be confusing to readers since it doesn't actually tell users how these options are passed into the method. It might be easier to understand if we say something more like "pass the optional options parameter" or "Pass an object as the optional third parameter to ``updateMany()`` to specify additional options".

source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,26 @@
+=====================
+Update Many Documents
+=====================
+
+.. default-domain:: mongodb
+
+You can update multiple documents using the ``updateMany()`` method. 
+``updateMany()`` accepts a filter object, and updates the first document
```
nathan-contino-mongo: I'd be willing to be that updateMany actually updates all documents that match the filter, not just the first

source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,26 @@
...
+that matches the filter using a provided update document. The update
+document requires an :manual:`Update Operator
+</reference/operator/update>` to modify a field in a document.
+
+Create an `Object
+<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object>`_
+to specify additional options. Set the ``upsert`` option to ``true`` to create a new
+document if no documents match the filter.
+
+The ``updateOne()`` method returns a `Promise
```
nathan-contino-mongo: updateOne -> updateMany

source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,26 @@
+=====================
+Update Many Documents
```
nathan-contino-mongo: Changing "Many" to "Multiple" might make this a little more generic and less of a reference discussing updateMany exclusively.

source/code-snippets/usage-examples/updateMany.js
```diff
@@ -0,0 +1,35 @@
...
+const client = new MongoClient(uri);
+
+async function run() {
+  try {
+    await client.connect();
+
+    const database = client.db("sample_mflix");
+    const collection = database.collection("movies");
+
+    // create a filter for movies to update
```
nathan-contino-mongo: Try to make your comments more descriptive of the action being committed here -- this comment applies to a lot of our code samples. Something like "Match only movies with a 'G' Rating" might help readers more when they try to parse this unfamiliar code.

source/code-snippets/usage-examples/updateMany.js
```diff
@@ -0,0 +1,35 @@
...
+    const filter = { rated: "G" };
+    const options = {};
+    options.upsert = true; // create a document if no documents match the filter
+    // create a document that sets the plot of the movie
```
nathan-contino-mongo: Remove this extraneous comment

source/code-snippets/usage-examples/updateMany.js
mongomoe: *  This example is actually similar to 'update one' on purpose
* I'd like to keep some additional options to adhere to the M.O.R (and sometimes .E) paradigm. 
* In my opinion, upsert is the best one for `updateMany` (see: https://mongodb.github.io/node-mongodb-native/3.4/api/Collection.html#updateMany) because we don't want to talk about write concern at the method level and we don't want to bypass-document validation. 
* upsert here does create documents if there are no 'g' rated movies

source/usage-examples/updateMany.txt
mongomoe: Originally it was 'create an additional options object' as part of another PR, but someone suggested this current sentence. Notice, this is the same sentence for this as many of the other usage-examples. I'd like to keep it in the current format for this PR (atleast for the sake of this PR), and then once we do a once over - consider changing the sentence for all of the usage-examples that have it. 

I also, think it's not particularly confusing with the example snippet (especially once we get the 2 column layout)

source/usage-examples/updateMany.txt
mongomoe: Fixed

source/usage-examples/updateMany.txt
mongomoe: Fixed

source/usage-examples/updateMany.txt
mongomoe: You're right - silly mistake of me (probably copy & paste job). Fixed

source/usage-examples/updateMany.txt
mongomoe: fixed

source/code-snippets/usage-examples/updateMany.js
mongomoe: fixed

source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,26 @@
+=========================
+Update Multiple Documents
+=========================
+
+.. default-domain:: mongodb
+
+You can update multiple documents using the ``updateMany()`` method. 
+``updateMany()`` accepts a filter object, and updates the documents
+that matches the filter using a provided update document. The update
```
nathan-contino-mongo: matches -> match

source/code-snippets/usage-examples/updateMany.js
```diff
@@ -0,0 +1,34 @@
...
+  try {
+    await client.connect();
+
+    const database = client.db("sample_mflix");
+    const collection = database.collection("movies");
+
+    // create a filter to update all movies with a 'G' rating
+    const filter = { rated: "G" };
+    const options = {};
+    options.upsert = true; // create a document if no documents have a 'G' rating
```
nathan-contino-mongo: I think my issue with an upsert here is that creating a new document that looks like this:
```
{
   _id: 2,
   rated: 'G',
   num_mflix_comments: 2
}
```
is a slightly strange way for this method to behave on upsert. In reality you wouldn't ever want to create a movie with no title, no summary, etc.

It's not the end of the world, but I'd suggest changing it to a situation where you match a movie by title (where multiple movies share the same title, e.g. "Black Sheep") and update the rating or something. Just because in that case you're at least creating a movie with a title, so it's a little bit more of a valid document than one that just has a rating/num_comments after the upsert.

source/usage-examples/updateMany.txt
```diff
@@ -0,0 +1,26 @@
...
+document requires an :manual:`Update Operator
+</reference/operator/update>` to modify a field in a document.
+
+Create an `Object
+<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object>`_
+to specify additional options. Set the ``upsert`` option to ``true`` to create a new
+document if no documents match the filter.
```
nathan-contino-mongo: I think you should add a sentence here explaining how upsert will create a new document with any literal values provided in your filter, THEN apply your update changes to that new document. It's unintuitive behavior so i think it's valuable to point it out.

source/usage-examples/updateOne.txt
```diff
@@ -8,9 +8,7 @@ You can update a single document using the `updateOne()
 <https://mongodb.github.io/node-mongodb-native/3.3/api/Collection.html#updateOne>`_
 method. ``updateOne()`` accepts a filter document, and updates the first
 document that matches the filter using a provided update document. The
-update document requires an :manual:`Update
-Operator</reference/operator/update/#update-operators>`
-to modify a field in a document.
+update document requires an :manual:`Update Operator </reference/operator/update/#update-operators>` to modify a field in a document.
```
nathan-contino-mongo: while we're updating this, I would suggest changing the uppercased "Update Operator" to "update operator" since it's not actually a proper noun.

source/usage-examples/updateMany.txt
mongomoe: fixed

source/code-snippets/usage-examples/updateMany.js
mongomoe: Fixed - removed all instances of upsert

source/usage-examples/updateMany.txt
mongomoe: Fixed - removed all instances of upsert

source/usage-examples/updateOne.txt
mongomoe: Fixed

source/code-snippets/usage-examples/updateMany.js
```diff
@@ -0,0 +1,33 @@
...
+async function run() {
+  try {
+    await client.connect();
+
+    const database = client.db("sample_mflix");
+    const collection = database.collection("movies");
+
+    // create a filter to update all movies with a 'G' rating
+    const filter = { rated: "G" };
+    const options = {};
```
nathan-contino-mongo: I would suggest omitting options entirely instead of creating/passing an empty object.

source/code-snippets/usage-examples/updateMany.js
```diff
@@ -0,0 +1,33 @@
...
+
+    // increment every document matching the filter with 2 more comments
+    const updateDoc = {
+      $inc: {
+        num_mflix_comments: 2,
+      },
+    };
+    const result = await collection.updateMany(filter, updateDoc, options);
```
nathan-contino-mongo: suggesting to remove options, unless you feel it's valuable to include to illustrate parameters.

source/code-snippets/usage-examples/updateMany.js
mongomoe: Good idea - removed

source/code-snippets/usage-examples/updateMany.js
mongomoe: Good idea - removed

nathan-contino-mongo:
LGTM

PR #25, opened 2020-01-13
Docsp 8054 run a command usage

mongomoe:
Stage:
------
https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-8054-Run-a-Command-Usage/usage-examples/command

JIRA:
-----
https://jira.mongodb.org/browse/DOCSP-8054

### DIFF
--- a/source/usage-examples/command.txt
+++ b/source/usage-examples/command.txt
@@ -1,0 +1,17 @@
+=============
+Run A Command
+=============
+
+.. default-domain:: mongodb
+
+Overview
+--------
+
+You can run a wide-range of diagnostic and administrative tasks using the ``db.commands``
+
+
+Database commands allow you to perform a wide range of diagnostic and administrative
+tasks with the Node.js driver. For example, the
+:manual:`dbStats </reference/command/dbStats/>` command returns
+storage statistics for a given database. Use the ``command`` function to access
+database commands.
--- a/source/usage-examples/command.txt
+++ b/source/usage-examples/command.txt
@@ -7,13 +7,22 @@ Run A Command
 Overview
 --------
 
-You can run all :manual:`non CRUD </crud/>` database operations using the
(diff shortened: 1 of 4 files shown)


source/usage-examples/command.txt
```diff
@@ -0,0 +1,31 @@
+=============
+Run A Command
+=============
+
+.. default-domain:: mongodb
+
+Overview
+--------
+
+You can run all :manual:`non CRUD </crud/>` database operations using the
```
cbush: "non CRUD" in the link feels off, since there should be a hyphen, and the link leads to CRUD, not a definition of "non-CRUD". I'd suggest rewording it to avoid the "non-" part.

source/usage-examples/command.txt
```diff
@@ -0,0 +1,31 @@
...
+--------
+
+You can run all :manual:`non CRUD </crud/>` database operations using the 
+:node-api:`command() </Admin.html#~command>` method. Typically ``command()`` is used for diagnostic and administrative tasks, such as fetching server stats or initializing a replica set. 
+
+.. note::
+    Using :manual:`collection methods </reference/method/js-collection>`
+    instead of raw database commands through the ``command()`` is
+    recommended for beginner users, since ``collection methods`` are
+    wrappers that abstract some configuration.
```
cbush: Can you clarify this? I'm not sure this is coming across so clearly.

source/usage-examples/command.txt
mongomoe: updated to "You can run all raw database operations, not including CRUD operations, using the command() method."

source/usage-examples/command.txt
mongomoe: updated to "Use collection methods instead of raw db commands whenever possible."

PR #27, opened 2020-01-16
chore: improve a couple of code examples

nathan-contino-mongo:
Fixed a couple of comment typos/improved them and switched the find example to use `count()` instead of `hasNext()` which can cause serious issues due to caching.

### DIFF
--- a/source/code-snippets/usage-examples/find.js
+++ b/source/code-snippets/usage-examples/find.js
@@ -32,7 +32,7 @@ async function run() {
       .project(projection);
 
     // print a message if no documents were found
-    if (!(await cursor.hasNext())) {
+    if ((await cursor.count()) == 0) {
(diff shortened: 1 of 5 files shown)


source/code-snippets/usage-examples/find.js
```diff
@@ -32,7 +32,7 @@ async function run() {
       .project(projection);
 
     // print a message if no documents were found
-    if (!(await cursor.hasNext())) {
+    if ((await cursor.count()) == 0) {
```
ccho-mongodb: Should we use `===` here to be clear that we're not casting any types?

nathan-contino-mongo: Great point, `===` is definitely the right choice here even if `==` works to show best practice. Changing for multiple occurrences of this now.

PR #29, opened 2020-01-16
u ain't nothin but a callback

nathan-contino-mongo:
stage: https://docs-mongodborg-staging.corp.mongodb.com/node/docsworker/callbackz/

### DIFF
--- a/source/code-snippets/usage-examples/deleteMany.js
+++ b/source/code-snippets/usage-examples/deleteMany.js
@@ -20,13 +20,8 @@ async function run() {
     // Query for all movies with the title "Santa Claus"
     const query = { title: "Santa Claus" };
 
-    collection.deleteMany(query, options, function(err, r) {
-      if (err) {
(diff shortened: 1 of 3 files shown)


source/code-snippets/usage-examples/deleteMany.js
```diff
@@ -20,13 +20,8 @@ async function run() {
...
     const query = { title: "Santa Claus" };
 
-    collection.deleteMany(query, options, function(err, r) {
-      if (err) {
-        console.log("Error: " + err.errmsg);
-      } else {
-        console.log("Deleted " + r.deletedCount + " documents");
-      }
-    });
+    const result = await collection.deleteMany(query);
```
nlarew: What happened to `options`? Looks like it was already gone in the previous code and we included it on accident?

nathan-contino-mongo: that is precisely what happened

PR #34, opened 2020-01-16
DOCSP-8052: logging and monitoring

ccho-mongodb:
JIRA: https://jira.mongodb.org/browse/DOCSP-8052

Staging:
https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker/DOCSP-8052-logging-and-monitoring/fundamentals/logging

https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker/DOCSP-8052-logging-and-monitoring/fundamentals/monitoring


source/fundamentals/logging.txt
```diff
@@ -0,0 +1,194 @@
...
+     }
+   }
+   run().catch(console.dir);
+
+Filter on a Specific Class
+--------------------------
+
+You can set the Logger to only produce log messages that are reported by
+specific classes. The following example demonstrates how to log messages only
+the ``Db`` class.
```
nathan-contino-mongo: "from only"

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,396 @@
...
+an event is created.
+
+Read this guide if you need to record topology changes in your application or
+want to explore the information provided in these events.
+
+Event Descriptions
+------------------
+
+The following table lists all the SDAM events that are made available to your
+application by the driver:
```
nathan-contino-mongo: This table feels very reference-y. I would suggest moving it down lower in the page, towards the very bottom, for this reason. It might be a good idea to give a little more introductory exposition (explain the high level of how you subscribe to these events programmatically, for instance) instead before the first long-form example.

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,396 @@
...
+
+Event Subscription Example
+--------------------------
+
+You can access SDAM events through the driver by subscribing to the event
+in your application. The following example demonstrates connecting to a
+replica set and subscribing to all the SDAM events created by the MongoDB
+deployment:
+
+.. code-block:: js
```
nathan-contino-mongo: Both this example and the first example on the logging page are full programs, runnable aside from the uri. I would suggest moving both of these into separate files in the code-samples directory and literalincluding them (you can deindent to make them look just like a code-block) instead to ease testing when we eventually get to that point. Not a requirement, but a suggestion to make our lives easier in the future.

source/fundamentals/monitoring.txt
```diff
@@ -0,0 +1,396 @@
...
+
+   const { MongoClient } = require("mongodb");
+
+   // Connection URI
+   const uri = "mongodb+srv://<clusterUrl>/?replicaSet=rs";
```
nathan-contino-mongo: I would suggest copying/pasting the exact same UI/comment we use in the usage examples for consistency.

source/fundamentals/monitoring.txt
nathan-contino-mongo: On second thought, I think it would fit perfectly after the first example but before "Example event documents"

source/fundamentals/logging.txt
ccho-mongodb: Good catch, will fix.

source/fundamentals/monitoring.txt
ccho-mongodb: Good idea, I'll review and update.

source/fundamentals/monitoring.txt
ccho-mongodb: Sure, although I had noticed them in your CRUD guide, I was not certain we were doing this for all code snippets. I'll have to review what I have for the Authentication section since I did not include the code there either.

source/fundamentals/monitoring.txt
ccho-mongodb: Thanks, will fix.

source/fundamentals/monitoring.txt
ccho-mongodb: I wonder if I should just show only one generic example of subscribing to an event since they haven't been introduced in the "Event Subscription Example" section which would be first in order if I made the change. I'll ask the team for their opinions.

ccho-mongodb: Decided to produce a generic example to support the reordering.

PR #38, opened 2020-01-17
Docsp 7905 change streams

mongomoe:
JIRA:
-----
https://jira.mongodb.org/browse/DOCSP-7905

Stage:
------
https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-7905-ChangeStreams/usage-examples/changeStream


source/code-snippets/usage-examples/changeStream.js
```diff
@@ -0,0 +1,31 @@
...
+
+    const database = client.db("sample_mflix");
+    const collection = database.collection("movies");
+
+    // create a change stream on the movies collection
+    const changeStream = await collection.watch();
+    // use the '.next()' iterator to see the change events
+    const next = await changeStream.next();
+    console.log("a change to the collection happened: \t", next);
+    // process the document ...
```
ccho-mongodb: I wonder if "Add your logic to process the document here" might be more clear than "// process the document ...". The latter could be confused for an unfinished thought.

source/code-snippets/usage-examples/changeStream.js
```diff
@@ -0,0 +1,31 @@
...
+async function run() {
+  try {
+    await client.connect();
+
+    const database = client.db("sample_mflix");
+    const collection = database.collection("movies");
+
+    // create a change stream on the movies collection
+    const changeStream = await collection.watch();
+    // use the '.next()' iterator to see the change events
```
ccho-mongodb: I think the current comment might make sense to someone more familiar with how it works, but what do you think about making it even more clear to a wider audience? How about: "use the next() function on the changeStream iterator to access the next event"

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,23 @@
+=================
+Watch for Changes
+=================
+
+.. default-domain:: mongodb
+
+You can watch for changes in MongoDB by using the `watch()
```
ccho-mongodb: Could you specify what targets you can watch() either in this sentence or elsewhere in the guide? I wasn't able to find this info in your description and had to reference the server manual:
https://docs.mongodb.com/manual/changeStreams/#watch-collection-database-deployment

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,23 @@
...
+<https://mongodb.github.io/node-mongodb-native/3.5/api/Collection.html#watch>`_ method. ``watch()`` opens a :manual:`Change Stream</changeStreams/>`, that allows applications to watch for changes to data and react to those changes. After opening a ``Change Stream``, you can process the change events using the `changeStream.next()
```
ccho-mongodb: Could you use the `node-api` role here and the other locations in which you're linking to the Node API docs to make it easier to update? See:
https://github.com/mongodb/snooty-parser/blob/master/snooty/rstspec.toml#L534 for the role info.

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,23 @@
+=================
```
ccho-mongodb: Nitpick: the line length is much longer than the 80 characters that our team wants to limit our docs to when possible. Could you hard wrap the long lines when possible?

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,23 @@
...
+Watch for Changes
+=================
+
+.. default-domain:: mongodb
+
+You can watch for changes in MongoDB by using the `watch()
+<https://mongodb.github.io/node-mongodb-native/3.5/api/Collection.html#watch>`_ method. ``watch()`` opens a :manual:`Change Stream</changeStreams/>`, that allows applications to watch for changes to data and react to those changes. After opening a ``Change Stream``, you can process the change events using the `changeStream.next()
+<http://mongodb.github.io/node-mongodb-native/3.5/api/ChangeStream.html#next>`_ method. ``watch()`` takes an optional ``pipeline`` parameter that is an array of :manual:`stages </reference/operator/aggregation-pipeline/#stages>` to filter the change events output. You can close a change stream
```
ccho-mongodb: I wonder if it might be more clear to define what it means to close a change stream here. For example:
"You can stop watching for changes and free the related resources by closing the change stream using the ..."

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,23 @@
...
+You can watch for changes in MongoDB by using the `watch()
+<https://mongodb.github.io/node-mongodb-native/3.5/api/Collection.html#watch>`_ method. ``watch()`` opens a :manual:`Change Stream</changeStreams/>`, that allows applications to watch for changes to data and react to those changes. After opening a ``Change Stream``, you can process the change events using the `changeStream.next()
+<http://mongodb.github.io/node-mongodb-native/3.5/api/ChangeStream.html#next>`_ method. ``watch()`` takes an optional ``pipeline`` parameter that is an array of :manual:`stages </reference/operator/aggregation-pipeline/#stages>` to filter the change events output. You can close a change stream 
+using the `changeStream.close()
+<https://mongodb.github.io/node-mongodb-native/3.5/api/ChangeStream.html#close>`_ method.
+
+Create an `Object
+<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object>`_
+to specify additional options. Set the ``fullDocument`` option to the
+string ``updateLookup`` to include the changes to the document as well as
```
ccho-mongodb: Would clarify here:
Set ... to ... in order to receive an event that contains changes to the document as well as ..."

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,23 @@
...
+the entire altered document.
+
+The ``watch()`` command returns a ``Change Stream`` instance. The ``changeStream.next()`` method returns a promise that resolves to an object with the change events. The ``documentKey`` field of this object is an object that contains the _id of the changed document.
```
ccho-mongodb: Nitpick: Let's be consistent with capitalization of the change stream term whether capitalized. This should also apply to the term "change event" and "object".

ccho-mongodb: Nitpick: could you use rst preformatted for "_id" just for clarity that it's a specific field?

mongomoe: Fixed: 
* Made capitalization consistent for 'Change Stream'
* Formatted _id to ``_id``

source/usage-examples/changeStream.txt
mongomoe: Altered your suggestion slightly to:

Set ``fullDocument`` to ``updateLookup`` to receive an event that contains the changes to the document as well as the entire altered document.

source/usage-examples/changeStream.txt
mongomoe: I've altered your suggestion slightly to: 

"You can stop watching for changes and free resources by closing the change stream using the..."

I think that makes it very clear to a Jasmine-level reader

source/usage-examples/changeStream.txt
mongomoe: Fixed

source/usage-examples/changeStream.txt
mongomoe: Fixed

source/usage-examples/changeStream.txt
mongomoe: Changed to:

"You can watch for changes in MongoDB, such as changes to a collection, database, or deployment, by using the watch() method."

source/code-snippets/usage-examples/changeStream.js
mongomoe: Fixed

source/code-snippets/usage-examples/changeStream.js
mongomoe: Fixed

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,35 @@
...
+
+.. default-domain:: mongodb
+
+You can watch for changes in MongoDB, such as changes to a collection,
+database, or deployment, by using the :node-api:`watch()
+</Collection.html#watch>` method. ``watch()`` opens a :manual:`Change
+Stream</changeStreams/>`, that allows applications to watch for changes
+to data and react to those changes. After opening a ``Change Stream``,
+you can process the change events using the
+:node-api:`changeStream.next() </ChangeStream.html#next>` method.
```
ccho-mongodb: Nitpick: Here, I think you need to identify what `changeStream` is (an instance of a Change Stream) since it is not implied just by the name. You can refer to it as "the ``next()`` method on an instance of a Change Stream."

source/code-snippets/usage-examples/changeStream.js
```diff
@@ -0,0 +1,31 @@
...
+async function run() {
+  try {
+    await client.connect();
+
+    const database = client.db("sample_mflix");
+    const collection = database.collection("movies");
+
+    // create a Change Stream on the movies collection
+    const changeStream = await collection.watch();
+    // use the next() function on the changeStream iterator to access the next event
```
ccho-mongodb: Nitpick: I'm picking on my own comment about your comment, but we should probably use "method" instead of "function" to stay consistent.

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,35 @@
+=================
+Watch for Changes
+=================
+
+.. default-domain:: mongodb
+
+You can watch for changes in MongoDB, such as changes to a collection,
```
ccho-mongodb: Thanks for adding specificity here. While it's nearly there, I think the paragraph could flow more smoothly. What do you think about reordering the sentences to the following:

-Explain what Change Streams are and what they are and what they can be opened on
-Explain how to open a Change Stream using NodeJS
-Explain how to iterate through the events in a Change Stream using NodeJS
-Explain how to filter events using NodeJS
-Explain how to close Change Streams using NodeJS

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,35 @@
...
+free resources by closing the ``Change Stream`` using the
+:node-api:`changeStream.close() </ChangeStream.html#close>` method.
+
+Create an `Object
+<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object>`_
+to specify additional options. Set ``fullDocument`` to ``'updateLookup'``
+to receive an event that contains the changes to the document as well as
+the entire altered document.
+
+The ``watch()`` command returns a ``Change Stream`` instance. The
```
ccho-mongodb: Isn't this method explained in the initial paragraph? Do you think the detail belongs in the initial paragraph, or that the method details should be omitted from there so that you can explain them in later paragraphs?

source/usage-examples/changeStream.txt
mongomoe: Hm, it looks like my current paragraph almost follows that structure. One thing I don't agree with is: "-Explain what Change Streams are and what they are and what they can be opened on". I believe we should not explain what a change stream is, that goes out of the scope of this usage example, and is likely too complex for jasmine. We only want to give an opinionated way to use the `watch()` method for a usage example.

That being said, what do you think about the following paragraph:

You can watch for changes in MongoDB by using the watch() method. Calling the ``watch()`` method on a collection, database, or deployment opens a ``Change Stream`` that allows applications to watch for changes to data and react to those changes. After opening a Change Stream, you can iterate through the change events using the next() method on an instance of a Change Stream. ``watch()`` takes an optional pipeline parameter that is an array of stages to filter the change events output. You can stop watching for changes and free resources by closing the Change Stream using the close() method on an instance of a Change Stream.

It follows the following structure:
- Explains what action the ``watch()`` method does ("You can do <some action that the usage-example method does>" is the structure most usage examples follow)
-Explain how to open a Change Stream using NodeJS
-Explain how to iterate through the events in a Change Stream using NodeJS
-Explain how to filter events using NodeJS
-Explain how to close Change Streams using NodeJS

source/usage-examples/changeStream.txt
mongomoe: In the initial paragraph, we explain that it opens a Change Stream instance. So it is similar, this paragraph is here to illustrate the return type. This usage example follows the M.O.R.E format.

source/code-snippets/usage-examples/changeStream.js
mongomoe: fair enough - fixed

source/usage-examples/changeStream.txt
mongomoe: Fixed to:
After opening a ``Change Stream``,
you can process the change events using the
`next()` method on an instance of a ``Change Stream``

source/usage-examples/changeStream.txt
ccho-mongodb: > "-Explain what Change Streams are and what they are and what they can be opened on". I believe we should not explain what a change stream is, that goes out of the scope of this usage example, and is likely too complex for jasmine. 

You already explain this, I am just reordering the content you have. If I take your revised content and paraphrase:

"A Change Stream allows applications to watch for changes to data and react to those changes. <Some text that connects change events that to change streams>".

The reason I think this explanation belongs first is because the reader may not understand what the concept is, but you're introducing syntax which might be more suitable for a Usage Example (in which we assume the reader already knows the concept and we're just explaining the syntax).

ccho-mongodb: Sorry, I guess this is a Usage Example. However, the sentence structure does not flow smoothly yet.

source/usage-examples/changeStream.txt
ccho-mongodb: Ok, I wonder if you can make this distinction more clear to the reader. Perhaps:
"When the watch() command is successful, it returns an instance of the Change Stream that it opened."

ccho-mongodb: Now that I see this is a usage example, I am not sure the strict MORE format makes sense since you're describing more than a single method whereas all the prior ones describe a single method. There's mention of all the following in the first paragraph: `watch()`, `next(), and `close()` (Methods). Then, you've listed Options without specifying which ,method they belong to. Then, you list Return types for both `watch()` and `next()`.
Instead, maybe we can divide this into two sections that follows the same MORE format, but one function for each. The first section can talk about opening a Change Stream and have its own Method, Options, Return, Example, and then the second can be about iterating through them and also have it's own Method, Options, Return, and Example sections.

source/usage-examples/changeStream.txt
ccho-mongodb: Perhaps you can omit the first part of the sentence to prevent repetition of a proper noun, or replace the second instance with "it".
At this point in the guide, it may not be clear that opening a Change Stream results in an instance since you're mentioning what it returns only on line 26 in the "Return" section.

mongomoe:
Closing because of the large change in the document structure, please refer to the PR: https://github.com/mongodb/docs-node/pull/52

PR #43, opened 2020-01-21
DOCSP-8454: Migrate TLS Options to docs-node

ccho-mongodb:
JIRA:
https://jira.mongodb.org/browse/DOCSP-8454

Staging:
https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker/DOCSP-8454-TLS-page/fundamentals/authentication/mechanisms

https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker/DOCSP-8454-TLS-page/fundamentals/connection


ccho-mongodb:
Moved code examples from authentication/mechanisms.txt into separate files. Added table to show tls options: 
https://github.com/mongodb/docs-node/pull/43/files#diff-794b7f6ba48b300846b88f306f274cfa

ccho-mongodb:
> ... seems to imply it can only be used with X.509.

Do you have a suggestion on how to present it in order to show that you can enable TLS without using X.509 authn?
Perhaps we need another section to set up the context for it? I wonder if we should create a "Getting Started > Connection Guide" and bring back some parts of https://github.com/mongodb/node-mongodb-native/blob/master/docs/guide/connection-settings.txt ; specifically, how to connect to a single instance, important options not covered in the manual (include tls here), and how to properly connect to a replica set. What do you think?

ccho-mongodb:
This ticket originally addressed providing TLS connection options, but the existing pages seemed insufficient to properly categorize all the options. Therefore, I added a new "connection" guide which includes mention of tls in the "Connection Options" section as well as information on how to set up your connection uri.

snooty.toml
```diff
@@ -1,5 +1,5 @@
 name = "node"
-toc_landing_pages = ["/crud", "/crud/read-operations", "/crud/write-operations"]
+#toc_landing_pages = ["/crud", "/crud/read-operations", "/crud/write-operations"]
```
mongomoe: should this '#' be removed? I think this line was accidentally commented out

source/fundamentals/connection.txt
```diff
@@ -0,0 +1,235 @@
...
+.. code-block::
+
+     mongodb+srv://sample-hostname:27017/?poolSize=20&useUnifiedTopology=true
+
+The first part identifies the connection string format. In the example above,
+we use ``mongodb+srv`` which specifies the :manual:`DNS Seedlist Connection
+Format </reference/connection-string/#dns-seedlist-connection-format>` which
+indicates that the hostname following it corresponds to the DNS SRV record of
+your MongoDB instance or deployment. If your instance of deployment does not
+have a DNS SRV record, use ``mongodb`` to specify the
```
mongomoe: Can we capitalize 'mongodb' to 'MongoDB' for consistency?
this should read:
``If your instance of deployment does not
have a DNS SRV record, use ``MongoDB`` to specify the ... ``

source/code-snippets/authentication/cr.js
```diff
@@ -0,0 +1,33 @@
+// ignored first line
+const { MongoClient } = require("mongodb");
+const fs = require("fs");
+
+// specify the placeholder values for your environment in the following lines
+const username = encodeURIComponent("<username>");
+const password = encodeURIComponent("<password>");
```
mongomoe: is the password variable necessary? It isn't used anywhere in this file

source/code-snippets/connection/srv.js
```diff
@@ -0,0 +1,23 @@
+const { MongoClient } = require("mongodb");
+
+// Connection URI
+const uri =
+  "mongodb+srv://sample-hostname:27017/?poolSize=20&useUnifiedTopology=true";
```
mongomoe: suggestion (not required for lgtm):
perhaps this "sample-hostname" can be replaced with <sample-hostname> or <cluster-url> to be consistent the format we introduced in other code snippets. For example:

``const uri =
  "mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&w=majority";`` is the format we've used for other snippets

snooty.toml
ccho-mongodb: It was intentionally commented out. I was debugging a build issue and Andrew Aldridge mentioned this is not yet supported and asked me to comment it out until they officially support it.

source/code-snippets/authentication/cr.js
ccho-mongodb: Good catch. I forgot to add the password variable in the connection uri.

source/fundamentals/connection.txt
ccho-mongodb: Are you sure that would be consistent? If anyone has been using `MongoDB://` in the connection URI, we should probably change it to use the same lowercase syntax as in the specification:
https://github.com/mongodb/specifications/blob/master/source/initial-dns-seedlist-discovery/initial-dns-seedlist-discovery.rst#seedlist-discovery

source/code-snippets/connection/srv.js
ccho-mongodb: This snippet is slightly different in that I was directly referencing sample values in the explanation here: https://github.com/mongodb/docs-node/pull/43/files#diff-14c50d2a0fdcbcb3c3aaa64d68a8cc94R50
I thought it might be strange in this context:

> The next part of the connection string specifies the hostname or IP address and
port of your MongoDB instance. In the example above, we use ``sample-hostname``
as the hostname and ``27017`` as the port. Replace these values to point to
your MongoDB instance.

I could see something like the following if you think it's important for this to also be consistent:

> The next part of the connection string specifies the hostname or IP address and
port of your MongoDB instance. In the example above, we use ``sample-hostname``
as the hostname and ``27017`` as the port. Replace the `<cluster-url>` with your hostname and port, separated with a colon (e.g. `sample-hostname:27017`).

snooty.toml
mongomoe: Ah I understand, thanks for the clarification

source/fundamentals/connection.txt
mongomoe: Sorry I misinterpreted this sentence - you're right, it would not be `MongoDB://` it should be `mongodb://` as it currently is.

source/code-snippets/authentication/cr.js
mongomoe: great!

source/code-snippets/connection/srv.js
mongomoe: I do like your suggested paragraph: 
`The next part of the connection string specifies the hostname or IP address and
port of your MongoDB instance. In the example above, we use sample-hostname
as the hostname and 27017 as the port. Replace the <cluster-url> with your hostname and port, separated with a colon (e.g. sample-hostname:27017).`  

It's much more clear than " Replace these values to point to
your MongoDB instance." 

Though I won't require the paragraph change for LGTM, I think it's more personal preference at this point - I get why the code snippet itself doesn't need to include `<cluster-url>`

ccho-mongodb: Replaced with a diagram that labels each part

PR #44, opened 2020-01-23
DOCSP-8055 adds page about indexes

schmalliso:
Obviously, I'll squash my commits into something less ugly when we're done 😺 

Staged: https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker/DOCSP-8055-index-tutorial/fundamentals/indexes

Jira: https://jira.mongodb.org/browse/DOCSP-8055

### DIFF
--- a/source/code-snippets/indexes/compound.js
+++ b/source/code-snippets/indexes/compound.js
@@ -1,0 +1,35 @@
+const { MongoClient } = require("mongodb");
+
+// Replace the following with your MongoDB deployment's connection
+// string.
+const uri =
+  "mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&w=majority";
+
+const client = new MongoClient(uri, {
+  useNewUrlParser: true,
+  useUnifiedTopology: true,
+});
+
+async function run() {
+  try {
+    await client.connect();
+
+    // begin-ex
+    //
+    // (Connection code omitted for space.)
+    //
+    // Specify the "sample_mflix" database.
+    const mflix = client.db("sample_mflix");
+
+    // Create an ascending index on the "type" and "genre" fields
+    // in the "movies" collection.
+    const result = await mflix
+      .collection("movies")
+      .createIndex({ type: 1, genre: 1 });
+    console.log("Index " + result + " created.");
+    // end-example
+  } finally {
+    await client.close();
+  }
+}
+run().catch(console.dir);
--- a/source/fundamentals/indexes.txt
+++ b/source/fundamentals/indexes.txt
@@ -1,0 +1,194 @@
+=======
+Indexes
+=======
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol   
+
+Overview
+--------
+
+Indexes support the efficient execution of queries in MongoDB. Without
+indexes, MongoDB must scan *every* document in a collection to find the
+documents that match each query. Collection scans are slow and can
+negatively affect the performance of your application. When an
+appropriate index exists for a query, MongoDB uses that index to limit
+the number of documents scanned, thus improving query performance. 
+
+The :manual:`indexes section </indexes>` of the MongoDB manual provides a
+detailed overview of indexes in MongoDB.
+
+Query Coverage and Performance
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When you execute a query against MongoDB, your query can include three
+parts:
+
+- query criteria that specify field(s) and value(s) you are looking for,
+- options that affect the query's execution (e.g. read concern),
+
+- (optional) projection criteria to specify the fields MongoDB should
+  return
+
+When all the fields specified in the query criteria and projection of a
+query are indexed, MongoDB returns results directly from the index
+without scanning any documents or bringing documents into memory. Queries
+that use indexes are more performant than queries that rely on
+collection scans.
+
+The MongoDB manual addresses both :manual:`query coverage
+</core/query-optimization/#read-operations-covered-query>` and
+:manual:`index intersection </core/index-intersection>` (using numerous
+indexes to fulfil a query) in more detail.
+
+Operational Considerations
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To improve query performance, you should build indexes on fields that
+appear often in queries and for all operations that involve sorted
+results. Each index that you add consumes disk space and memory when
+active so you should track index memory and disk usage for capacity
+planning. In addition, when a write operation updates an indexed field,
+MongoDB also has to update the index. Familiarize yourself with the :manual:`Indexing Strategies </applications/indexes>` and :manual:`Data Modeling and Indexes </core/data-model-operations/#data-model-indexes>` documentation when designing your data model and choosing the indexes appropriate for your application.
+
+Index Types
+-----------
+
+MongoDB supports a number of different index types to support querying
+your data. The following sections describe the most common index types
+and provide sample code that creates each index type.
+
+Single Field and Compound Indexes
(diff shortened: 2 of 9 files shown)


schmalliso:
@terakilobyte New and improved with Promises!

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,184 @@
...
+
+Operational Considerations
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To improve query performance, you should build indexes on fields that
+appear often in queries and for all operations that involve sorted
+results. Each index that you add consumes disk space and memory
+when active so you should track index
+memory and disk usage for capacity planning. Adding indexes also has a
+negative performance impact for write operations. When designing your data model and choosing the indexes your application requires, you must weigh the benefits and impacts of each index that you add.
```
nathan-contino-mongo: I personally feel like clarifying the "why" for the negative performance impact upon writes might be helpful here, especially for users who are reading about indexes for the first time/aren't very familiar with them. Maybe linking this to the idea that indexes are good for values that are rarely written/updated, and mostly read?

nathan-contino-mongo: (especially since the write performance hit is small compared to the potential read performance gains, which can be multiple orders of magnitude)

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,184 @@
...
+-----------
+
+MongoDB supports a number of different index types to support querying
+your data. The following sections describe the most common index types
+and provide sample code that creates each index type.
+
+Single Field and Compound Indexes
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Single field indexes are user-defined ascending or descending indexes on
```
nathan-contino-mongo: linking to a definition of ascending or descending might help here -- otherwise I think we might be better off without it, since the direction doesn't matter in 99% of cases.

nathan-contino-mongo: Even though you mention it below, it's still kind of confusing in my opinion because *why* an index needs a direction is unclear.

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,184 @@
...
+The following example uses the ``createIndex()`` method to create an
+ascending index on the ``title`` field in the ``movies`` collection in
+the ``sample_mflix`` database.
+
+.. literalinclude:: /code-snippets/indexes/single-field.js
+   :language: js
+   :start-after: begin-ex
+   :end-before: end-ex
+
+Compound indexes are user-defined indexes on *multiple* fields. When you create a compound index, you specify the direction (i.e. ascending or descending) for each field in the index.
```
nathan-contino-mongo: Something along the lines of "a compound index organizes values first by field1, then by field2, etc. etc." might be useful to add in here as well without getting too low-level.

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,184 @@
...
+   :language: js
+   :start-after: begin-ex
+   :end-before: end-ex
+
+Multikey Indexes (Indexes on Array Fields)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Multikey indexes are user-defined ascending or descending indexes on
+fields that contain an array value. You define a multikey index in the
+same fashion as you would a single and compound indexes: you do not
```
nathan-contino-mongo: "a single" -> single

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,184 @@
...
+Multikey indexes behave differently from non-multikey indexes in terms of
+query coverage, index bound computation, and sort behavior. For a full
+explanation of multikey indexes, including a discussion of their behavior
+and limitations, refer to the :manual:`Multikey Indexes page
+</core/index-multikey>` in the MongoDB manual.
+
+Text Indexes
+~~~~~~~~~~~~
+
+:manual:`Text indexes </core/index-text>` support text search queries on
```
nathan-contino-mongo: Could you link to the text search page of the crud guide here?

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,184 @@
...
+   :language: js
+   :start-after: begin-ex
+   :end-before: end-ex
+
+For a full explanation of text search with MongoDB, refer to :manual:`Text Indexes </core/index-text>` in the MongoDB manual.
+
+Geospatial Indexes
+~~~~~~~~~~~~~~~~~~
+
+MongoDB supports queries of geospatial coordinate data using :manual:`2dsphere </core/2dsphere>` indexes. With a ``2dsphere`` index, you can query for inclusion, intersection, and proximity. The :manual:`Geospatial Queries page </geospatial-queries>` in the MongoDB manual provides a detailed explanation of geospatial query operations in MongoDB.
```
nathan-contino-mongo: could you link to the geo page of the crud guide here?

nathan-contino-mongo: technically geospatial queries can use either ``2d`` **or** ``2dsphere`` indexes -- I feel like it might be a good idea to mention both, even if we only want to provide an example of one. (unfortunately both have distinct use cases, so 2dsphere does not actually replace 2d)

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,184 @@
...
+      "_id" : ObjectId("59a47286cfa9a3a73e51e75c"),
+      "theaterId" : 104,
+      "location" : {
+         "address" : {
+            "street1" : "5000 W 147th St",
+            "city" : "Hawthorne",
+            "state" : "CA",
+            "zipcode" : "90250"
+         },
+         "geo" : {
```
nathan-contino-mongo: For an example, I feel like we might be better off ditching all the fields besides "geo" or at least simplifying them. Right now I feel like this makes it seem like geoJSON objects have a lot more data than they really do -- maybe we can meet in the middle and more specifically call out the location.geo field before throwing users into this example?

nathan-contino-mongo:
Almost forgot: if you do choose to add links to the text/geo CRUD guide pages, I'd also appreciate it if you backlinked those pages to the relevant index page sections since i'd rather show users to another section of the node docs than link them into the very large and very detailed manual.

source/code-snippets/indexes/compound.js
```diff
@@ -0,0 +1,39 @@
...
+    // (Connection code omitted for space.)
+    //
+    // Specify the "sample_mflix" database.
+    const mflix = client.db("sample_mflix");
+
+    // Create an ascending index on the "type" and "genre" fields
+    // in the "movies" collection.
+    mflix
+      .collection("movies")
+      .createIndex({ type: 1, genre: 1 }, function(error, result) {
```
terakilobyte: Still one example using callbacks

source/fundamentals/indexes.txt
```diff
@@ -0,0 +1,184 @@
...
+indexes to fulfil a query) in more detail.
+
+Operational Considerations
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To improve query performance, you should build indexes on fields that
+appear often in queries and for all operations that involve sorted
+results. Each index that you add consumes disk space and memory
+when active so you should track index
+memory and disk usage for capacity planning. Adding indexes also has a
```
terakilobyte:
```suggestion
memory and disk usage for capacity planning. Adding too many indexes can have a
```

terakilobyte: I'm unsure its worth being entirely accurate here. Technically, all indexes will have some performance impact for writes, but this is kind of a given. I think we should frame the message as "too many".

source/fundamentals/indexes.txt
terakilobyte:
```suggestion
same fashion as you would for both single and compound indexes: you do not 
```

terakilobyte:
@schmalliso Anything I can do to help merge this in?

PR #51, opened 2020-01-30
DOCSP-8676 fix grammar and consistency

ccho-mongodb:
JIRA: https://jira.mongodb.org/browse/DOCSP-8676

Staging: https://docs-mongodborg-staging.corp.mongodb.com/e0d7705/node/docsworker/DOCSP-8676-fix-grammar/


source/fundamentals/connection.txt
```diff
@@ -101,8 +101,8 @@ URI to specify the behavior of the client.
    * - **validateOptions**
      - boolean
      - ``false``
-     - Specifies whether to error when an unknown or incorrect option is
-       provided. If ``false``, the driver produces warnings only.
+     - Specifies whether to error when it is passed an unknown or incorrect
```
nathan-contino-mongo: How about: 
```suggestion
     - Specifies whether to error when passed an unknown or incorrect
```

source/fundamentals/logging.txt
```diff
@@ -56,9 +56,9 @@ level:
 Filter on a Specific Class
 --------------------------
 
-You can set the Logger to only produce log messages that are reported by
-specific classes. The following example demonstrates how to log messages from
-only the ``Db`` class.
+You can set the Logger to only produce log messages generated by specific
+classes. The following example demonstrates how to log messages from the
```
nathan-contino-mongo: I feel like "only" is actually quite valuable here. What do you think?

source/fundamentals/connection.txt
cbush: This is still passive voice, but it's ok to do so to avoid blaming the user.

source/fundamentals/monitoring.txt
```diff
@@ -38,7 +38,7 @@ deployment:
 Event Descriptions
 ------------------
 
-The following table lists all the SDAM events that are made available to your
+The following table lists all the SDAM events that are available to your
```
nathan-contino-mongo: This sentence sounds wrong to me. I would suggest:
```suggestion
The following table lists all the SDAM events made available to your
```

source/fundamentals/monitoring.txt
```diff
@@ -48,24 +48,24 @@ application by the driver:
...
-     - Created when a connection to an instance is closed.
+     - Created when a connection to an instance closes.
 
    * - ``serverDescriptionChanged``
      - Created when an instance state changes (such as from secondary to
        primary).
 
    * - ``topologyOpening``
-     - Created before any instance connections are attempted.
+     - Created prior to attempting a connection to an instance.
```
nathan-contino-mongo:
```suggestion
     - Created prior to establishing a connection to an instance.
```

source/usage-examples/findOne.txt
```diff
@@ -10,23 +10,21 @@ for a single document in a collection with ``collection.findOne()``.
...
+MongoDB matches all documents in the collection. However, the operation
+only returns the first matched document.
 
 You can also define additional query options such as
 :doc:`sort </fundamentals/crud/read-operations/sort>`
 and :doc:`projection </fundamentals/crud/read-operations/project>`
-to configure the returned document. You should specify these options
-through the ``options`` object passed as the second parameter of the
-``findOne`` method. For detailed reference documentation, see
+to configure the returned document. Pass them in the ``options`` object as the
```
nathan-contino-mongo: I think this revision is actually a regression. The new sentence begins with a phrase that doesn't sound right to me. "Pass them in the options object as the..." is just confusing to me. I would suggest changing this back.

source/usage-examples/insertMany.txt
```diff
@@ -4,14 +4,14 @@ Insert Multiple Documents
...
-You can insert multiple documents using the :node-api:`insertMany()
+You can insert multiple documents using the :node-api:`collection.insertMany()
 <Collection.html#insertMany>` method. The ``insertMany()`` takes an array
 of documents to insert into the specified collection.
 
-You can create an :mdn:`Object
-<Web/JavaScript/Reference/Global_Objects/Object>` to specify additional
-options. Specify ``ordered:true`` to prevent inserting the remaining
-documents if the insertion failed for a previous document in the array.
+You can specify additional options in the ``options`` object passed in the
```
nathan-contino-mongo: If you standardize this with the desciption of options used in ``findOne`` I think this would read better. As for now, I think it could use one little tweak: 
```suggestion
You can specify additional options in the ``options`` object passed as the
```

source/usage-examples/replaceOne.txt
```diff
@@ -4,31 +4,31 @@ Replace a Document
...
-document that matches the query with the provided replacement document.
-All fields and values in the original document are removed with the
-exception of the ``_id`` value, which remains the same unless you
-explicitly specify a new value for ``_id`` in the replacement document.
+You can replace a single document using the
+:node-api:`collection.replaceOne() <Collection.html#replaceOne>` method.
+``replaceOne()`` accepts a query document and a replacement document. If
+the query matches a document in the collection, it replaces the first
+match with the replacement document. The operation replaces all fields and
+values in the original document with the replacement document. The value of
```
nathan-contino-mongo: This sentence:
```
The operation replaces all fields and
exception of the ``_id`` value, which remains the same unless you
values in the original document with the replacement document.
```

Seems like a regression compared to:
```
 All fields and values in the original document are removed with the 
exception of the ``_id`` value, which remains the same unless you
 explicitly specify a new value for ``_id`` in the replacement document. 
```

Otherwise it is unclear what you mean by "replaces... with the replacement document."

Alternative: you could rephrase like this:
```suggestion
values in the original document with the fields and values in the replacement document. The value of
```

source/usage-examples/updateMany.txt
```diff
@@ -4,19 +4,22 @@ Update Multiple Documents
...
+You can update multiple documents using the
+:node-api:`updateMany() <Collection.html#updateMany>` method.
+``updateMany()`` accepts a filter object and update document, and
+replaces documents in a collection that match the filter with the update
+document. The update document requires an :manual:`update operator
 </reference/operator/update>` to modify a field in a document.
 
-You can define additional query options using the options object passed
-as the third parameter of the ``updateMany()`` method.
+You can specify additional options in the ``options`` object passed in
```
nathan-contino-mongo:
```suggestion
You can specify additional options in the ``options`` object passed as
```

source/usage-examples/updateOne.txt
```diff
@@ -4,35 +4,38 @@ Update a Document
...
-</reference/operator/update/#update-operators>` to modify a field in a
-document.
-
-Create an :mdn:`Object
-<Web/JavaScript/Reference/Global_Objects/Object>` to specify additional
-options. Set the ``upsert`` option to ``true`` to create a new document
-if no documents match the filter.
+You can update a single document using the :node-api:`collection.updateOne()
+<Collection.html#updateOne>` method. ``updateOne()`` accepts a filter
+document and a replacement document. If the query matches documents in the
```
nathan-contino-mongo: Looks like a copy/paste typo from ``replaceOne``:
```suggestion
document and a update document. If the query matches documents in the
```

source/usage-examples/updateOne.txt
```diff
@@ -4,35 +4,38 @@ Update a Document
...
+collection, it replaces the matches with the replacement document. The
```
nathan-contino-mongo: This sentence probably requires some significant restructuring to work in the context of an update document rather than a replacement document. Something that specifies fields & values, perhaps? At the very least we need to move away from the word "replace" :smile:

source/fundamentals/monitoring.txt
ccho-mongodb: I think the nuance communicated with "attempting" is that it could fail whereas establishing implies that it was successful. I could be more specific and use "attempting to establish a connection" if you think "attempting a connection" sounds awkward. Let me know what you think. @nathan-contino-mongo

nathan-contino-mongo: I do think that "attempting to establish a connection" would sound better, but that's also a good chunk of words, so... "attempting" seems fine to me. Good point!

source/fundamentals/connection.txt
ccho-mongodb: Thanks to both of you for identifying the ambiguity of this sentence. I am going to reword this to specify the object to which an incorrect param is passed:

> Specifies whether to error when the method parameters contain an unknown or incorrect option.

source/fundamentals/monitoring.txt
ccho-mongodb: Updated to be more direct:

> You can subscribe to any of the following SDAM events:

source/usage-examples/updateOne.txt
ccho-mongodb: Makes sense. I'll _update_ it.

source/fundamentals/logging.txt
ccho-mongodb: Ok. I added it to the end to make it slightly less ambiguous what it was modifying.

source/usage-examples/replaceOne.txt
ccho-mongodb: The prior version had passive voice ("... are removed...") and was a compound sentence that presented two new concepts. I wanted to break this apart into one part that references only the functionality of replaceOne without the exception case, and then another that explains the exception case.
I think the wording you suggested implies (to me) that there's a mapping between the documents that selectively replaces field-value pairs. I wanted to emphasize that the entire document is replaced (_id field is an exception) rather than just the specified fields. Is there a better wording that you can think of to accomplish this? @nathan-contino-mongo

nathan-contino-mongo: Here's my idea:

```
You can replace a single document using the :node-api:`collection.replaceOne()
<Collection.html#replaceOne>` method. ``replaceOne()`` accepts a
query document and a replacement document. If the query matches a
document in the collection, it replaces the first document that matches
the query with the provided replacement document. This operation
removes all fields and values in the original document and replaces them
with the fields and values in the replacement document. The value of the
``_id`` field remains the same unless you explicitly specify a new value
for ``_id`` in the replacement document.
```

nathan-contino-mongo: I think my biggest issue is the phrase "all fields and values in the original document with the replacement document" because that makes it sound like every field and value becomes a nested document (at least, that's how it sounds to me). This uses all of the grammar/standardization improvements you suggested while keeping the operation's effect explicit.

ccho-mongodb: Ok, your suggestion works for me. Thanks for typing it out. I had interpreted your original suggestion slightly differently; it seemed like it only described replacement of fields that existed in the original document which would not be correct.

I guess replacement is a concept that might be interpreted/visualized differently regardless of word choice, so it's good that you've described it with two sentences to reinforce the statement.

nathan-contino-mongo: Thanks! I thought about this a lot when I originally wrote this document so I wanted to be sure that we kept that level of explanation. The differences between update and replace aren't necessarily obvious to a newcomer!

ccho-mongodb:
Ok, everything should have been addressed. 
@nathan-contino-mongo please take a look when you get a chance.

PR #52, opened 2020-02-03
Docsp 7905 change streams

mongomoe:
JIRA:
-----
https://jira.mongodb.org/browse/DOCSP-7905

Stage:
------
https://docs-mongodbcom-staging.corp.mongodb.com/b22a4a5/node/docsworker-xlarge/DOCSP-7905-ChangeStreams/usage-examples/changeStream

Old PR for reference:
---------------------
https://github.com/mongodb/docs-node/pull/38


mongomoe:
Note for @terakilobyte and @ccho-mongodb 
The code snippet currently in `source/code-snippets/usage-examples/changeStream.js` is a very slightly altered version Nathan L's suggested code snippet, I propose the following two code snippets for an alternative to the one that is currently here (but if everyone prefers the current snippet we can leave it as is):

An alternate snippet that is mostly similar to the current one, but instead of using `setInterval` with some logic, uses` setTimeout`:
--------------------------------
```
// Replace the following with your MongoDB deployment's connection string.
const uri =
  "mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&w=majority";

let changeStream;
async function run() {
  try {
    await client.connect();
    const collection = database.collection("movies");
    // create a change stream on the movies collection
    changeStream = collection.watch();
    // set up an action on each change event
    changeStream.on("change", next => {
      // process a change event
      console.log("a change to the collection happened: \t", next);
    });
    setTimeout(async () => {
      console.log("inserting");
      await collection.insertOne({ test: "change stream" });
      setTimeout(async () => {
        console.log("deleting");
        await collection.deleteOne({ test: "change stream" });
      });
    }, 1000);
  } finally {
    setTimeout(async () => {
      await changeStream.close();
      await client.close();
    }, 2500);
  }
}
run().catch(console.dir);
```


Alternative snippet that is very different and relies on identifying if some operation has been performed in the event listener callback, and if that operation was performed then close the change stream:
----------------
```
// Replace the following with your MongoDB deployment's connection string.
const uri =
  "mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&w=majority";

let changeStream;
let sampleMovieDocumentId;
async function run() {
  try {
    await client.connect();
    const collection = database.collection("movies");
    // create a change stream on the movies collection
    changeStream = collection.watch();
    // set up an action on each change event
    changeStream.on("change", async next => {
      // process a change event
      console.log("a change to the collection happened: \t", next);

      // record the _id of the movie titled "sample movie document"
      if (
        next.operationType === "insert" &&
        next.fullDocument.title === "sample movie document"
      ) {
        sampleMovieDocumentId = next.documentKey._id;
      }

      // if the movie titled "sample movie document" is deleted, we are done listening for change events to the 'movies' collection
      if (
        next.operationType === "delete" &&
        String(sampleMovieDocumentId) === String(next.documentKey._id)
      ) {
        // close the 'Change Stream'
        await changeStream.close();
      }
    });

    setTimeout(async () => {
      console.log("inserting");
      await collection.insertOne({
        title: "sample movie document"
      });
      setTimeout(async () => {
        console.log("deleting");
        await collection.deleteOne({
          title: "sample movie document"
        });
      }, 500);
    }, 500);
  } finally {
    setTimeout(async () => {
      await client.close();
    }, 2500);
  }
}
run().catch(console.dir);

```

ccho-mongodb:
> Note for @terakilobyte and @ccho-mongodb
> The code snippet currently in `source/code-snippets/usage-examples/changeStream.js` is a very slightly altered version Nathan L's suggested code snippet, I propose the following two code snippets for an alternative to the one that is currently here (but if everyone prefers the current snippet we can leave it as is):

Does the second snippet need timeouts on the insertOne and deleteOne operations? We do not use them in our usage examples, so if we do, we should change those as well.

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,57 @@
+=================
+Watch for Changes
+=================
+
+.. default-domain:: mongodb
+
+Open a Change Stream
+--------------------
+
+You can keep track of changes of data in MongoDB, such as changes to a
```
ccho-mongodb: Nitpick: comma is unnecessary here since what follows it is not an independent clause.

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,57 @@
...
+
+To free up resources once you are done processing the change events,
+call the :node-api:`close() </ChangeStream.html#close>` method to close
+the change stream instance and stop watching for changes.
+
+Example
+-------
+
+The following example opens a ``Change Stream`` on the ``movies``
+collection. We simulate two real-world events, an ``insertOne()`` and a
```
ccho-mongodb: Nitpick: I think "real-world" seems unnecessary here and can be confusing since it's a vague term.

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,57 @@
...
+``deleteOne()`` operation on the collection, that trigger the change event.
+After the two ``change events`` are logged, we close the ``Change Stream``
+and the client.
```
ccho-mongodb: Nitpick: it sounds awkward to "close [a] client". The `close` method on MongoClient closes the connections to the db: https://mongodb.github.io/node-mongodb-native/3.5/api/MongoClient.html#close

ccho-mongodb: Another nitpick: would it seem easier to understand for Jasmine if we were consistent with "close the change stream instance" on lines 42-43 and "close the Change Stream" here?

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,57 @@
...
+``changeStream.on()`` with the string ``'change'`` as the
+first parameter and a :mdn:`callback function
+<Glossary/Callback_function>` as the second parameter. The callback will
+trigger when a ``change event`` is emitted, providing the next available
+document. You can use the callback to specify logic to process the
+document when the event is received. 
+
+To free up resources once you are done processing the change events,
+call the :node-api:`close() </ChangeStream.html#close>` method to close
+the change stream instance and stop watching for changes.
```
ccho-mongodb: Nitpick: This sentence contains two actions ("call close()" and "to close and stop") rather than a single one which makes it less easy to read. Either split this into two sentences or omit one of the actions. E.g.:
1) Call the close() method on the change stream instance to stop processing change events. This method closes the change stream and frees resources.
2) To stop watching for changes, call close() on the change stream instance.

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,57 @@
...
+<https://nodejs.org/api/events.html#events_emitter_on_eventname_listener>`_.
+method on the ``Change Stream`` instance to add a listener function. Call
+``changeStream.on()`` with the string ``'change'`` as the
+first parameter and a :mdn:`callback function
+<Glossary/Callback_function>` as the second parameter. The callback will
+trigger when a ``change event`` is emitted, providing the next available
+document. You can use the callback to specify logic to process the
+document when the event is received. 
+
+To free up resources once you are done processing the change events,
```
ccho-mongodb: Nitpick: avoid idioms such as "free up". You can use "free" in this instance to convey the exact same message.

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,57 @@
...
+Open a Change Stream
+--------------------
+
+You can keep track of changes of data in MongoDB, such as changes to a
+collection, database, or deployment by opening a :manual:`Change
+Stream</changeStreams/>`. A ``Change Stream`` allows applications to
+watch for changes to data and react to those changes. You can open a
+``Change Stream`` by calling the :node-api:`watch()
+</Collection.html#watch>` method on a collection, database, or
+deployment. The ``watch()`` method takes an optional pipeline as the
```
ccho-mongodb: Should this sentence be in the second paragraph since it's a parameter (using the Method-Options-Return-Error structure)?

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,57 @@
...
+first parameter that is an array of stages to filter the :manual:`change
+events </reference/change-events/>` output. 
+
+The ``watch()`` method can take in additional options as the second
+parameter. Create an `Object
+<https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object>`_
```
ccho-mongodb: Nitpick: on Friday, we decided not to link to Object anymore.

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,57 @@
...
+to specify additional options. Set ``fullDocument`` to
+``'updateLookup'`` to receive an event that contains the changes to the
+document as well as the entire altered document.
```
ccho-mongodb: Since there are only two accepted values for `fullDocument`, I wonder if it would be useful to contrast them here. Alternatively, you could reference the API docs with a link and skip the description of the `updateLookup` setting.

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,57 @@
...
+
+Process the Change Stream Events
+--------------------------------
+
+The ``watch()`` command returns a ``Change Stream`` instance. Use the
+`EventEmitter.on()
+<https://nodejs.org/api/events.html#events_emitter_on_eventname_listener>`_.
+method on the ``Change Stream`` instance to add a listener function. Call
```
ccho-mongodb: Nitpick: it could be more clear to users new to these concepts to mention what the methods do before mentioning the names of them. Also, we should try to be consistent with "X command does Y" and "Use X for Y", at least within the same paragraph. Here's an example (combination of both suggestions):
You can capture events from a change stream instance by using a listener function. Call the watch() command to get a Change Stream instance. Add your listener function by calling the EventEmitter.on() method on that instance. Set the value of the first parameter to the string 'change' and add your callback function to the second parameter...

source/code-snippets/usage-examples/changeStream.js
```diff
@@ -0,0 +1,42 @@
...
+
+const client = new MongoClient(uri);
+
+let changeStream, interval;
+async function run() {
+  try {
+    await client.connect();
+    const database = client.db("sample_mflix");
+    const collection = database.collection("movies");
+    // create a change stream on the movies collection
```
ccho-mongodb: Nitpick: I think the terminology you used in the description was "open a change stream" so we should try to stay consistent.

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,57 @@
...
+-------
+
+The following example opens a ``Change Stream`` on the ``movies``
+collection. We simulate two real-world events, an ``insertOne()`` and a
+``deleteOne()`` operation on the collection, that trigger the change event.
+After the two ``change events`` are logged, we close the ``Change Stream``
+and the client. 
+
+
+.. literalinclude:: /code-snippets/usage-examples/changeStream.js
```
ccho-mongodb: The current example contains code that seems to attempt to clean changes made to the db. In our insert, modify, and delete usage examples, we do not provide that logic. There is no guarantee here that the inserted records will be deleted either, so perhaps we don't need two separate operations which would keep the example more concise.  I do not think we need to change the example, but perhaps an admonition might be useful. If you do make changes to the example, you might consider moving the operations into another function since the main focus is operations on the change stream.

source/usage-examples/changeStream.txt
mongomoe: I think it's actually valid the way it's currently written. See: https://www.grammarly.com/blog/such-as-comma/

Screen Shot 2020-02-03 at 5 14 34 PM

source/usage-examples/changeStream.txt
mongomoe: I can agree with that, I've removed that wording

source/usage-examples/changeStream.txt
mongomoe: Altered the sentence to adhere to your suggestions above. Here is the updated sentence:

" After the two change events are logged, we close the Change Stream instance and the connection to the ``db``."

source/usage-examples/changeStream.txt
mongomoe: Thanks for the suggested sentences, I've replaced my original with your first sentence which captures what I intended well and sounds clean.

source/usage-examples/changeStream.txt
mongomoe: Resolving since I've replaced this entire sentence w/ the suggestion above ^

source/usage-examples/changeStream.txt
mongomoe: I can't find that in my notes, was there a reason we wanted to not link to Object anymore? Some of the other usage examples follow a similar structure

mongomoe: Fixed

source/usage-examples/changeStream.txt
mongomoe: Fixed

source/usage-examples/changeStream.txt
mongomoe: How would you feel about the suggested additional sentence after this one: 

`
Alternatively, you can leave  ``fullDocument`` or set it to the string ``'default'`` to only receive the altered document, but not the changes to the document. 
`

source/usage-examples/changeStream.txt
mongomoe: Fixed - almost verbatim

source/code-snippets/usage-examples/changeStream.js
mongomoe: Fixed

source/usage-examples/changeStream.txt
mongomoe: Pending @terakilobyte thoughts on our 3 suggested example code snippets

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,56 @@
+=================
+Watch for Changes
+=================
+
+.. default-domain:: mongodb
+
+Open a Change Stream
+--------------------
+
+You can keep track of changes of data in MongoDB, such as changes to a
```
ccho-mongodb: Nitpick: "changes to data" seems like a more appropriate preposition for data and avoids "x of y of z"

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,56 @@
...
+collection, database, or deployment by opening a :manual:`Change
+Stream</changeStreams/>`. A ``Change Stream`` allows applications to
+watch for changes to data and react to those changes. You can open a
+``Change Stream`` by calling the :node-api:`watch()
+</Collection.html#watch>` method on a collection, database, or
+deployment. 
+
+The ``watch()`` method takes an optional pipeline as the first parameter
+that is an array of stages to filter the :manual:`change events
+</reference/change-events/>` output. Also, ``watch()`` can take in
```
ccho-mongodb: You might consider "can take an additional options object" or "can accept an additional options object". I think "take in" has other meanings which makes it ambiguous.

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,56 @@
...
+additional options object as the second parameter. Set ``fullDocument``
```
ccho-mongodb: Should there be an article (the) before "``fullDocument``"?

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,56 @@
...
+field of the additional options object to ``'updateLookup'`` to receive
+an event that contains the changes to the document as well as the entire
+altered document. Alternatively, you can leave  ``fullDocument`` or set it to the string ``'default'`` to only receive the altered document, but not the changes to the document.
```
ccho-mongodb: I think you may have an incomplete thought: "you can leave ``fullDocument`` or...". I'm not certain what it means to do that -- did you mean leave it blank? Also, could you fix the column wrap when you get the chance?

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,56 @@
...
+Process the Change Stream Events
+--------------------------------
+
+You can capture events from a change stream instance by using a listener
+function. Call the watch() command to get a Change Stream instance. Add
+your listener function by calling the `EventEmitter.on()
+<https://nodejs.org/api/events.html#events_emitter_on_eventname_listener>`_
+method on that instance. Set the value of the first parameter to the
+string ``'change'`` and add your :mdn:`callback function
+<Glossary/Callback_function>` function as the second parameter.  The
```
ccho-mongodb: I think you only want one "function" here, not "callback function function".

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,56 @@
...
+callback will trigger when a change event is emitted, providing the next
```
ccho-mongodb: Nitpick: rather than future tense, use simple present: "The callback triggers". I think we avoid future tense because of it's direct relationship to passive voice.

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,56 @@
...
+available document. You can use the callback to specify logic to process
```
ccho-mongodb: Nitpick: This can be more direct and succinct by omitting the general statement "can use the callback". E.g.
"You can specify logic in the callback to process the event document when it is received."

mongomoe: Simple but clear wording change - thanks, fixed verbatim

source/usage-examples/changeStream.txt
mongomoe: Fixed verbatim

source/usage-examples/changeStream.txt
mongomoe: Fixed

source/usage-examples/changeStream.txt
mongomoe: Changed to - " Alternatively, you can set ``fullDocument`` to the
string ``'default'`` to only receive the altered document, but not the
changes to the document." 

and wrapped

source/usage-examples/changeStream.txt
mongomoe: Yes thanks - fixed

source/usage-examples/changeStream.txt
mongomoe: altered to:
"Also, ``watch()`` can take an additional options object as the second parameter."

source/usage-examples/changeStream.txt
mongomoe: Fixed

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,66 @@
...
+
+Call the :node-api:`close() </ChangeStream.html#close>` method on the
+change stream instance to stop processing change events. This method
+closes the change stream and frees resources.
+
+Example
+-------
+
+The following example opens a ``Change Stream`` on the ``movies``
+collection. We create a ``changeListener`` variable that is listener
```
ccho-mongodb: If you omit the `consts` that never get used (per earlier comment), you could make this more clear by mentioning that you're using `changeStream.on()` to define a listener function which will print change events when they occur. Youv'e already mentioned `EventEmitter.on()` so this could serve as a reminder to the reader.

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,66 @@
...
+function to receive and print change events that occur. 
+
+To emit a change event, we want to perform a change to the collection, such as insertion
```
ccho-mongodb: Nitpick: "To do x, we want to..." sounds less decisive than "To do x, perform a change..."

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,66 @@
...
+with ``insertOne()``. To prevent the ``insertOne()`` from executing
```
ccho-mongodb: The use of "the" before ``insertOne()`` is awkward here. This should be similar to how you referenced it in the prior sentence (omit "the").

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,66 @@
...
+before ``changeListener`` can register, we create a ``firstTimer``
```
ccho-mongodb: Personally, I think it is unnecessary to create the unused variable as well as reference it that way. I think it's a creative idea, but including variables that never get used again is not great programming practice. I think we can omit reference to specific lines/variables and effectively communicate the description.

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,66 @@
...
+variable that uses :mdn:`setTimeout
+<Web/API/WindowOrWorkerGlobalScope/setTimeout>` to wait 1 second. After
+the insertion of the document, the listener should receive this event.
```
ccho-mongodb: "insertion of the document" is passive voice. How about:
"When a document is inserted, the listener receives a corresponding event."

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,66 @@
...
+
+After that, we want to close the ``Change Stream`` instance.  We can do
```
ccho-mongodb: Nitpick: You can omit "want to"  and "can do this by" since the rest of this description is not about building up code, but rather describing what is there. How about:
"We close the ``Change Stream`` instance using ``changeStream.close()`` one second after we insert the document to allow the event to trigger the listener.

source/code-snippets/usage-examples/changeStream.js
```diff
@@ -0,0 +1,44 @@
...
+    const database = client.db("sample_mflix");
+    const collection = database.collection("movies");
+
+    // open a Change Stream on the "movies" collection
+    changeStream = collection.watch();
+
+    // set up a listener when change events are emitted
+    const changeListener = changeStream.on("change", next => {
+      // process any change event
+      console.log("a change to the collection happened: \t", next);
```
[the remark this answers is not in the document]
mongomoe: Fixed

source/code-snippets/usage-examples/changeStream.js
```diff
@@ -0,0 +1,44 @@
...
+    });
+
+    // wrap the setTimeout methods in a new Promise to wait for the timers to run
+    await new Promise(resolve => {
+      // wait for the event listener to register before emitting a change event
+      const outerTimer = setTimeout(async () => {
+        // insert a document to emit an event
```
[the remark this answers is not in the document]
mongomoe: Fixed

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,66 @@
...
+
+Open a Change Stream
+--------------------
+
+You can keep track of changes to data in MongoDB, such as changes to a
+collection, database, or deployment by opening a :manual:`Change
+Stream</changeStreams/>`. A ``Change Stream`` allows applications to
+watch for changes to data and react to those changes. You can open a
+``Change Stream`` by calling the :node-api:`watch()
+</Collection.html#watch>` method on a collection, database, or
```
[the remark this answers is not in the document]
mongomoe: Fixed

source/code-snippets/usage-examples/changeStream.js
```diff
@@ -0,0 +1,44 @@
...
+
+    // set up a listener when change events are emitted
+    const changeListener = changeStream.on("change", next => {
+      // process any change event
+      console.log("a change to the collection happened: \t", next);
+    });
+
+    // wrap the setTimeout methods in a new Promise to wait for the timers to run
+    await new Promise(resolve => {
+      // wait for the event listener to register before emitting a change event
```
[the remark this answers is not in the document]
mongomoe: updated to " wait for the event listener to register before inserting a document". 

I don't think we need to mention that the driver emits the change event in this doc. I'm open to suggestions on alternative phrasing though

source/code-snippets/usage-examples/changeStream.js
```diff
@@ -0,0 +1,43 @@
...
+  try {
+    await client.connect();
+    const database = client.db("sample_mflix");
+    const collection = database.collection("movies");
+
+    // open a Change Stream on the "movies" collection so the driver can emit events
+    changeStream = collection.watch();
+
+    // set up a listener when change events are emitted
+    const changeListener = changeStream.on("change", next => {
```
cbush:
```suggestion
    changeStream.on("change", next => {
```

source/code-snippets/usage-examples/changeStream.js
```diff
@@ -0,0 +1,43 @@
...
+      // process any change event
+      console.log("received a change to the collection: \t", next);
+    });
+
+    // wrap the setTimeout methods in a new Promise to wait for the timers to run
+    await new Promise(resolve => {
+      // wait for the event listener to register before emitting a change event
+      const outerTimer = setTimeout(async () => {
```
cbush:
```suggestion
      setTimeout(async () => {
```

source/code-snippets/usage-examples/changeStream.js
```diff
@@ -0,0 +1,43 @@
...
+        await collection.insertOne({
+          test: "sample movie document",
+        });
+        // wait to close `changeStream` after changeListener receives the event
+        const innerTimer = setTimeout(async () => {
```
cbush:
```suggestion
        setTimeout(async () => {
```

source/code-snippets/usage-examples/changeStream.js
```diff
@@ -0,0 +1,44 @@
...
+    // set up a listener when change events are emitted
+    const changeListener = changeStream.on("change", next => {
+      // process any change event
+      console.log("a change to the collection happened: \t", next);
+    });
+
+    // wrap the setTimeout methods in a new Promise to wait for the timers to run
+    await new Promise(resolve => {
+      // wait for the event listener to register before emitting a change event
+      const outerTimer = setTimeout(async () => {
```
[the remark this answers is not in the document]
mongomoe: Fixed

source/usage-examples/changeStream.txt
mongomoe: I'd like to keep it this way. I think there are some cases where we want to follow the following format:

"<to do x, we want to do y>" 
"<we can do y by performing some action>" 

to prevent the monotony and runny-sentence-ness of "<to do x, we can do y performing some action>"

source/usage-examples/changeStream.txt
mongomoe: Fixed

source/usage-examples/changeStream.txt
mongomoe: Fixed

source/usage-examples/changeStream.txt
mongomoe: Actually my sentence is already active voice, whereas the proposed sentence is not. I still struggle with this, but active voice is when the subject of the sentence performs the action (https://www.grammarly.com/blog/active-vs-passive-voice/)

I'm no English prof but broken down, I believe the following is accurate:
 
In my sentence: "After the insertion of the document, the listener should receive this event." 
- the subject is the listener, and the listener receives the event
- the first half of the sentence is not using passive voice, it refers to an event that occurred on an object (the document) that is not the subject of the sentence.

Whereas for the sentence: "When a document is inserted, the listener receives a corresponding event.":
- both the document and the listener are compound subjects. 
-  "the listener receives a corresponding event" is using active voice since the subject ("listener") performs the action (receiving an event).
- "When a document is inserted" is passive voice because the subject (a document) is being acted on by an outside party ("is inserted")

That being said, I do like your sentence generally, but I prefer the way it's currently written, and don't think changing it should block this PR

source/usage-examples/changeStream.txt
mongomoe: I disagree that the description is about building up code. The description is solely a tool to help readers understand the code through whatever means (whether that is describing the code or talking about how the code is built). 

Some things I like about my current version:
- "After that, we want to close the Change Stream instance. We can do this by creating a second timer to wait for an additional second ..." is more explanatory than "We close the Change Stream instance using ``changeStream.close()`` one second after"
- It's less run-on than the suggestion

I don't think either is right or wrong, in fact, I think both communicate the same thing to a reasonable degree of effectiveness. As the writer of the doc, I'd like to keep it as is.

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,66 @@
...
+closes the change stream and frees resources.
+
+Example
+-------
+
+The following example opens a ``Change Stream`` on the ``movies``
+collection. We create a ``changeListener`` variable that is listener
+function to receive and print change events that occur. 
+
+To emit a change event, we want to perform a change to the collection, such as insertion
```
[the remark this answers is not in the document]
mongomoe: I prefer to keep this as is, see https://github.com/mongodb/docs-node/pull/52#discussion_r377182351 for further explanation

source/usage-examples/changeStream.txt
```diff
@@ -0,0 +1,65 @@
...
+
+You can keep track of changes to data in MongoDB, such as changes to a
+collection, database, or deployment, by opening a :manual:`Change
+Stream</changeStreams/>`. A ``Change Stream`` allows applications to
+watch for changes to data and react to those changes. You can open a
+``Change Stream`` by calling the :node-api:`watch()
+</Collection.html#watch>` method on a collection or the equivalent
+``watch()`` method on a database, or deployment. 
+
+The ``watch()`` method takes an optional pipeline as the first parameter
```
ccho-mongodb: Optional: I think "pipeline" and "array of stages" might be too far apart in this sentence, and "change events output" is a little ambiguous. Consider:

> The ``watch()`` method optionally takes a pipeline, an array of stages, as the first parameter to filter the change events.

source/usage-examples/changeStream.txt
ccho-mongodb: Sorry, you're right that my recommendation is passive voice! I was conflating my recommendation with a sentence containing two independent clauses. My new suggestion is  "When you insert a document, the listener receives the corresponding event". What do you think about that?

The phrase "insertion of the document" sounds awkward, or perhaps the cause-effect is not strongly-associated. Consider a different context with a parallel structure:

Your version: "After the payment of taxes, the government should receive money."

vs

My new suggested version: "When you pay taxes, the government receives the corresponding money."

I prefer the latter, but if you want to stick with the former, that's fine. If I mentioned that this "[blocks] this PR", I didn't mean to as there were several other requested changes.

source/usage-examples/changeStream.txt
ccho-mongodb: In my comment, I mentioned "the description is not about building up code" which I think you agree with. The request was to omit "we can do this by".

source/usage-examples/changeStream.txt
ccho-mongodb: You're saying:
> to prevent the monotony and runny-sentence-ness of "<to do x, we can do y performing some action>"

I said:
> "To do x, we want to..." sounds less decisive than "To do x, perform a change..."

I don't think these are the same. That said, what do you think of what I suggested? I think "we want to" is extraneous because it does not provide any additional information.

mongomoe: Suggestion that omit any mention of "we want":

Example
--------
**The following example opens a ``Change Stream`` on the ``movies`` collection. We create a listener function to receive and print change events that occur.**

**To emit a change event, we perform a change to the collection, such as insertion with ``insertOne()``. To prevent ``insertOne()`` from executing before the listener function can register, we create a timer that uses setTimeout to wait 1 second. After the insertion of the document, the listener should receive this event.**

**After that, we create a second timer to wait an additional second after the insertion of the document to close the ``Change Stream`` instance using ``changeStream.close()``.**

source/usage-examples/changeStream.txt
mongomoe: "When you insert a document, the listener receives the corresponding event" sounds fine to me. See above comment

source/usage-examples/changeStream.txt
mongomoe: Sorry I meant: "I disagree that the description is **not necessarily** about building up code." 

But I'm game to change it to the sugesstion, see https://github.com/mongodb/docs-node/pull/52#discussion_r377282082 for further thoughts

source/usage-examples/changeStream.txt
mongomoe: Fixed almost verbatim to: ```The watch() method optionally takes a pipeline, an array of stages, as the first parameter to filter the `change events</link>`_ output.```

PR #55, opened 2020-02-06
DOCSP-8733: Collation Fundamentals guide

ccho-mongodb:
JIRA: https://jira.mongodb.org/browse/DOCSP-8733

Staging:
https://docs-mongodborg-staging.corp.mongodb.com/b874313/node/docsworker/DOCSP-8733-collation-fundamentals/fundamentals/collations


source/fundamentals/collations.txt
```diff
@@ -0,0 +1,280 @@
...
+collation* is:
+
+::
+
+   cote
+   coté
+   côte
+   côté
+
+If we apply the Canadian French collation, the order is:
```
nathan-contino-mongo: Same here:

```
Canadian French collation sorts the french words in the following order:
```

source/fundamentals/collations.txt
```diff
@@ -0,0 +1,280 @@
...
+::
+
+   cote
+   côte
+   coté
+   côté
+
+.. note::
+
+   Collations are available in MongoDB 3.4 and later.
```
nathan-contino-mongo: This note feels like it might be more suited to the very top of the page. (beneath "on this page"). What do you think?

source/fundamentals/collations.txt
```diff
@@ -0,0 +1,280 @@
...
+
+-----
+Usage
+-----
+
+You can specify a collation when you create a **new collection** or
+**new index**. You can also specify a collation for :doc:`CRUD operations
+</fundamentals/crud>` and aggregations.
```
nathan-contino-mongo: Breaking the categories of collation usage into "new collection or new index" and "crud operations and aggregations" certainly makes sense, but I think it might be better to split them up differently. Something like:
```
You can specify a default collation for a collection when you create a new collection.
When creating an index, you can specify the collation used by that index. You can
also specify a collation when querying MongoDB using CRUD operations and
Aggregations.
```
I think capital "A" Aggregation is what we usually use in other docs (not sure how consistent we are about that in docs-node though).

source/fundamentals/collations.txt
```diff
@@ -0,0 +1,280 @@
...
+different one.
+
+When you create a **new index** with a collation, you specify the sort order
+for operations that use that index. To use the collation in the index, you
+must provide a matching collation in the operation, and the operation must
+use the index. While most index types support collation, the following
+types support only binary comparison:
+- :manual:`text </reference/method/db.collection.createIndex/#options-for-text-indexes>`
+- :manual:`2d </reference/method/db.collection.createIndex/#options-for-2dsphere-indexes>`
+- :manual:`geoHaystack </reference/method/db.collection.createIndex/#options-for-geohaystack-indexes>`
```
nathan-contino-mongo: It looks like your bulleted list here isn't rendering correctly.

source/fundamentals/collations.txt
```diff
@@ -0,0 +1,280 @@
...
+
+-----
+Usage
+-----
+
+You can specify a collation when you create a **new collection** or
+**new index**. You can also specify a collation for :doc:`CRUD operations
+</fundamentals/crud>` and aggregations.
+
+When you create a **new collection** with a collation, you define the default
```
nathan-contino-mongo: Not sure how I feel about bolding "new collection". I feel like we might be better off leaving the bold out since "new collection" and "new index" aren't really concepts. This almost makes me feel like I'm reading java (`new Collection()`).

source/fundamentals/collations.txt
```diff
@@ -0,0 +1,280 @@
...
+types support only binary comparison:
+- :manual:`text </reference/method/db.collection.createIndex/#options-for-text-indexes>`
+- :manual:`2d </reference/method/db.collection.createIndex/#options-for-2dsphere-indexes>`
+- :manual:`geoHaystack </reference/method/db.collection.createIndex/#options-for-geohaystack-indexes>`
+
+.. note::
+
+   Currently, you cannot create a collation on an existing collection. To use
+   collations with an existing collection, create an index with the collation
+   and specify the same collation in your operations on it.
```
nathan-contino-mongo: This note probably flows better beneath the new collection paragraph.

source/fundamentals/collations.txt
```diff
@@ -0,0 +1,280 @@
...
+
+In the following example, we create a new collection called ``souvenirs`` on
+the ``sample_mflix`` database and assign a default collation with the
+"``fr_CA``" locale. The collation applies to all :manual:`operations that
+support collation </reference/collation/#operations-that-support-collation>`
+performed on that collection.
+
+.. literalinclude:: /code-snippets/collation/collection-collation.js
+   :language: javascript
+   :dedent: 4
```
nathan-contino-mongo: It might be a good idea to add an example of a query against this collection that *does not* specify a collation, and point out that such a query uses the default collation of ``fr_CA`` that we specified above. I know when I read through this and hit the index collation example beneath I felt like there was a missing example above for the collection collation (and I like the idea of showing how to define a collation for a collection, then how to use a collation on a collection, then how to define a collation for an index, and then how to use the index based on collation)

source/code-snippets/collation/find-sort-collation.js
```diff
@@ -0,0 +1,29 @@
...
+
+    // print results
+    await cursor.forEach(console.dir);
+  } finally {
+    // Ensures that the client will close when you finish/error
+    await client.close();
+  }
+}
+run().catch(console.dir);
+~
```
nathan-contino-mongo: looks like a sneaky tilde managed to find its way into your example

source/fundamentals/collations.txt
```diff
@@ -0,0 +1,280 @@
...
+
+.. literalinclude:: /code-snippets/collation/find-sort-collation.js
+   :language: javascript
+   :dedent: 4
+
+findOneAndUpdate() Example
+``````````````````````````
+
+The following example calls the ``findOneAndUpdate()`` operation on a
+collection called ``names`` in which contains the following documents:
```
nathan-contino-mongo: It might be helpful to point out that this collection uses the default binary collation since that configuration was just discussed and does influence the behavior in the example.

source/fundamentals/collations.txt
```diff
@@ -0,0 +1,280 @@
...
+which **does not** specify a collation:
+
+.. literalinclude:: /code-snippets/collation/findOneAndUpdate-collation.js
+   :language: javascript
+   :start-after: start findOneAndUpdate without collation
+   :end-before: end findOneAndUpdate without collation
+   :dedent: 4
+
+Since "Gunter" is the first result in the sorted results when using a binary
+collation, the operation does not update any documents. Consider the same
```
nathan-contino-mongo: I'd suggest explaining *why* the Gunter document doesn't update here.

source/fundamentals/collations.txt
```diff
@@ -0,0 +1,280 @@
...
+before the same characters without umlauts.
+
+.. literalinclude:: /code-snippets/collation/findOneAndUpdate-collation.js
+   :language: javascript
+   :start-after: start findOneAndUpdate with collation
+   :end-before: end findOneAndUpdate with collation
+   :dedent: 4
+
+The ``findOneAndUpdate()`` operation above returns the following updated
+document:
```
nathan-contino-mongo: Similarly, I think it's worth explaining why the Gumlauter document *does* update here.

source/code-snippets/collation/findOneAndDelete-collation.js
```diff
@@ -0,0 +1,58 @@
...
+  try {
+    // Connect the client to the server
+    await client.connect();
+
+    const db = client.db("myproject");
+    const collection = db.collection("numbers");
+
+    const result = await collection.findOneAndDelete(
+      { a: { $gt: "100" } },
+      { collation: { locale: "en", numericOrdering: true } },
```
nathan-contino-mongo: If `locale` isn't required here, I'd suggest removing it since it distracts uses from `numericOrdering` which is the point of this example. If it is required, can we at least switch the order of the collation fields so `numericOrdering` comes first?

source/fundamentals/collations.txt
```diff
@@ -0,0 +1,280 @@
...
+
+::
+
+  { "_id" : 1, "a" : "16" }
+  { "_id" : 2, "a" : "84" }
+
+If you perform the same operation without collation on the original
+collection of three documents, the server deletes the first document it
+finds in which the lexical value of "``a``" is greater than the lexical value
+of "``100``".
```
nathan-contino-mongo: Calling out the exact `a` value that is deleted might be helpful here and below.

source/fundamentals/collations.txt
```diff
@@ -0,0 +1,280 @@
...
+   { "_id" : 2, "a" : "84" }
+   { "_id" : 3, "a" : "179" }
+
+
+Aggregation Example
+```````````````````
+
+To use collation with the :node-api:`aggregate </Collection.html#aggregate>`
+operation, pass the collation document in the options field, after the
+array of pipeline stages.
```
nathan-contino-mongo: It might be helpful here to discuss how the collation applies to the aggregation pipeline. I'm assuming that every pipeline stage uses the specified collation, but if that's not true, we might want to point that out specifically here.

source/code-snippets/collation/aggregation-collation.js
```diff
@@ -0,0 +1,33 @@
...
+      ],
+      { collation: { locale: "de@collation=phonebook" } },
+    );
+
+    console.log(await cursor.toArray());
+  } finally {
+    // Ensures that the client will close when you finish/error
+    await client.close();
+  }
+}
```
nathan-contino-mongo: I feel like it's hard to read these examples because they use an entire runnable file, which distracts from the lines that are actually important. Can we trim down the included lines using `:startsafter:` and `:endsbefore:` like I did for the CRUD examples? I'm thinking that we really just need the `collection.<method>` code to get the point across - users can always consult the usage examples for full runnable files.

source/fundamentals/collations.txt
ccho-mongodb: I thought of placing it at the top as well, but it seemed too early on in the document to make an admonition before describing what a collation is. What if I just used some italic text, similar to this pattern used in the server manual: https://docs.mongodb.com/manual/tutorial/configure-encryption/

source/fundamentals/collations.txt
ccho-mongodb: The server manual uses lowercase (unless it links to an article in title case or is at the beginning of a sentence) and `source/fundamentals/crud.txt` too. Which docs use capital "A"?

source/fundamentals/collations.txt
ccho-mongodb: Weird, I had put a linebreak between 68-69 (the reason why it doesn't render correctly) but it's not there anymore. Thanks for catching this!

source/code-snippets/collation/find-sort-collation.js
ccho-mongodb: Thanks for catching this.

source/fundamentals/collations.txt
ccho-mongodb: I guess I need to clarify this. It's not "a", but rather the value assigned to "a" which hopefully is apparent when listing each of the values of "a". 
Perhaps "the lexical value of the "a" field (e.g. ...)" ?

source/fundamentals/collations.txt
ccho-mongodb: I'm not absolutely certain when it applies in the aggregation pipeline, but I know that you can only specify one (and can include that in the docs if you think it's useful to provide that depth). Can you point me to any info on this? I checked the server docs and couldn't find anything.

source/fundamentals/collations.txt
ccho-mongodb: This was an editorial choice I made to emphasize that it must be a "new" collection or index and not an existing one. I'll take away the formatting.

source/fundamentals/collations.txt
ccho-mongodb: Instead of "New in version 3.2." it could read "Collations are available in MongoDB 3.4 and later" to prevent confusion with the driver version. The top of the configure-encryption page is also a mess.

source/code-snippets/collation/findOneAndDelete-collation.js
ccho-mongodb: Good recommendation to shorten the example, but per the documentation, the `locale` field is always required. 😛

ccho-mongodb: Regardless, the order is a great point. I'm not sure there's an advantage with playing with the order. I've always placed required params first in associative arrays/maps followed by optional ones which is similar to parameter ordering conventions in several programming languages.

source/code-snippets/collation/aggregation-collation.js
ccho-mongodb: I'm removing the calls and any additional code that makes it seem like it's being called inside an async function. I'll update the descriptions too since the database and collection names are not necessary.

source/fundamentals/collations.txt
ccho-mongodb: Great feedback, I think the newer version does not force the user to infer anything anymore.

source/fundamentals/collations.txt
ccho-mongodb: I'll specify this in each example since a reader might jump to this point from anywhere

source/fundamentals/collations.txt
nathan-contino-mongo: I like this solution a lot!

source/fundamentals/collations.txt
nathan-contino-mongo: I saw some recent changes to the Stitch docs that explicitly used a capital A -- pinging @nlarew for his perspective, since he can confirm if that's the norm.

source/code-snippets/collation/findOneAndDelete-collation.js
nathan-contino-mongo: Yeah I can definitely understand listing it first -- it's also a parameter that users are already familiar with, so not a problem at all IMO. Leaving it as is seems perfectly fine!

source/fundamentals/collations.txt
nathan-contino-mongo: Your suggested change is definitely in the right direction, but my point was moreso that pointing out the actual deleted value could help readers follow along better -- I felt doing the manual diff (and scrolling up quite a ways to check the contents of the original collection) was a bit of a chore as a reader.

source/fundamentals/collations.txt
nathan-contino-mongo: The docs aren't particularly clear (might be a good ticket to file for the manual) but I'd probably infer based on "You cannot specify multiple collations for an operation." and the fact that collation is a global option for all stages of an aggregation pipeline that the collation applies to all stages. Definitely an area where the docs could be more explicit!

source/fundamentals/collations.txt
```diff
@@ -108,22 +106,43 @@ Collation Examples
...
+
+.. literalinclude:: /code-snippets/collation/collection-collation.js
+   :language: javascript
+   :start-after: start collection query without collation
+   :end-before: end collection query without collation
+   :dedent: 4
+
+You can specify a different collation as a parameter in an operation that
+supports collations. The following query specifies the "``is``" Iceland locale
+and ``caseFirst`` optional parameter with the value "``upper``":
```
nathan-contino-mongo: Glad we're spicing the examples up with some additional options, this is great!

source/fundamentals/collations.txt
```diff
@@ -188,22 +208,27 @@ which **does not** specify a collation:
    :end-before: end findOneAndUpdate without collation
    :dedent: 4
 
-Since "Gunter" is the first result in the sorted results when using a binary
-collation, the operation does not update any documents. Consider the same
-operation with a collation specified with the locale set to
+Since "Gunter" is the first sorted result when using a binary collation, none
+of the documents come lexically before and match the ``$lt`` comparison
+operator in the query document. As a result, the operation does not update any
+documents.
```
nathan-contino-mongo: yes, this is exactly what I was hoping to see. Definitely makes the "why" much clearer than the previous version, fantastic job!

source/fundamentals/collations.txt
```diff
@@ -0,0 +1,311 @@
...
+lexical value
+of "``100``".
+
+.. literalinclude:: /code-snippets/collation/findOneAndDelete-collation.js
+   :language: javascript
+   :start-after: start findOneAndDelete example without collation
+   :end-before: end findOneAndDelete example without collation
+   :dedent: 4
+
+Since lexical value of all the values of ``a`` ("``16``", "``84``", and
```
nathan-contino-mongo: I missed this before -- you should probably add a "the" here:
```suggestion
Since the lexical value of all the values of ``a`` ("``16``", "``84``", and
```
(and then change "are greater" to "is greater" below)

source/code-snippets/collation/collection-collation.js
```diff
@@ -0,0 +1,18 @@
...
+
+// start create collection with collation
+// Create the collection with a collation
+db.createCollection("souvenirs", {
+  collation: { locale: "fr_CA" },
+});
+// end create collection with collation
+
+// start collection query without collation
+collection.find({type: "photograph"});
```
nathan-contino-mongo: I think using the collection name itself (for instance, `souvenirs.find` here, `names.find` for the last example) might make it a little easier for readers to follow since they don't have all the extra context now of the db/collection object instantiation now. Not sure if it's necessary, though.

ccho-mongodb: One thing we may need to unify is the usage example code snippet and CRUD guide conventions for variable naming of the db and connection. In all the usage examples we use the following convention:
 ``` 
const db = client.db('sample_mflix');
const collection = db.collection('movies');
```
In your CRUD guide examples, I see a mix:
```
const collection = database.collection("movies");
```
and
```
const database = client.db("test");
const orders = database.collection("orders");
```

I don't think there's a good answer, but we should be consistent. I'll leave this as-is for now and make a mental note for later.

source/fundamentals/collations.txt
ccho-mongodb: I think the wording of these two paragraphs is still awkward. I'm updating it to provide better clarity.

source/code-snippets/collation/collection-collation.js
nathan-contino-mongo: Really good point -- we'll table the discussion for later. I'll create a ticket for this -- is there a "Node Enhancements" epic I can throw it in?

ccho-mongodb: https://jira.mongodb.org/browse/DOCSP-8820 @nathan-contino-mongo

PR #59, opened 2020-02-07
Moe client code consistent

mongomoe:
Changes:
----------
* Made code snippets for client connection consistent to look like this:
```
// Replace the following string with your MongoDB deployment's connection string.
const uri =
  "mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&w=majority&useUnifiedTopology=true";

```
* added note linking users to connection guide to learn how to connect

Stage:
------
https://docs-mongodbcom-staging.corp.mongodb.com/28e28d4/node/docsworker-xlarge/Moe-Client-Code-Consistent


source/code-snippets/usage-examples/bulkWrite-example.js
```diff
@@ -44,7 +40,7 @@ async function run() {
           console.log("Number of documents inserted: " + result.nInserted);
         } else {
           console.log(
-            "No documents were inserted during the bulk write operation"
+            "No documents were inserted during the bulk write operation",
```
nathan-contino-mongo: Nit: This comma could be confusing for newcomers to our documentation - I'd advise removing it (from all examples)
```suggestion
            "No documents were inserted during the bulk write operation"
```

source/code-snippets/usage-examples/bulkWrite-example.js
```diff
@@ -1,15 +1,11 @@
...
-// Replace the following with your MongoDB deployment's connection
-// string.
+// Replace the following string with your MongoDB deployment's connection string.
 const uri =
-  "mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&w=majority";
+  "mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&w=majority&useUnifiedTopology=true";
 
-const client = new MongoClient(uri, {
-  useNewUrlParser: true,
-  useUnifiedTopology: true
```
nathan-contino-mongo: Considering that most of our users will (as our comment suggests) *replace* the URI above, I think adding options to it likely won't have the intended result. If the new url parser is now used by default, I'd suggest leaving the client unified topology option in place. It also helps demonstrate how to pass options to the client, which has a lot of value for a usage example IMO.

source/usage-examples/command.txt
```diff
@@ -23,6 +23,9 @@ The ``command()`` method returns a :mdn:`Promise
 <Web/JavaScript/Reference/Global_Objects/Promise>`. The Promise returns
 the results of the command operation when it completes.
 
+.. include:: /includes/connect-guide-note.rst
+
+
```
nathan-contino-mongo: While we're add it, why don't we add an "Example" heading to each code snippet, since there are some code snippets that have that already (some with exposition). I'd suggest the following formula:
```
Example
------------

my text talking about the example doing exampley things

<note>

<code snippet>
```

Point being: if there's an "example" heading, I think it's a little weird for this note to show up *before* the heading, since that's not usually how we use notes in other places in the docs.

mongomoe: Not against this, but I'd prefer to add this to a separate PR

source/code-snippets/usage-examples/bulkWrite-example.js
mongomoe: I can do that it's just not the purpose of this PR, the only thing added to this line was the comma (which was added by eslint)

mongomoe: Fixed

source/code-snippets/usage-examples/bulkWrite-example.js
mongomoe: Fixed

source/usage-examples/command.txt
mongomoe: Fixed

source/code-snippets/usage-examples/bulkWrite-example.js
mongomoe: Fixed

PR #65, opened 2020-02-11
standardizing change stream example with other usage exs

nathan-contino-mongo:
- added note linking users to the connection guide above the usage example snippet
- updated ``fullDocument`` field documentation to match the definition found [here](https://docs.mongodb.com/manual/reference/method/db.collection.watch/) (subtle differences in interpretation -- this definition matches my own testing)
- limited usage of monospace font to specific coding terms

Staging: https://docs-mongodbcom-staging.corp.mongodb.com/ce3554b/node/docsworker/standardization/usage-examples/changeStream

### DIFF
--- a/source/index.txt
+++ b/source/index.txt
@@ -13,7 +13,7 @@ MongoDB Node Driver
   /quick-start
   /fundamentals
   /usage-examples
-  API Documentation <http://mongodb.github.io/node-mongodb-native/3.5/api>
+  API Documentation <http://mongodb.github.io/node-mongodb-native/3.5/api/>
   /faq
   /issues-and-help
   /compatibility
--- a/source/index.txt
+++ b/source/index.txt
@@ -13,7 +13,7 @@ MongoDB Node Driver
   /quick-start
   /fundamentals
   /usage-examples
-  /api-documentation
+  API Documentation :node-api:`API <>`
   /faq
   /issues-and-help
   /compatibility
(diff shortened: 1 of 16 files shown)


source/usage-examples/changeStream.txt
```diff
@@ -9,59 +9,59 @@ Open a Change Stream
...
-``Change Stream`` instance to stop processing ``change events``. This method
-closes the ``Change Stream`` and frees resources.
+``ChangeStream`` instance to stop processing change events. This method
+closes the Change Stream and frees resources.
 
 Example
 -------
 
-The following example opens a ``Change Stream`` on the ``movies``
+The following example opens a Change Stream on the ``movies``
```
ccho-mongodb: I think if we want to use ``ChangeStream`` for the instance and Change Stream to describe the term, it should not be a proper noun in the latter, e.g. "change stream" which has an analog: "change event"

source/index.txt
```diff
@@ -13,7 +13,7 @@ MongoDB Node Driver
   /quick-start
   /fundamentals
   /usage-examples
-  /api-documentation
+  API Documentation <http://mongodb.github.io/node-mongodb-native/3.5/api/>
```
ccho-mongodb: Did this work?

nathan-contino-mongo: yep!

PR #68, opened 2020-02-20
DOCSP-9029: remove retryWrites from uri connection string

ccho-mongodb:
JIRA: https://jira.mongodb.org/browse/DOCSP-9029

Staging: https://docs-mongodbcom-staging.corp.mongodb.com/7496942/node/docsworker/DOCSP-9029-remove-retrywrites/


source/code-snippets/crud/arrayFilters.js
```diff
@@ -4,7 +4,7 @@ const stream = require("stream");
 
 // Replace the following string with your MongoDB deployment's connection string.
 const uri =
-  "mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&w=majority&useUnifiedTopology=true";
+  "mongodb+srv://<user>:<password>@<cluster-url>?w=majority&useUnifiedTopology=true";
```
mbroadst: while we're here, I think you should remove `useUnifiedTopology=true` from the connection string as well. It's likely by the time these docs are released we will will have released v4 of the driver (where this won't be required anymore), or users will have already enabled the flag. Just trying to spare you the pain of having a new PR to modify every one of these strings again.

Also, FWIW the connection strings in this PR aren't consistent in terms of use of this connection string option.

ccho-mongodb: I'll remove all the instances of `useUnifiedTopology=true` and replace the example in fundamentals/connection with another parameter. I think it's ok to feature different connection string parameters across examples for now. Perhaps in the future, we'll have a better way to indicate -- whether implicitly or explicitly -- how to specify parameters in the connection string.

ccho-mongodb:
Connection guide reference to useUnifiedTopology is in this PR: https://github.com/mongodb/docs-node/pull/69

source/code-snippets/crud/arrayFilters.js
ccho-mongodb: Would you mind taking a quick look at the changes I made in response to your comment when you get the chance? @mbroadst

PR #73, opened 2020-02-25
DOCSP-9048: Quick Start user study updates

ccho-mongodb:
JIRA:
https://jira.mongodb.org/browse/DOCSP-9048

Staging:
https://docs-mongodbcom-staging.corp.mongodb.com/1487bd4/node/docsworker-xlarge/DOCSP-9048-quickstart-user-study-updates/quick-start


source/quick-start.txt
```diff
@@ -7,193 +7,182 @@ Quick Start
 .. contents:: On this page
    :local:
    :backlinks: none
-   :depth: 1
+   :depth: 2
    :class: singlecol
 
-This guide shows you how to create an application that uses the Node.js
-driver to connect to a MongoDB Atlas cluster.
+*Estimated time to complete this guide: 35 minutes.*
```
mongomoe: Love the concept of estimated time to complete. Is there any specific way you've come to this number?

source/quick-start.txt
```diff
@@ -7,193 +7,182 @@ Quick Start
...
 
 Add MongoDB as a Dependency
----------------------------
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Next, add the Node MongoDB driver as a project dependency. Use the
-``--save`` flag to save the dependency to the ``package.json`` file that
-was automatically generated in the previous step.
+Next, add the offical Node.js MongoDB driver to your project dependencies. Use
+NPM to download and install the ``mongodb`` package and provide the ``--save``
```
mongomoe: Could we omit the ``--save`` mention? As of `npm v5+` it is no longer required. See: https://stackoverflow.com/questions/19578796/what-is-the-save-option-for-npm-install

source/quick-start.txt
```diff
@@ -7,193 +7,182 @@ Quick Start
...
-----------------------
+~~~~~~~~~~~~~~~~~~~~~~
 
-Now that our system meets the Node driver's requirements, we can
-create a new project.
+After you verify that you installed the required version of Node and a
+compatible version of NPM, create a new project.
 
-First, create a directory for your project:
+First, create a directory for your project in your command line interface:
```
mongomoe: should this be `command-line` instead of `command line`?

source/quick-start.txt
ccho-mongodb: I summed the time for the pages in the Atlas Getting Started Guide (29 minutes) and then added some time to account for the rest of the sections (install node/npm/deps and run the code sample).

source/quick-start.txt
```diff
@@ -7,193 +7,182 @@ Quick Start
...
+  interactively select your project settings.
 
 Add MongoDB as a Dependency
----------------------------
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Next, add the Node MongoDB driver as a project dependency. Use the
-``--save`` flag to save the dependency to the ``package.json`` file that
-was automatically generated in the previous step.
+Next, add the offical Node.js MongoDB driver to your project dependencies. Use
```
mongomoe: Typo: "offical" should say "official"

the full sentence should read:
"Next, add the official Node.js MongoDB driver to your project dependencies. Use ..."

source/quick-start.txt
ccho-mongodb: I can remove it. Do you know if you can use a version of `npm` < v5 with Node v8? I just want to make sure that this change couldn't break anything for users that might meet the Node requirement as there's no npm version requirement.

mongomoe: Looks like it - found this chart online: 
https://nodejs.org/en/download/releases/

mongomoe: Screen Shot 2020-02-25 at 4 38 26 PM
Screen Shot 2020-02-25 at 4 38 21 PM

ccho-mongodb: This looks like what was released with node 8, but not necessarily what's compatible. I'll update the docs to mention that we require `npm` 5 and then get rid of the `save` option.

source/quick-start.txt
gilberto-mdb: I like the idea too but it might be discouraging to see that as the first line of a "quick" start. There's not much we can do to decrease this number outside of actual improvements to products we don't own. Do we need to include it?

source/quick-start.txt
```diff
@@ -7,193 +7,182 @@ Quick Start
...
+   :depth: 2
    :class: singlecol
 
-This guide shows you how to create an application that uses the Node.js
-driver to connect to a MongoDB Atlas cluster.
+*Estimated time to complete this guide: 35 minutes.*
 
-Check for Node and NPM
-----------------------
+This guide shows you how to create an application that uses the
```
gilberto-mdb: Do we need a title here? "Overview" etc.? Seems a bit strange to have this just floating between the "On this page" box and the "Set up Your Project" section.

source/quick-start.txt
```diff
@@ -7,193 +7,182 @@ Quick Start
...
+MongoDB Atlas is a fully-managed cloud database service that hosts your data
+on MongoDB instances. We show you how to get started with your own free (no
+credit card required) instance in this guide.
 
-.. code-block:: sh
+Follow the steps below to connect your Node.js application with a MongoDB
+instance.
 
-   node -v
+Set up Your Project
```
gilberto-mdb: Should "up" be "Up" ?

source/quick-start.txt
gilberto-mdb: Would you consider the code we provide to users below an "application"? It may be an application in the technical sense, but may not match the mental models our users associate with that word. This could lead to an unsatisfactory outcome for users of this guide.

source/quick-start.txt
```diff
@@ -7,193 +7,182 @@ Quick Start
...
 
-Check for Node and NPM
-----------------------
+This guide shows you how to create an application that uses the
+**Node.js driver** to connect to a **MongoDB Atlas cluster**.
 
-Ensure your system has Node version 8 or later and NPM (Node Package
-Manager) installed. Run the following commands to check that you meet
-these requirements:
+The Node.js driver is an interface through which you can connect to and
```
gilberto-mdb: "You can connect to and communicate with MongoDB instances using the Node.js interface."

Trying to frame this as outcome-oriented vs. feature-oriented ("do this thing" vs. "this is what this is").

source/quick-start.txt
```diff
@@ -7,193 +7,182 @@ Quick Start
...
+communicate with MongoDB instances. If you prefer to connect to MongoDB using
+a different programming language, select from the ones on our
```
gilberto-mdb: We say here that users can connect with a "programming language" but in the previous sentence we say they can connect using an "interface". Let's try to be consistent with our terminology, as we already use "driver" and "interface".

source/quick-start.txt
```diff
@@ -7,193 +7,182 @@ Quick Start
...
+Create a MongoDB Cluster
+------------------------
 
-.. code-block:: sh
+Set up a Free Tier Cluster in Atlas
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-  npm install --save dotenv
+After installing the Node MongoDB driver, create a MongoDB instance to store
+and manage your data. Go through the
```
gilberto-mdb: "Go through" might be confusing for international users. "Complete" ?

source/quick-start.txt
```diff
@@ -7,193 +7,182 @@ Quick Start
...
+data.
 
-.. code-block:: sh
+To connect to Atlas using the Node.js driver, we provide instructions to the
+driver on where and how to connect to your instance in a string called the
+*connection string*.
 
-  touch index.js
+Follow the
+:atlas:`Connect via Driver <driver-connection>` guide for instructions on
```
gilberto-mdb: I don't think this is the best page to link users to, as it doesn't have any screenshots and, unfortunately, it's not straightforward (from a UI perspective) to get the URI string. What about https://docs.atlas.mongodb.com/tutorial/connect-to-your-cluster/?

source/quick-start.txt
```diff
@@ -7,193 +7,182 @@ Quick Start
...
     try {
-      // Connect the client to the server
       await client.connect();
-      // Establish and verify connection
-      await client.db("admin").command({ ping: 1 });
-      console.log("Successfully connected to Atlas.");
-      // your code after this comment
 
-      // your code before this comment
+      const database = client.db('sample_mflix');
```
gilberto-mdb: I like this code! How do we let users know that this code may fail / not match the expected output if they did not load a sample dataset?

source/quick-start.txt
ccho-mongodb: I can remove it if you want. I thought it was helpful to set expectations, but can see how surfacing an estimated time that's more than a few minutes can cause a user to bounce.

source/quick-start.txt
ccho-mongodb: By title case standards, it's lowercase because it's a preposition.

source/quick-start.txt
ccho-mongodb: Ok, I think to address this, I can just change "ones" to "drivers".

source/quick-start.txt
ccho-mongodb: This is to address your request to include "What is the Node.js driver?" (without linking to another page) and "Where do I go if I'm not on the page I'm looking for". The prior sentence (opening sentence) mentions what it's used for. I'm not sure this needs to be either outcome-oriented or feature-oriented as it's a definition, similar to the next paragraph about what Atlas is.
The sentence you provided restates the first sentence, so I'm not sure how to use it. Do you have a different suggestion?

source/quick-start.txt
ccho-mongodb: How about this sentence after the sample output:
"If or receive an error, check whether you included the proper connection string in the application and loaded the sample dataset."

source/quick-start.txt
ccho-mongodb: Was there feedback about that? We did not have a title for the lead-in sentence before (see the original ll. 13-14)

ccho-mongodb: Yes, it is an application, but what other words would you suggest based on users' mental models? If a Jasmine persona looks for specific keywords, I'd change this, but otherwise Lucas personas should understand it since it is accompanied with a description of its functionality.

source/quick-start.txt
ccho-mongodb: You don't think they would be more confused by needing to install the Python driver or mongo shell on that page?

source/quick-start.txt
gilberto-mdb: Thanks

source/quick-start.txt
gilberto-mdb: Not sure if there was previous feedback by others on this. Perhaps it's because the "On this page" box wasn't here in Nathan's version? I'm flexible on this given that we'll be removing the estimated completion time sentence. It seemed like Michael Waltzer had feedback on the inclusion of the "On this page" box, but I can't add him to GitHub apparently.

gilberto-mdb: I was initially concerned given its usage so early on the page, but as I continued through the content, it looks like we use it consistently so I'm less concerned that there will be confusion.

source/quick-start.txt
gilberto-mdb: Perhaps the order is what's confusing here. What if the first sentence on this page (the first paragraph) becomes the second or the last sentence of the second paragraph? 

Another suggestion is:
```
The Node.js driver is an interface through which you can connect to and communicate with MongoDB instances. This guide shows you how to create an application that uses the Node.js driver to connect to a MongoDB Atlas cluster.

If you prefer to connect to MongoDB using a different programming language, select from the ones on our drivers documentation page.
```

source/quick-start.txt
gilberto-mdb: My feedback was centered around the possible confusion around "connecting **with** a programming language" – they're connecting with a tool written in a language they prefer, vs. just the programming language (which to me, is a bit confusing – "Connect to your cluster with Python" wouldn't make too much sense to me).

I meant something like:

"If you prefer to connect to MongoDB using a driver [for/with] a different programming language, please select a different driver listed on our..."

We'd select either "for" or "with" in the example sentence above ^

source/quick-start.txt
gilberto-mdb: I think they would be confused by both pages, but I think users will be confused less by pages that use screenshots. Personally, I'm averse to linking them to any page here as it doesn't seem like we have a good page to link them to. Can we add a screenshot here instead?

source/quick-start.txt
gilberto-mdb: Excellent

source/quick-start.txt
ccho-mongodb: The order I placed it in is:
-What is the purpose of the guide
-What is the Node.js driver
-Where should I go if I'm in the wrong place
-What is Atlas

I think the order you are suggesting is:
-What is the Node.js driver
-What is the purpose of the guide
-?
-?

Doesn't it make more sense to mention the purpose of the guide before defining the Node.js driver?

source/quick-start.txt
ccho-mongodb: Just to clarify, do you want screenshots of:
https://docs.atlas.mongodb.com/tutorial/connect-to-your-cluster/#id2
and
https://docs.atlas.mongodb.com/tutorial/connect-to-your-cluster/#copy-the-provided-connection-string-from-the-connection-string-only-dialog-tab
?

source/quick-start.txt
ccho-mongodb: 1)  I think the correction I identified needs to exist to distinguish between "programming language" and "driver".

2) Can we punt on "with" vs "using? I'm not sure I see any advantage to changing it since we use those interchangeably across our docs. I prefer the latter because it implies the object is a requirement whereas "with" does not have as strong an association and is used more to describe "accompanied by" but does not imply the object is required.

MikeyDubs:
I'd feel more comfortable putting this in front of users to make sure these updates are easy to follow. My biggest concern is the new Atlas docs links - we're sending them somewhere new (that we hadn't in our previous test), and I'd want to be sure they're able to follow those docs. This is where a lot of people tripped up. I'd also want to validate that the connection string is easy to find, and that users understand what it is.  Probably the biggest pain point.

Also curious, why did we remove checking the version of Node? People seemed to like that.

Also curious as to why we removed the .env file stuff - one person mentioned they liked it, and it never got in the way for any user who we tested this with.

Again, I'd want to test the updated version with at least 2 - 3 users to ensure these updates successfully take care of the pain points from the previous version.

ccho-mongodb:
Thanks for your feedback! Gilberto and I had discussed the user feedback and decided on the changes listed in the JIRA ticket (https://jira.mongodb.org/browse/DOCSP-9048). I wasn't sure whether you shared Gilberto's opinions, but since you were also involved in the user studies, next time we'll make sure you're included in this process.

I'll respond inline to your comments:

>  My biggest concern is the new Atlas docs links - we're sending them somewhere new (that we hadn't in our previous test), and I'd want to be sure they're able to follow those docs. This is where a lot of people tripped up. 

The prior version had three links to different sections of the Atlas docs and this version has reduced that number by one. I'm not sure how we can avoid linking since the alternative is to inline the instructions and inevitably take on the responsibility of updating all driver docs whenever something small changes in the Atlas flow.

> I'd also want to validate that the connection string is easy to find, and that users understand what it is. Probably the biggest pain point.

What would you suggest that we do differently? This version explains what the connection string is and links to an explanation whereas the prior version did not, and referred to placeholder values without explaining them.

> Also curious, why did we remove checking the version of Node? People seemed to like that.

It's still a section, but it's linked to the official Node docs. Gilberto and I discussed this section, focused on how much space this took, and how it might require creating instructions for the specific OS the user might be using. Our solution was to link to the official docs.

> Also curious as to why we removed the .env file stuff - one person mentioned they liked it, and it never got in the way for any user who we tested this with.

Gilberto and I discussed the relevance of this to a Quick Start and comments that Matt Broadstone had made on using dotenv. We decided that maybe this material would belong in a "prepare your application for a production environment" section if we decided to provide that, but decided it should not be mandatory (the final example code also included a check to make sure you set this up and would otherwise error and exit) and that it got in the way of an expedient completion.

Let me know what you think when you get the chance.

source/quick-start.txt
gilberto-mdb: Current order (taken from your comment):
-What is the purpose of the guide
-What is the Node.js driver
-Where should I go if I'm in the wrong place
-What is Atlas

Potential proposed order:
-What is the purpose of the guide
-Where should I go if I'm in the wrong place
-What is the Node.js driver
-What is Atlas


Thoughts? If you don't agree then we can leave it as-is.

source/quick-start.txt
gilberto-mdb: 1. I'm lost now.

2. Sure.

source/quick-start.txt
gilberto-mdb: Yeah, what do you think?

The first screenshot is helpful. The second screenshot is helpful too, but there is a modal that a user needs to go through to get from the first screenshot to the second.

gilberto-mdb:
Agree with Cho on all the points here. The only things I'll add:

(1) 21 total users interviews over the course of 4 months for 1 page in the docs is more than enough and any additional interviews would be exceedingly gratuitous. At this point, it's much cheaper and much faster to launch to a full audience now and get feedback vs. setting up more sessions. We can incorporate feedback as it comes in (this is a standard go-to-market process for new content).

(2) We decreased the total number of different Atlas links we send them to.

(3) We're likely going to add screenshots to help users find the connection string vs. linking them out, if possible.

(4) Although no one had issues with the .env content, it was unnecessary and adds to the overall time required to complete a "quick start". Cho is correct in moving it to a different place in the docs.

(5) I am curious to see how users respond to the removal of the node -v and npm -v content vs. the links to the official docs. It's not a blocker to launch, and adding that content back in if it turns out to be an issue is trivial.

@ccho-mongodb @MikeyDubs

source/quick-start.txt
ccho-mongodb: I think either way is ok with me. The ordering I used seeks to explain what the Node.js driver is immediately so the user can decide whether they're in the right place (focus on Jasmine persona), and your proposed ordering centralizes the definitions.

I am not certain what the best way and there might ultimately not be one. I'll try it your way and maybe one of your questions in future user testing or metrics analysis can be "were you able to identify that you're on the documentation page for the appropriate language driver?"

gilberto-mdb: Sounds good, thanks Chris.

source/quick-start.txt
```diff
@@ -7,193 +7,202 @@ Quick Start
...
+click the **Connect** button for the cluster in which you want to connect
+as shown below.
 
-Connect to your Cluster
------------------------
+.. figure:: /includes/figures/atlas_connection_select_cluster.png
 
-For the final step, we'll write some code that tests your environment's
-connection to your instance of MongoDB.
+Follow the instructions in the dialog to whitelist your IP address and create
```
gilberto-mdb: Probably doesn't matter, but if they've completed all steps of the Getting Started with Atlas guide, then they'll have already done these two things (IP whitelist and new MongoDB user).

source/quick-start.txt
```diff
@@ -7,193 +7,202 @@ Quick Start
...
+.. figure:: /includes/figures/atlas_connection_copy_string.png
 
-  touch index.js
+Save your connection string to a safe location that you can access in the
+next step.
 
-Open this file and paste in the following code:
+Next, create a file to contain your application called ``index.js`` in your
+``node_quickstart`` directory. Add the following code, replacing the ``uri``
+variable with your connection string. Make sure to replace the "<password>"
```
gilberto-mdb: I know Atlas pre-fills in the user for you in the connection string, just hoping there isn't much confusion here.

source/quick-start.txt
```diff
@@ -7,193 +7,202 @@ Quick Start
...
 
-      // your code before this comment
+      const database = client.db('sample_mflix');
+      const collection = database.collection('movies');
+
+      // Query for a movie that has the title 'Back to the Future'
+      const query = { title: 'Back to the Future' };
+      const movie = await collection.findOne(query);
+
+      console.log(movie);
```
gilberto-mdb: console.log(collection.findOne({})?

Or

doc = collection.findOne({});
console.log(doc);

? 

I'm flexible here honestly, just thought a query might add unnecessary confusion.

source/quick-start.txt
ccho-mongodb: Not sure if we're referring to the same thing, but I only mentioned that they need to replace "<password>" because that is not filled in by Atlas, probably for security reasons.

source/quick-start.txt
ccho-mongodb: I would strongly prefer to keep the same patterns across all our CRUD operations.
The reason I provided a query is so that the user has control of what they are querying for as well as to match the sample output. This might also encourage experimentation with different queries whereas findOne without a query is more difficult for a user to figure out how to modify.

source/quick-start.txt
ccho-mongodb: I'll go back and check the Getting Started with Atlas. While I remember there was a step to create a user, I do not remember whether it was through the "Connect" dialog wizard.

ccho-mongodb: Ah, I see the user goes through the Python or Mongo Shell example that I thought would be confusing in "Get Started with Atlas". I guess I will update this to only remind them how to navigate to retrieve the info.

source/quick-start.txt
gilberto-mdb: Yeah, we're referring to the same thing.

source/quick-start.txt
gilberto-mdb: Sounds good.

source/quick-start.txt
ccho-mongodb: Ok, is there anything that you wanted to change then? The sentence mentions that they should replace the "`<password>`" placeholder and nothing else.

gilberto-mdb: I was just noting that we ask them to replace "`<password>`", which makes sense. I was wondering whether or not we need to make a note about the username, which I know does not need to be replaced because presumably Atlas auto-fills it in for users in the connection string. "Your username should appear in the connection string..." etc. This is a non-blocking comment and I am OK to go live as-is.

ccho-mongodb: I see. I don't think it's necessary since Atlas auto-fills it. If they're not using Atlas, they'll need to read some other literature on how to connect. Since there wasn't any link to a doc to describe how to connect without Atlas in the original doc, I'm guessing that was intentional.

PR #78, opened 2020-03-12
DOCSP-9028: index page and issues and help page fixes

ccho-mongodb:
JIRA: 
https://jira.mongodb.org/browse/DOCSP-9028

Staging:
index page: https://docs-mongodborg-staging.corp.mongodb.com/6664b62/node/docsworker-xlarge/DOCSP-9028-index-page-fixes/

issues and help page: https://docs-mongodborg-staging.corp.mongodb.com/6664b62/node/docsworker-xlarge/DOCSP-9028-index-page-fixes/issues-and-help

### DIFF
--- a/source/index.txt
+++ b/source/index.txt
@@ -23,7 +23,7 @@ Introduction
 ------------
 The official MongoDB Node.js driver allows Node.js applications to
 connect to MongoDB and work with data. The driver features both
-callback-based and Promise-based interaction with MongoDB.
+callbacks and Promises when communicating with MongoDB.
 
 .. note::
 
@@ -67,7 +67,7 @@ material on using the Node.js driver for the following:
 * :doc:`Indexes </fundamentals/indexes>`: create and design indexes to make
   your queries efficient
 
-* :doc:`Collations </fundamentals/collations>`: apply language-specific 
+* :doc:`Collations </fundamentals/collations>`: apply language-specific
   sorting rules to your query results
 
 * :doc:`Logging </fundamentals/logging>`: configure the driver to log
--- a/source/index.txt
+++ b/source/index.txt
@@ -22,8 +22,9 @@ MongoDB Node Driver
 Introduction
 ------------
 The official MongoDB Node.js driver allows Node.js applications to
-connect to MongoDB and work with data. The driver features both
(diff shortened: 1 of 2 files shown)


source/index.txt
```diff
@@ -23,7 +23,7 @@ Introduction
 ------------
 The official MongoDB Node.js driver allows Node.js applications to
 connect to MongoDB and work with data. The driver features both
-callback-based and Promise-based interaction with MongoDB.
+callbacks and Promises when communicating with MongoDB.
```
mbroadst: I would say "All asynchronous API in the driver supports returning Promises if a callback is not provided"

Did y'all decide that having a "fundamentals" page on this was not beneficial to our users? It seems like a `/fundamentals/promises` page would be really useful to answer a few common questions I get from users:
- what happens if I omit a callback?
- does the driver support async/await (answer: yes! anything that returns a Promise can be awaited)
- confusion around common async/await mistakes (forgetting `await` in `while (cursor.hasNext())`)

ccho-mongodb: That sounds like a good idea, I'll create a ticket for a new page on Promises. 

Our DevEd team had a discussion about whether to create a page on Promises, focused on what they were and the syntax related to them, but decided to link out to MDN docs instead. Given this information on common questions that you receive, I think there would be enough driver-relevant material to justify this new page.

Aside from the new page, I'm not sure I understand the recommendation you provided:
"All asynchronous API in the driver supports returning Promises if a callback is not provided"

Did you mean something like this:
"The official MongoDB Node.js driver allows Node.js applications to
connect to MongoDB and work with data. The driver features an asynchronous API which allows you to specify callbacks in method calls, or otherwise return values using Promises."

ccho-mongodb: New ticket for Fundamentals page on Promises https://jira.mongodb.org/browse/DOCSP-9706

mbroadst: Yeah I think that's closer, I just wanted to be "Promise-positive" by saying that all the async api returns a Promise _if_ you don't provide a callback. Small difference, but we're trying to push our users to use async/await as much as possible

ccho-mongodb: Makes sense, I'll swap the mention order.

PR #83, opened 2020-03-27
DOCSP-9027: Update documentation to reflect that NodeJS driver 4.0 requires Node 10 or later

ccho-mongodb:
JIRA:
https://jira.mongodb.org/browse/DOCSP-9027

Staging:
https://docs-mongodborg-staging.corp.mongodb.com/4b524d6/node/docsworker-xlarge/DOCSP-9027-node10/


source/faq.txt
```diff
@@ -24,13 +24,22 @@ What Is the Difference Between "connectTimeoutMS", "socketTimeoutMS" and "maxTim
      - Description
    * - connectTimeoutMS
      - 30000
-     - ``connectTimeoutMS`` is a `connection-setting <https://mongodb.github.io/node-mongodb-native/3.5/reference/connecting/connection-settings/>`_ option that sets the time in milliseconds to establish a connection to the MongoDB server before timing out.
+     - ``connectTimeoutMS`` is a :doc:`connection option
```
mbroadst: we actually just received feedback for this in the [community forums](https://community.mongodb.com/t/connect-timeout-and-execution-timeout-in-nodejs-driver/21290), it seems that this options is easily confused with gating the behavior of `MongoClient#connect`. I think the language should be updated to indicate that it is specific to individual connections in a connection pool. Specifically, this option defines the amount of time the driver's connection pool will wait before determining a connection to a MongoDB server has failed to connect. You might want to additionally call out that this is not directly related to `MongoClient#connect`, and they should instead see `serverSelectionTimeoutMS` for that.

ccho-mongodb: I think I'm confused now after reading this part of your explanation:
> You might want to additionally call out that this is not directly related to MongoClient#connect, and they should instead see serverSelectionTimeoutMS for that. 

From various versions of the NodeJS documentation, I thought this setting controlled the maximum interval of time to wait in which no response has been received from the server (for each connection requested in the driver's connection pool, as you clarified) during a call to connect and that `serverSelectionTimeoutMS` related to waiting for replica sets to connect to one another.

If you get the chance, could you invite me to that community forums conversation that you linked? I get an error message that it doesn't exist or is private after logging in. Thanks.

mbroadst: Oops, I updated the link.

The problem here is that people think when they set `connectTimeoutMS`, that value will be used as an upper bound on how long it takes for `MongoClient.connect` to complete. That's not the case. `MongoClient.connect` starts an internal server monitor, that monitor creates a connection and `connectTimeoutMS` is used to determine how long we are willing to wait for _that_ connection to connect to a mongod. `MongoClient.connect` is internally implemented as a "server selection", meaning it will attempt to find a suitable server for up to `serverSelectionTimeoutMS`, and otherwise fail. This is not specific to replica sets.

ccho-mongodb: Is there any reading you can recommend on how the internal server monitor works (for personal edification)? 

If I understand correctly, during MongoClient.connect:
1) internal server monitor is created - serverSelectionTimeoutMS timer starts
2) monitor requests a connection from the connection pool - connectTimeoutMS starts
3) connection attempts to communicate with the server. If one fails, does it retry with another?
4) connection is established - serverSelectionTimeoutMS and connectTimeoutMS end
Is this correct?

I'm currently adding a new entry for "serverSelectionTimeoutMS" which might resemble the following to help clarify the FAQ answer:
 
>    * - serverSelectionTimeoutMS
>      - 30000
>      - ``serverSelectionTimeoutMS``is a connection option that sets the
>        number of milliseconds to wait for a successful connection with
>        a MongoDB server when calling ``MongoClient#connect``. The default
>        value allows enough time for a typical server primary election to
>        complete. Use this option to limit the time ``MongoClient#connect``
>        waits before failing.

mbroadst: Sure, it's documented in the [Server Discovery and Monitoring specification](https://github.com/mongodb/specifications/blob/master/source/server-discovery-and-monitoring/server-discovery-and-monitoring.rst). 

I don't think we need to document this here, we just need to say that `connectTimeoutMS` is related to establishing individual TCP connections to a server, and `serverSelectionTimeoutMS` is probably what they're looking for if they are trying to make timing out `MongoClient#connect` faster. You won't find specification documentation for a `connect` method, because most drivers don't have one, it's just something weird that the node driver has.

source/faq.txt
```diff
@@ -24,13 +24,25 @@ What Is the Difference Between "connectTimeoutMS", "socketTimeoutMS" and "maxTim
...
+       </fundamentals/connection#connection-options>` that sets the time in
+       milliseconds for an individual connection from your connection pool to
+       establish a TCP connection to the MongoDB server before timing out.
+       Use the ``serverSelectionTimeoutMS`` option instead to fail when
+       :node-api:``MongoClient.connect <MongoClient.html#.connect>` is unable
+       to establish a connection with the MongoDB server.
    * - socketTimeoutMS
      - 360000
-     - ``socketTimeoutMS`` is a ``connection-settings`` option that sets the number of milliseconds a socket stays inactive after the driver has successfully connected before closing the connection.
+     - ``socketTimeoutMS`` is a :doc:`connection option
```
mbroadst, line 36: a few things here:
- `socketTimeoutMS` is a connection option as well as an option supported on individual operations
- I think the option is better described this way: "`socketTimeoutMS` is the amount of time the driver will wait for an inactive socket before closing it. The default is to never time out the socket. This only applies to sockets which have already been connected"

source/faq.txt
```diff
@@ -24,13 +24,25 @@ What Is the Difference Between "connectTimeoutMS", "socketTimeoutMS" and "maxTim
...
+       </fundamentals/connection#connection-options>` that sets the number of
+       milliseconds a socket stays inactive after the driver has successfully
+       connected before closing the connection.
    * - maxTimeMS
      - N/A
-     - :node-api:`maxTimeMS() </Cursor.html#maxTimeMS>` is a :manual:`cursor method </reference/method/js-cursor>` that specifies the max time limit for processing an operation on a cursor. If an operation runs over the time allotted it will return a timeout error.
+     - :node-api:`maxTimeMS() </Cursor.html#maxTimeMS>` is a method that
```
mbroadst, line 42: a few things here:
- `maxTimeMS` is a connection option, cursor option, as well as something you can set per-operation (through `options`) 
- "`maxTimeMS` is the maximum amount of time the server should wait for an operation to complete, once it has reached the server. If an operations runs over the specified time, it will return a timeout error"

source/faq.txt
```diff
@@ -24,13 +24,25 @@ What Is the Difference Between "connectTimeoutMS", "socketTimeoutMS" and "maxTim
      - Description
    * - connectTimeoutMS
      - 30000
-     - ``connectTimeoutMS`` is a `connection-setting <https://mongodb.github.io/node-mongodb-native/3.5/reference/connecting/connection-settings/>`_ option that sets the time in milliseconds to establish a connection to the MongoDB server before timing out.
+     - ``connectTimeoutMS`` is a :doc:`connection option
+       </fundamentals/connection#connection-options>` that sets the time in
+       milliseconds for an individual connection from your connection pool to
+       establish a TCP connection to the MongoDB server before timing out.
+       Use the ``serverSelectionTimeoutMS`` option instead to fail when
```
mbroadst, line 31: "This is not directly related to :node-api:``MongoClient.connect <MongoClient.html#.connect>`, if you wish to control the timeout for an initial connection to a cluster please see <link to `serverSelectionTimeoutMS` docs>"

source/faq.txt
ccho-mongodb: Is the default value of "360000" incorrect since in your second sentence, you mention that the default is to never time out? Or does the default apply only when it is a connection option and not for individual operations?

mbroadst: Hm, you're right that for some reason the legacy options apply a 360000 default, but the idea is that this is effectively "forever" and is actually specified as such. Let's say its no timeout for v4, we'll fix that by then

PR #88, opened 2020-04-13
DOCSP-9706: Promises and Callbacks

ccho-mongodb:
JIRA:
https://jira.mongodb.org/browse/DOCSP-9706

Staging:
Changes to Index Page:
https://docs-mongodbcom-staging.corp.mongodb.com/3de88bd/node/docsworker/DOCSP-9706-promises/

Fundamentals > Promises and Callbacks:
https://docs-mongodbcom-staging.corp.mongodb.com/3de88bd/node/docsworker/DOCSP-9706-promises/fundamentals/promises

### DIFF
--- a/source/fundamentals/promises.txt
+++ b/source/fundamentals/promises.txt
@@ -104,11 +104,11 @@ in the following code snippet, it throws the following error:
 ``MongoError: Cursor is closed``.
 
 .. code-block:: js
-
+   
+   // WARNING: this snippet throws a MongoError
    const cursor = collection.find();
 
    while (cursor.hasNext()) {
-      // await to get the result of the Promise
       console.log(await cursor.next());
    }
 
@@ -124,15 +124,24 @@ document objects.
 
 .. code-block:: js
 
+   // WARNING: this snippet prints Promises instead of the objects they resolve to
    const cursor = collection.find();
 
    while (await cursor.hasNext()) {
-      // print the Promises rather than the types they resolve to
       console.log(cursor.next());
    }
 
 Use ``await`` before both the ``hasNext()`` and ``next()`` method calls to
-ensure that you are operating on the correct return values.
+ensure that you are operating on the correct return values as demonstrated
+in the following code:
+
+.. code-block:: js
+
+   const cursor = collection.find();
+
+   while (await cursor.hasNext()) {
+      console.log(await cursor.next());
+   } 
 
 .. note::
 
--- a/source/fundamentals/promises.txt
+++ b/source/fundamentals/promises.txt
@@ -13,17 +13,17 @@ Promises and Callbacks
 Overview
 --------
 
-NodeJS driver uses the asynchronous Javascript API to communicate with
+The NodeJS driver uses the asynchronous Javascript API to communicate with
 your MongoDB cluster.
 
-Asynchronous Javascript allows you to execute operations without waiting for 
+Asynchronous Javascript allows you to execute operations without waiting for
 the processing thread to become free. This helps prevent your application
 from becoming unresponsive to web UI rendering or network requests when
-executing long-running operations. For more information about asynchronous 
-Javascript, see the MDN web documentation on 
+executing long-running operations. For more information about asynchronous
+Javascript, see the MDN web documentation on
 `Asynchronous Javascript <https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous>`_.
 
-This section describes two features of asynchronous Javascript -- 
+This section describes two features of asynchronous Javascript --
 `Promises`_ and `Callbacks`_ -- that you can use with the NodeJS driver to
 access the results of your method calls to your MongoDB cluster.
 
@@ -32,17 +32,53 @@ Promises
 
 A Promise is an object returned by the asynchronous method call that allows
 you to access information on the eventual success or failure of the operation.
-For more information on Promises and related terminology, see the MDN
-documentation on 
+They are often used to execute operations that take some time and have
+when you want to notify once it's completed. For more information on Promises
+and related terminology, see the MDN documentation on
 :mdn:`Promises <Web/JavaScript/Reference/Global_Objects/Promise>`.
 
-If you call a method such as ``findOneAndModify()``, ``countDocuments()``, 
-or ``update()``, the NodeJS driver returns a Promise.
+If you call a driver method such as ``findOneAndModify()``,
+``countDocuments()``, or ``update()``, the NodeJS driver returns a Promise.
 
-You can use the ``await`` operator on the Promise to
-cause the enclosing ``async`` method to pause until the Promise reaches
+You can use the ``await`` operator inside of an ``async`` method on the
+Promise to cause the enclosing method to pause until the Promise reaches
 either the **Fulfilled** or **Rejected** state. If it reaches the **Fulfilled**
 state, the reference contains the operation result as an object or value.
+
+.. code-block:: js
+
+   async function run() {
+     ...
+     const result = await collection.insertOne({ name: "Mount Everest", meters: 8848 });
+
+     // print the result of the insertOne() call
+     console.log(result);
+   }
+
+If you want to execute logic once the Promise reaches the **Fulfilled**
+or **Rejected** state, you can append the ``then()`` method to the Promise.
+The first parameter is the method that gets called when the Promise reaches
+the **Fulfilled** state and the optional second parameter is the method that
+gets called when it reaches the **Rejected** state. These methods are called
+**callbacks** because they execute only after the Promise completes.
+
+.. code-block:: js
+
+   ...
+   collection.updateOne({ name: "Mount McKinley" }, { meters: 6190 })
+     .then(
+       res => console.log(`Updated ${res.result.n} documents`),
+       err => console.error(`Something went wrong: ${err}`)
+     );
+
+If you only need to handle transitions to the **Rejected** state, use the
+``catch()`` method which accepts a single callback, executed when the Promise
+transitions to the **Rejected** state.
+
+.. code-block:: js
+
+     const deleteResult = await collection.deleteOne({ name: "Mount Doom" })
+       .catch(err => console.error(`Fatal error occurred: ${err}`));
 
 .. note::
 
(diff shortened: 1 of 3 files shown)


source/fundamentals/promises.txt
```diff
@@ -0,0 +1,147 @@
...
+Overview
+--------
+
+NodeJS driver uses the asynchronous Javascript API to communicate with
+your MongoDB cluster.
+
+Asynchronous Javascript allows you to execute operations without waiting for 
+the processing thread to become free. This helps prevent your application
+from becoming unresponsive to web UI rendering or network requests when
+executing long-running operations. For more information about asynchronous
```
mongomoe, line 22: This is just personal preference, but we could remove the "For more information about asynchronous Javascript, see the MDN ..." line and just link to the MDN link in the first sentence of the paragraph: 
"`Asynchronous Javascript<developer.mozzila.org/../Asynchronous/>`_ allows you to execute operations without waiting for 
the processing thread to become free"

source/fundamentals/promises.txt
```diff
@@ -0,0 +1,147 @@
...
+This section describes two features of asynchronous Javascript -- 
+`Promises`_ and `Callbacks`_ -- that you can use with the NodeJS driver to
+access the results of your method calls to your MongoDB cluster.
+
+Promises
+--------
+
+A Promise is an object returned by the asynchronous method call that allows
+you to access information on the eventual success or failure of the operation.
+For more information on Promises and related terminology, see the MDN
```
mongomoe, line 35: Same as the above comment on "for more information"

Feel free to ignore this comment if you disagree with it. I'm partial to included the links within the sentence because it cuts down on the bulk of the doc, and achieves the same purpose.

source/fundamentals/promises.txt
```diff
@@ -0,0 +1,147 @@
...
+**type** specified in the :node-api:`API documentation <>`.
+
+ .. note::
+ 
+    If you specify a callback, the method *does not* return a Promise.
+
+Operational Considerations
+--------------------------
+
+When using ``async`` methods, make sure to use ``await`` when you call methods
```
mongomoe, line 76: The phrase "make sure" makes it seem suggested but not required to use ``await``. Can you rephrase this to illustrate that ``await`` is required when using ``async`` functions

source/fundamentals/promises.txt
```diff
@@ -0,0 +1,147 @@
...
+
+   while (await cursor.hasNext()) {
+      console.log(await cursor.next());
+   } 
+
+.. note::
+
+   For additional information on using Promises and Callbacks with the MongoDB
+   NodeJS driver, see this
+   `MongoDB University m220js course video <https://youtu.be/R4AEyKehpss>`_.
```
mongomoe, line 147: it might be helpful to change the link text to:  see this Mongodb University M220js course video on `asynchronous programming <youtube link>`_. So that the user gets more of a preview on what they're clicking (a video about async programming) rather than the general  term "MongoDB University m220js course video"

source/fundamentals/promises.txt
```diff
@@ -0,0 +1,147 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecols
+
+Overview
+--------
+
+NodeJS driver uses the asynchronous Javascript API to communicate with
```
terakilobyte, line 16:
```suggestion
The NodeJS driver uses the asynchronous Javascript API to communicate with
```

source/fundamentals/promises.txt
ccho-mongodb: Thanks, I appreciate your thoughts on streamlining this. I intended to notify the user where the link goes, especially since it potentially draws users away from the page and our domain. The structure I'm using seems aligned with that goal:
1) Describe the term
2) Explain when it's used
3) Direct to other resources for further learning

source/fundamentals/promises.txt
```diff
@@ -0,0 +1,194 @@
...
+Overview
+--------
+
+The NodeJS driver uses the asynchronous Javascript API to communicate with
+your MongoDB cluster.
+
+Asynchronous Javascript allows you to execute operations without waiting for
+the processing thread to become free. This helps prevent your application
+from becoming unresponsive to web UI rendering or network requests when
+executing long-running operations. For more information about asynchronous
```
terakilobyte, line 22:
```suggestion
from becoming unresponsive when
executing long-running operations. For more information about asynchronous
```

source/fundamentals/promises.txt
```diff
@@ -0,0 +1,194 @@
...
+Javascript, see the MDN web documentation on
+`Asynchronous Javascript <https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous>`_.
+
+This section describes two features of asynchronous Javascript --
+`Promises`_ and `Callbacks`_ -- that you can use with the NodeJS driver to
```
terakilobyte, line 27:
```suggestion
``Promises`` and ``Callbacks`` -- that you can use with the NodeJS driver to
```

source/fundamentals/promises.txt
```diff
@@ -0,0 +1,194 @@
...
+access the results of your method calls to your MongoDB cluster.
+
+Promises
+--------
+
+A Promise is an object returned by the asynchronous method call that allows
+you to access information on the eventual success or failure of the operation.
+They are often used to execute operations that take some time and have
+when you want to notify once it's completed. For more information on Promises
```
terakilobyte, line 36:
```suggestion
They are used to execute operations upon the completion of other operations that take
an indeterminate time and have. For more information on Promises
```

source/fundamentals/promises.txt
```diff
@@ -0,0 +1,194 @@
...
+and related terminology, see the MDN documentation on
+:mdn:`Promises <Web/JavaScript/Reference/Global_Objects/Promise>`.
+
+If you call a driver method such as ``findOneAndModify()``,
+``countDocuments()``, or ``update()``, the NodeJS driver returns a Promise.
+
```
terakilobyte, line 42:
```suggestion

Await
~~~~

```

terakilobyte: I think the entire await section should be after talking about `.then()` and `.catch()`, personally. With the current split it feels like we're mixing topics. Thoughts?

source/fundamentals/promises.txt
```diff
@@ -0,0 +1,194 @@
...
+gets called when it reaches the **Rejected** state. These methods are called
+**callbacks** because they execute only after the Promise completes.
+
+.. code-block:: js
+
+   ...
+   collection.updateOne({ name: "Mount McKinley" }, { meters: 6190 })
+     .then(
+       res => console.log(`Updated ${res.result.n} documents`),
+       err => console.error(`Something went wrong: ${err}`)
```
terakilobyte, line 71: ? I believe if the operation errors it will end up being an unhandled promise rejection in this example because there is no `.catch()` chain. Am I mistaken?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#Chained_Promises

source/fundamentals/promises.txt
```diff
@@ -0,0 +1,194 @@
...
+or **Rejected** state, you can append the ``then()`` method to the Promise.
+The first parameter is the method that gets called when the Promise reaches
+the **Fulfilled** state and the optional second parameter is the method that
+gets called when it reaches the **Rejected** state. These methods are called
+**callbacks** because they execute only after the Promise completes.
+
+.. code-block:: js
+
+   ...
+   collection.updateOne({ name: "Mount McKinley" }, { meters: 6190 })
```
terakilobyte, line 68: Updates require an update operator

source/fundamentals/promises.txt
```diff
@@ -0,0 +1,194 @@
...
+used to enforce the order of processing commands.
+
+In the MongoDB NodeJS driver, you can optionally declare a callback method to
+async operations that normally return Promises such as ``findOneAndUpdate()``,
+``countDocuments()``, or ``update()``. Once the operation completes execution,
+the callback method executes as shown in the following code snippet:
+
+.. code-block:: js
+
+   collection.findOneAndUpdate({ name: "Barronette Peak" }, { name: "Baronette Peak" },
```
terakilobyte, line 104: An update operator is required here.

source/fundamentals/promises.txt
```diff
@@ -0,0 +1,194 @@
...
+When using ``async`` methods, use ``await`` when you call methods
+that return Promises to ensure they complete before executing any other
+logic that depends on the result. Consider the following example in which
+we iterate over a cursor using ``hasNext()``, which returns a Promise that
+resolves to a boolean that indicates whether additional results exist, and
+``next()`` which returns a Promise that resolves to the next entry the
+cursor is pointing to.
+
+One common mistake is to forget to use ``await`` to get the value of the
+result rather than the Promise object. Consider the following code:
```
terakilobyte, line 132: This feels like a rehash of what is being said in the above paragraph and interrupt the flow of this section to me

source/fundamentals/promises.txt
```diff
@@ -0,0 +1,194 @@
...
+
+   // WARNING: this snippet may cause an infinite loop
+   const cursor = collection.find();
+
+   while (cursor.hasNext()) {
+      console.log(cursor.next());
+   }
+
+Since the call to ``hasNext()`` returns a ``Promise``, the conditional
+statement returns ``true`` regardless of the value that it resolves to.
```
terakilobyte, line 144: Awesome

source/fundamentals/promises.txt
ccho-mongodb: Oops, you mean options object right? https://mongodb.github.io/node-mongodb-native/3.5/api/Collection.html#findOneAndUpdate

source/fundamentals/promises.txt
ccho-mongodb: This example is demonstrating using both at the same time. The examples you linked omit the catch in each of the `then()` methods, and then catch them at the end which is a different use case. I can show that too, if it's necessary.

source/fundamentals/promises.txt
ccho-mongodb: I think this could be misleading to the reader because:
-It's not a requirement for the operation to take an indeterminate time to use Promises
-It's not a requirement to use Promises for operations that take an indeterminate time to run
-I think there's a sentence fragment here: "and have".
-There's no mention about notification once it's completed which is a key feature of what it replaces/wraps (callbacks)

source/fundamentals/promises.txt
ccho-mongodb: Sure, makes sense to me.

source/fundamentals/promises.txt
ccho-mongodb: Never mind! Need both

source/fundamentals/promises.txt
ccho-mongodb: I noticed I repeated nearly the same sentence, so I'm getting rid of one.

source/fundamentals/promises.txt
```diff
@@ -0,0 +1,242 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 1
+   :class: singlecols
+
+Overview
+--------
+
+The NodeJS driver uses the asynchronous Javascript API to communicate with
```
nraboy, line 16: Why do we refer to it as `NodeJS` when the official Node site refers to it as `Node.js`?

Not having the correct naming is like saying `Javascript` instead of `JavaScript`. Unless there is a specific reason, I think we lose credibility with developers.

source/fundamentals/promises.txt
```diff
@@ -0,0 +1,242 @@
...
+a ``catch()`` method to the end of a Promise chain.
+
+.. code-block:: js
+
+   deleteOne({ name: "Mount Doom" })
+     .then(result => {
+       if (result.deletedCount !== 1) {
+         throw "Could not find Mount Doom!";
+       }
+       return incinerateNazgul();
```
nraboy, line 84: Does `incinerateNazgul()` return a promise? Is this function part of the driver?

I realize this is just an example, but to the new user or developer, this may be confusing. It is worth elaborating what this function is or return the following instead:

```
return new Promise((resolve, reject) => {
    // ...
});
```

source/fundamentals/promises.txt
ccho-mongodb: I think this was a team decision made some time ago. Perhaps @terakilobyte or @mbroadst can weigh in on how we want to represent it and then we can change it across all the driver documentation.

source/fundamentals/promises.txt
ccho-mongodb: Would it be clear enough if I renamed the method `incinerateNazgulPromise()` or something to that effect? I wanted to focus this example on chaining Promises rather than show how to declare a Promise.

source/fundamentals/promises.txt
mbroadst: :wave: It should be `Node.js`

source/fundamentals/promises.txt
nraboy: I think that is probably fine. Or add a comment above it that says it is a function that returns a promise.

source/fundamentals/promises.txt
ccho-mongodb: Sounds good, created a ticket to do this across all the guidance docs:
https://jira.mongodb.org/browse/DOCSP-10080

PR #89, opened 2020-04-29
DOCSP-10080: Replace NodeJS with Node.js

ccho-mongodb:
JIRA: 
https://jira.mongodb.org/browse/DOCSP-10080

Staging:
https://docs-mongodborg-staging.corp.mongodb.com/04b6a02/node/docsworker/DOCSP-10080-node-js/

### DIFF
--- a/source/faq.txt
+++ b/source/faq.txt
@@ -42,7 +42,8 @@ How Can I Prevent the Driver From Hanging During Connection or From Spending Too
 To prevent the driver from hanging during connection or to prevent the
 driver from spending too long trying to reach unreachable replica sets,
 you can set the ``connectTimeoutMS`` option of your
-``connection-settings``. Generally, you should ensure that the
+:doc:`connection options </fundamentals/connection#connection-options>`. 
+Generally, you should ensure that the
 ``connectTimeoutMS`` setting is not lower than the longest network
 latency you have to a member of the set. If one of the secondary members
 is on the other side of the planet and has a latency of 10000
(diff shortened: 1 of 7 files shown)


source/faq.txt
```diff
@@ -42,7 +42,8 @@ How Can I Prevent the Driver From Hanging During Connection or From Spending Too
 To prevent the driver from hanging during connection or to prevent the
 driver from spending too long trying to reach unreachable replica sets,
 you can set the ``connectTimeoutMS`` option of your
-``connection-settings``. Generally, you should ensure that the
```
nathan-contino-mongo, line 45: Are connection-settings and connection options two separate things? I feel like a link for "connection-settings" would make sense if they are distinct, and if they are the same we should probably standardize the name and link at the first mention instead of the second. Apologies if these are separate concepts!

source/fundamentals/connection.txt
```diff
@@ -44,8 +44,8 @@ your MongoDB instance.
 
 The last part of the connection string contains connection and authentication
 options as parameters. In the example above, we set two connection options:
-``poolSize=20`` and ``w=majority``. 
-For more information on connection options, skip to the :ref:`Connection Options` section.
```
nathan-contino-mongo, line 48: Should connection options (above) also use a hyphen?

source/faq.txt
ccho-mongodb: I'm not sure where "connection settings" came from. I've replaced them with "connection options" in this PR. Were there other mentions of "connection settings" that I missed?

source/fundamentals/connection.txt
ccho-mongodb: This is a `ref` to a section within the page whereas the other instances are `doc` links that refer to this page.

ccho-mongodb:
> Guess I should have finished my coffee _before_ leaving this review -- I must have looked at the wrong side of the diff in `faq.txt`. Anyway, LGTM now!

NP, thanks for the feedback regardless!

PR #92, opened 2020-05-01
DOCSP-10102: address feedback on limit results

ccho-mongodb:
JIRA:
https://jira.mongodb.org/browse/DOCSP-10102

Staging:
https://docs-mongodbcom-staging.corp.mongodb.com/968780f/node/docsworker/DOCSP-10102-feedback-limit-results/fundamentals/crud/read-operations/limit

I did not include the following feedback items since this is a document on limit, and perhaps would belong in the page on sort instead:
- In the second sort, it now includes an `author`, without any explanation of why. (omitted the second sort instead as it was not necessary)
- For the sort example with `length`, it might be worth mentioning how the results are sorted when multiple lengths are the same.

### DIFF
--- a/source/fundamentals/crud/read-operations.txt
+++ b/source/fundamentals/crud/read-operations.txt
@@ -8,9 +8,9 @@ Read Operations
    :caption: Read Operations
 
    /fundamentals/crud/read-operations/sort
+   /fundamentals/crud/read-operations/skip
    /fundamentals/crud/read-operations/limit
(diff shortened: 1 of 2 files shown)


ccho-mongodb:
I also wonder whether this should omit the skip example since there is a doc for that.

nraboy:
I think it is worth showing how to use `options` as well as the functional approach towards sort, limit, skip, etc.

PR #105, opened 2020-05-18
DOCSP-10420: address feedback on distinct usage example

ccho-mongodb:
JIRA:
https://jira.mongodb.org/browse/DOCSP-10420

Staging:
https://docs-mongodborg-staging.corp.mongodb.com/9377490/node/docsworker/DOCSP-10420-feedback-distinct/usage-examples/distinct

### DIFF
--- a/source/code-snippets/usage-examples/distinct.js
+++ b/source/code-snippets/usage-examples/distinct.js
@@ -22,6 +22,7 @@ async function run() {
     const query = { directors: "Barbra Streisand" };
 
     const distinctValues = await collection.distinct(fieldName, query);
+
     console.log(distinctValues);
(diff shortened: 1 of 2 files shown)


ccho-mongodb:
Addressed:
- show snippets of command for distinct with string parameter and array reference. Did not create one for the option parameter; none of the other usage examples that accept query option parameters provide one since it's not central to the concept and may confuse users looking for the large block of runnable code.

- I did not change the large block (example) as it could be confusing to see multiple calls to `distinct()` which is a guideline the rest of the usage examples adhere to. If we receive feedback that this command is confusing, we can create a new Fundamentals section dedicated to explaining it.

PR #125, opened 2020-08-15
Update delete.txt

rosald:
i think it's greater than 10 and less than 32768 , otherwise it cannot match any


ccho-mongodb:
You're right, thanks!

PR #138, opened 2021-01-27
DOCSP-14153: address feedback on find usage example

ccho-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-14153

https://docs-mongodbcom-staging.corp.mongodb.com/806775e/node/docsworker-xlarge/DOCSP-14153-find-feedback-fix/usage-examples/find

### Self-Review Checklist

### DIFF
--- a/source/code-snippets/usage-examples/find.js
+++ b/source/code-snippets/usage-examples/find.js
@@ -30,6 +30,7 @@ async function run() {
       console.log("No documents found!");
     }
 
+    // replace console.dir with your callback to access individual elements
     await cursor.forEach(console.dir);
   } finally {
     await client.close();
--- a/source/fundamentals/crud/read-operations/cursor.txt
+++ b/source/fundamentals/crud/read-operations/cursor.txt
@@ -29,6 +29,8 @@ and :doc:`Collection.watch() </usage-examples/changeStream>` use
 cursors internally, and return the results of the operations instead of
 a cursor.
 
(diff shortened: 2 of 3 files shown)


source/code-snippets/usage-examples/find.js
```diff
@@ -30,6 +30,7 @@ async function run() {
       console.log("No documents found!");
     }
 
+    // replace console.dir with your callback to access individual elements
```
nathan-contino-mongo, line 33: thought: I wonder if we could demonstrate a custom callback/lambda here that just calls console.dir so that readers could more easily extrapolate this into their own custom logic. Not necessary, and definitely not optimal or good coding, but it could be worth the tradeoff to demonstrate.

nathan-contino-mongo (approved):
LGTM with a thought.

source/code-snippets/usage-examples/find.js
ccho-mongodb: Yes, definitely a tough call! I thought about spelling out the callback rather than using this short form, but remembered we had some conversations about what to put here. I think "console.dir" was a recommendation from the drivers engineer team. It might not have been in the context of "put your own callback here", but rather "this idiomatic JS should serve the purpose of printing results". Since the user feedback appears to be related to the misunderstanding of the forEach parameter and not noticing existing information on the page, I think this way might be less prone to additional feedback related to this item.

PR #145, opened 2021-03-04
DOCSP-14304: document directConnection

terakilobyte:
## Pull Request Info

This documents the `directConnection` parameter in the options list and explains it in the context of replica sets.

Target is all branches so I will cherry-pick this to v3.6 and push.

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-14304

https://docs-mongodbcom-staging.corp.mongodb.com/d40be7e/node/docsworker-xlarge/DOCSP-14304/fundamentals/connection/#connect-to-a-replica-set

### Self-Review Checklist

### DIFF
--- a/source/fundamentals/connection.txt
+++ b/source/fundamentals/connection.txt
@@ -69,6 +69,13 @@ of the ``replicaSet`` parameter in the connection string.
 
    mongodb://host1:27017,host2:27017,host3:27017/?replicaSet=myRs
 
+
+.. note::
+
+   Although it is possible to connect to a replica set deployment by
+   providing only a single host, you should provide the driver with the
+   full list to ensure that it is able to connect in the event the specified host is unavailable.
+
 Direct Connection
 +++++++++++++++++
 
@@ -76,13 +83,6 @@ By default, the driver will discover all members of the replica set given the ad
 and will dispatch operations to the appropriate member, such as writes against the **primary**.
 
 To force operations on the server designated in the connection URI, specify the ``directConnection`` option.
-
-
-.. note::
-
-   Although it is possible to connect to a replica set deployment by
-   providing only a single host, you should provide the driver with the
-   full list to ensure that it is able to connect in the event the specified host is unavailable.
 
 .. _connection-options:
 
--- a/source/fundamentals/connection.txt
+++ b/source/fundamentals/connection.txt
@@ -69,20 +69,26 @@ of the ``replicaSet`` parameter in the connection string.
 
    mongodb://host1:27017,host2:27017,host3:27017/?replicaSet=myRs
 
+When making a connection, the driver takes the following actions by default:
 
-.. note::
+- Discovers all replica set members when given the address of any one member.
+- Dispatches operations to the appropriate member, such as write against the **primary**.
 
-   Although it is possible to connect to a replica set deployment by
-   providing only a single host, you should provide the driver with the
-   full list to ensure that it is able to connect in the event the specified host is unavailable.
+.. tip::
+
+   You only need to specify one host to connect to a replica set. However, to ensure connectivey if the specified host
+   is unavailable, provide the full list of hosts.
 
 Direct Connection
 +++++++++++++++++
 
-By default, the driver will discover all members of the replica set given the address of any one member,
-and will dispatch operations to the appropriate member, such as writes against the **primary**.
+To force operations on the host designated in the connection URI, specify the ``directConnection`` option. Direct
+connections:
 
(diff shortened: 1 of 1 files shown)


nathan-contino-mongo (approved):
LGTM

terakilobyte:
@emadum or @nbbeeken Any issues with this?

emadum (changes requested):
Can we move the Direct Connection section below the note about connecting with multiple seeds? I think it's a little confusing as structured now, because it applies to the section above  and doesn't make sense in the context of direct connection, where you *do* want only one server in your URI.

source/fundamentals/connection.txt
```diff
@@ -69,11 +69,20 @@ of the ``replicaSet`` parameter in the connection string.
 
    mongodb://host1:27017,host2:27017,host3:27017/?replicaSet=myRs
 
+
 .. note::
 
    Although it is possible to connect to a replica set deployment by
    providing only a single host, you should provide the driver with the
-   full list to ensure that it is able to connect even if one host fails.
+   full list to ensure that it is able to connect in the event the specified host is unavailable.
```
atsansone, line 77: **suggestion:** Tighten the language somewhat using second person and imperative mood.
```suggestion
   You only need to provide a single host to connect to a deploy replica set deployment. 
   To ensure connectivity when the specified host is unavailable, provide the driver with
   the full list of hosts
```

source/fundamentals/connection.txt
```diff
@@ -69,11 +69,20 @@ of the ``replicaSet`` parameter in the connection string.
...
+
+Direct Connection
++++++++++++++++++
+
+By default, the driver will discover all members of the replica set given the address of any one member,
+and will dispatch operations to the appropriate member, such as writes against the **primary**.
```
atsansone, line 83: **suggestion:** Change to present tense and split into a bulleted list
```suggestion
When making a direct connection, the driver takes the following actions by default:

- Discovers all replica set members when given the address of any one member.
- Dispatches operations to the appropriate member, such as writes against the **primary**.
```

source/fundamentals/connection.txt
```diff
@@ -69,11 +69,20 @@ of the ``replicaSet`` parameter in the connection string.
...
+
+To force operations on the server designated in the connection URI, specify the ``directConnection`` option.
```
atsansone, line 85: **question:** Should this be `host` to be consistent?
```suggestion
To force operations on the host designated in the connection URI, specify the ``directConnection`` option.
```

source/fundamentals/connection.txt
```diff
@@ -230,6 +239,12 @@ URI to specify the behavior of the client.
        server documentation on the
        :manual:`w Option </reference/write-concern/#w-option>`.
 
+   * - **directConnection**
+     - boolean
+     - ``false``
+     - Specifies whether to force dispatch **all** operations to the server
```
atsansone, line 245: **questions:**

1. Would it make more sense to say "funnels all operations"? This makes it clear it could create a bottleneck or other undesirable condition.
2. Would it not be `host`?

atsansone (changes requested):
@terakilobyte : I phrased as suggestions, but would still like to have a second chance to review. Thanks!

source/fundamentals/connection.txt
terakilobyte: Definitely agree with host. Funnel is incorrect here, it actually dispatches to the host.

source/fundamentals/connection.txt
```diff
@@ -69,11 +69,26 @@ of the ``replicaSet`` parameter in the connection string.
...
-   Although it is possible to connect to a replica set deployment by
-   providing only a single host, you should provide the driver with the
-   full list to ensure that it is able to connect even if one host fails.
+To force operations on the host designated in the connection URI, specify the ``directConnection`` option. Direct
+connections:
+
+- do not support SRV strings.
+- will error on writes when the specified host is not the **primary**.
+- require :manual:`enabling secondary reads </reference/method/rs.secondaryOk/>` when the specified host in not the
+  **primary**.
```
atsansone, line 91: **suggestion:** Clean a little of the language.

```suggestion
- Don't support SRV strings.
- Fails on writes when the specified host is not the **primary**.
- Require :manual:`enabling secondary reads </reference/method/rs.secondaryOk/>`
   when the specified host isn't the **primary**.
```

atsansone (approved):
LGTM % one suggestion.

emadum (approved):
LGTM 👍

PR #174, opened 2021-05-26
DOCSP-16097 Clarifying Intro

kyuan-mongodb:
## Pull Request Info
Clarified the intro of both Usage Examples > Find Operation pages.

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-16097

https://docs-mongodbcom-staging.corp.mongodb.com/7cb12b0/node/docsworker-xlarge/DOCSP-16097-ClarifyQueryOps/usage-examples/findOne/

https://docs-mongodbcom-staging.corp.mongodb.com/7cb12b0/node/docsworker-xlarge/DOCSP-16097-ClarifyQueryOps/usage-examples/find/

### Self-Review Checklist

### DIFF
--- a/source/usage-examples/find.txt
+++ b/source/usage-examples/find.txt
@@ -6,12 +6,12 @@ Find Multiple Documents
 
 You can query for multiple documents in a collection with the
 ``collection.find()`` method. The ``find()`` method uses a query
-document that contains one or more query operators to determine the
-documents to include in the result set. If you don't provide a query
-document or if you provide an empty document, MongoDB returns all
-documents in the collection. For more information on how to 
-specify query operators in a query document, see our
-:doc:`documentation on query documents
+document to determine what to include in the result set. Your query
+document can contain query operators. Query operators are keywords that
+begin with a ``$`` that let you specify complex query behavior. If you
+don't provide a query document or if you provide an empty document,
+MongoDB returns all documents in the collection. For more information on
+how to use query documents and query operators, see our :doc:`guide on query documents
 </fundamentals/crud/query-document>`. 
 
 You can also define additional query options such as
--- a/source/usage-examples/find.txt
+++ b/source/usage-examples/find.txt
@@ -4,13 +4,16 @@ Find Multiple Documents
 
 .. default-domain:: mongodb
 
-You can query for multiple documents in a collection with
-``collection.find()``. The ``find()`` method uses a query document that you
-provide to match the subset of the documents in the collection that match the
-query. If you don't provide a query document (or if you provide an empty
-document), MongoDB returns all documents in the collection. For more
-information on querying MongoDB, see our
-:doc:`documentation on query documents </fundamentals/crud/query-document>`.
+You can query for multiple documents in a collection by passing a
+query document to the ``collection.find()`` method. A query document
+contains one or more query operators that apply to specific fields which
+determine the documents to include in the result set. For more
+information, see our :doc:`documentation on query documents </fundamentals/crud/query-document>`.
+
+.. note::
+
+  If you don't provide a query document (or if you provide an empty
+  document), MongoDB returns all documents in the collection. 
 
 You can also define additional query options such as
 :doc:`sort </fundamentals/crud/read-operations/sort>`
@@ -58,11 +61,11 @@ following:
 
 .. code-block:: javascript
 
-   { title: '10 Minutes', imdb: { rating: 7.9, votes: 743, id: 339976 } }
-   { title: '3x3', imdb: { rating: 6.9, votes: 206, id: 1654725 } }
-   { title: '7:35 in the Morning', imdb: { rating: 7.3, votes: 1555, id: 406501 } }
-   { title: '8', imdb: { rating: 7.8, votes: 883, id: 1592502 } }
-   ...
+  { title: '10 Minutes', imdb: { rating: 7.9, votes: 743, id: 339976 } }
+  { title: '3x3', imdb: { rating: 6.9, votes: 206, id: 1654725 } }
+  { title: '7:35 in the Morning', imdb: { rating: 7.3, votes: 1555, id: 406501 } }
+  { title: '8', imdb: { rating: 7.8, votes: 883, id: 1592502 } }
+  ...
 
 The ``sort`` and ``projection`` options can also be specified as methods
 (``sort()`` and ``project()``, respectively) chained to the ``findOne`` method.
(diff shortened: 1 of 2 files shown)


source/usage-examples/find.txt
```diff
@@ -58,17 +61,17 @@ following:
...
-   { title: '10 Minutes', imdb: { rating: 7.9, votes: 743, id: 339976 } }
-   { title: '3x3', imdb: { rating: 6.9, votes: 206, id: 1654725 } }
-   { title: '7:35 in the Morning', imdb: { rating: 7.3, votes: 1555, id: 406501 } }
-   { title: '8', imdb: { rating: 7.8, votes: 883, id: 1592502 } }
-   ...
+  { title: '10 Minutes', imdb: { rating: 7.9, votes: 743, id: 339976 } }
+  { title: '3x3', imdb: { rating: 6.9, votes: 206, id: 1654725 } }
+  { title: '7:35 in the Morning', imdb: { rating: 7.3, votes: 1555, id: 406501 } }
+  { title: '8', imdb: { rating: 7.8, votes: 883, id: 1592502 } }
+  ...
```
biniona-mongodb, line 68: issue: My eyes may be playing tricks on me, but I think this was formatted correctly before. [I believe these are the lines in the file for this PR](https://github.com/mongodb/docs-node/blob/c973085ce4b297f2e5a53c961aa661d6af94cfa8/source/usage-examples/find.txt#L62-L68), it looks like the edit turns this into 2 spaces where as before it was 3.

I feel the [current pages in the master branch are](https://github.com/mongodb/docs-node/blob/master/source/usage-examples/find.txt) correct, and these edits introduce issues.

This applies to all instances of moving the indents in this PR.

source/usage-examples/find.txt
```diff
@@ -4,13 +4,16 @@ Find Multiple Documents
...
+You can query for multiple documents in a collection by passing a
+query document to the ``collection.find()`` method. A query document
+contains one or more query operators that apply to specific fields which
+determine the documents to include in the result set. For more
+information, see our :doc:`documentation on query documents </fundamentals/crud/query-document>`.
+
+.. note::
+
+  If you don't provide a query document (or if you provide an empty
+  document), MongoDB returns all documents in the collection.
```
biniona-mongodb, line 16: Suggestion: I think this content should not be broken out of the paragraph. I don't have supporting evidence for this unfortunately, but in previous reviews I was informed that readers are less likely to read content in notes. I think it is important that readers understand this behavior.

Applies to both instances of this in the PR.

source/usage-examples/find.txt
```diff
@@ -4,13 +4,16 @@ Find Multiple Documents
...
 
-You can query for multiple documents in a collection with
-``collection.find()``. The ``find()`` method uses a query document that you
-provide to match the subset of the documents in the collection that match the
-query. If you don't provide a query document (or if you provide an empty
-document), MongoDB returns all documents in the collection. For more
-information on querying MongoDB, see our
-:doc:`documentation on query documents </fundamentals/crud/query-document>`.
+You can query for multiple documents in a collection by passing a
+query document to the ``collection.find()`` method. A query document
```
biniona-mongodb, line 8:
```suggestion
You can query for multiple documents in a collection with the ``collection.find()`` method. A query document
```

Issue: I think was worded correctly before. As I understand it you are not required to pass a document, so I believe it is not correct to say that "passing a document causes the query to occur", so I think this should be reverted to the original wording.

Applies to both instances.

biniona-mongodb (changes requested):
Overall looks good. My eyes may be playing tricks on me, but I have a few larger asks.

source/usage-examples/find.txt
kyuan-mongodb: Apologies, I went to check and you are right. The pipe showing up made me think it was off, but it's just how it is in the node docs.

source/usage-examples/find.txt
```diff
@@ -4,13 +4,15 @@ Find Multiple Documents
...
-``collection.find()``. The ``find()`` method uses a query document that you
-provide to match the subset of the documents in the collection that match the
-query. If you don't provide a query document (or if you provide an empty
-document), MongoDB returns all documents in the collection. For more
-information on querying MongoDB, see our
-:doc:`documentation on query documents </fundamentals/crud/query-document>`.
+You can query for multiple documents in a collection with the
+``collection.find()`` method. The ``find()`` method uses a query
+document that contains one or more query operators to determine the
+documents to include in the result set. If you don't provide a query
```
biniona-mongodb, line 10:
```suggestion
document  to determine what to include in the result set. Your query document may contain query operators. Query operators are keywords that begin with a ``$`` that let you specify complex query behavior. If you don't provide a query
```

Issue: I don't believe a query document must contain a query operator ({"_id":1} I believe is a valid query document). I think query operators are optional. 

Suggestion: I feel it could be helpful to define the term query operator. 

Sources for suggested definition above:

[Operator Definition
](https://docs.mongodb.com/manual/reference/glossary/#std-term-operator)
[Query Operator Definition](https://docs.mongodb.com/manual/reference/operator/)

This suggestion applies to both pages

source/usage-examples/find.txt
```diff
@@ -4,13 +4,15 @@ Find Multiple Documents
...
+document or if you provide an empty document, MongoDB returns all
+documents in the collection. For more information on how to 
+specify query operators in a query document, see our
+:doc:`documentation on query documents
```
biniona-mongodb, line 14:
```suggestion
:doc:`guide on query documents
```

Nit: I think in other parts of the drivers docs we refer to our own pages as guides. Also the phrase "documentation on query documents" sounds off to me with the double use of "document". 

Applies to both instances

source/usage-examples/find.txt
```diff
@@ -4,13 +4,15 @@ Find Multiple Documents
...
-document), MongoDB returns all documents in the collection. For more
-information on querying MongoDB, see our
-:doc:`documentation on query documents </fundamentals/crud/query-document>`.
+You can query for multiple documents in a collection with the
+``collection.find()`` method. The ``find()`` method uses a query
+document that contains one or more query operators to determine the
+documents to include in the result set. If you don't provide a query
+document or if you provide an empty document, MongoDB returns all
+documents in the collection. For more information on how to 
+specify query operators in a query document, see our
```
biniona-mongodb, line 13:
```suggestion
use query documents and query operators, see our
```
Nit: I think this rephrasing more accurately describes the linked page (more about query documents than query operators). 

Applies to both instances.

biniona-mongodb (changes requested):
Looking really good, super close. One bigger issue and two smaller nits.

source/usage-examples/find.txt
kyuan-mongodb: Well the query document is the BSON we usually call "filter". And to have `{"_id":1}`, you'd need to use `$eq`, which is a query operator.

biniona-mongodb: Discussed synchronously. Edit clarified. Agree Query document does not necessarily contain update operator.

biniona-mongodb (approved):
LGTM!

PR #213, opened 2021-07-16
DOCSP:13409 aws session token auth

rustagir:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-13409

https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-13409-aws-session-token-auth/fundamentals/authentication/mechanisms/#mongodb-aws

### Self-Review Checklist

### DIFF
--- a/source/code-snippets/authentication/aws.js
+++ b/source/code-snippets/authentication/aws.js
@@ -26,7 +26,7 @@ async function run() {
     await client.db("admin").command({ ping: 1 });
     console.log("Connected successfully to server.");
   } finally {
-    // Ensure that the client will close when you finish/error.
+    // Ensure that the client closes when it finishes/errors.
     await client.close();
   }
 }
--- a/source/code-snippets/authentication/aws.js
+++ b/source/code-snippets/authentication/aws.js
@@ -4,32 +4,30 @@ const { MongoClient } = require("mongodb");
 const accessKeyId = encodeURIComponent("<AWS_ACCESS_KEY_ID>");
 const secretAccessKey = encodeURIComponent("<AWS_SECRET_ACCESS_KEY>");
 const clusterUrl = "<MongoDB cluster url>";
-
-const sessionToken = encodeURIComponent("<AWS_SESSION_TOKEN>"); // Optional
-
 const authMechanism = "MONGODB-AWS";
 
 // Replace the following with your MongoDB deployment's connection string.
-const uri =
+let uri =
   `mongodb+srv://${accessKeyId}:${secretAccessKey}@${clusterUrl}/?authMechanism=${authMechanism}`;
 
-// If you created a session token, append to the end of your URI string
-const tokenSetting = '&authMechanismProperties=${sessionToken}';
+// Uncomment the following lines if your AWS authentication setup requires a session token.
+// const sessionToken = encodeURIComponent("<AWS_SESSION_TOKEN>");
+// uri = uri.concat(`&authMechanismProperties=${sessionToken}`);
 
-// Create a new MongoClient
+// Create a new MongoClient.
 const client = new MongoClient(uri);
 
-// Function to connect to the server
+// Function to connect to the server.
 async function run() {
   try {
-    // Connect the client to the server
+    // Connect the client to the server.
     await client.connect();
 
-    // Establish and verify connection
+    // Establish and verify connection.
     await client.db("admin").command({ ping: 1 });
     console.log("Connected successfully to server");
   } finally {
-    // Ensures that the client will close when you finish/error
+    // Ensure that the client will close when you finish/error.
     await client.close();
   }
 }
--- a/source/code-snippets/authentication/aws.js
+++ b/source/code-snippets/authentication/aws.js
@@ -4,9 +4,9 @@ const { MongoClient } = require("mongodb");
 const accessKeyId = encodeURIComponent("<AWS_ACCESS_KEY_ID>");
 const secretAccessKey = encodeURIComponent("<AWS_SECRET_ACCESS_KEY>");
 const clusterUrl = "<MongoDB cluster url>";
+
 const authMechanism = "MONGODB-AWS";
 
-// Replace the following with your MongoDB deployment's connection string.
 let uri =
   `mongodb+srv://${accessKeyId}:${secretAccessKey}@${clusterUrl}/?authMechanism=${authMechanism}`;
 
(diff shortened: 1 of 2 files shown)


rustagir:
I just saw the comments on the JIRA ticket suggesting that this information could be best imparted inside of the full connection example with a comment indicating that the AWS session token is optional to include-- let me know if that would be a better approach to what I did!

source/code-snippets/authentication/aws.js
```diff
@@ -5,12 +5,17 @@ const accessKeyId = encodeURIComponent("<AWS_ACCESS_KEY_ID>");
...
+const sessionToken = encodeURIComponent("<AWS_SESSION_TOKEN>"); // Optional
+
 const authMechanism = "MONGODB-AWS";
 
 // Replace the following with your MongoDB deployment's connection string.
 const uri =
   `mongodb+srv://${accessKeyId}:${secretAccessKey}@${clusterUrl}/?authMechanism=${authMechanism}`;
 
+// If you created a session token, append to the end of your URI string
+const tokenSetting = '&authMechanismProperties=${sessionToken}';
```
ccho-mongodb, line 17: A couple items here:

1) I think for string interpolation (i.e. using the ${} syntax to parse a value into a string) you need backticks similar to what's on line 14. Also, we should probably show how to append it rather than storing it in a variable that doesn't get used. 

2) The comment should answer the question of when they need it more precisely. While you can create a session token on AWS, that is independent of whether you need it for authentication. Instead, how about something like "Uncomment the following line if your AWS authentication setup requires a session token." and comment out this line.

Putting these together, your code might look something like this:
```
// Uncomment the following line if your AWS authentication setup requires a session token 
// uri.concat(`&authMechanismProperties=${sessionToken}`);
```

source/code-snippets/authentication/aws.js
```diff
@@ -5,12 +5,17 @@ const accessKeyId = encodeURIComponent("<AWS_ACCESS_KEY_ID>");
 const secretAccessKey = encodeURIComponent("<AWS_SECRET_ACCESS_KEY>");
 const clusterUrl = "<MongoDB cluster url>";
 
+const sessionToken = encodeURIComponent("<AWS_SESSION_TOKEN>"); // Optional
```
ccho-mongodb, line 8: Let's follow the same format as the other code comments for this line and help the reader understand why it's "optional" by explaining in what case they would need it. Since it's optional, maybe it should be commented out and the comment preceding it should also mention that they should uncomment the line if they need it.

source/fundamentals/authentication/mechanisms.txt
```diff
@@ -139,7 +139,7 @@ To connect to a MongoDB instance with ``MONGODB-AWS`` authentication
 enabled, specify the ``MONGODB-AWS`` authentication mechanism and pass
 your ``AWS_ACCESS_KEY_ID`` and ``AWS_SECRET_ACCESS_KEY`` credentials to the
 driver when you attempt to connect. If your AWS login requires a session
-token, you must include your ``AWS_SESSION_TOKEN`` as well.
+token, you must include your ``AWS_SESSION_TOKEN`` as well.
```
ccho-mongodb, line 142: Nitpick: no reason to add a trailing space (I usually use a hotkey I set up in my text editor to strip all trailing spaces) in rst.

source/code-snippets/authentication/aws.js
```diff
@@ -5,12 +5,17 @@ const accessKeyId = encodeURIComponent("<AWS_ACCESS_KEY_ID>");
 const secretAccessKey = encodeURIComponent("<AWS_SECRET_ACCESS_KEY>");
 const clusterUrl = "<MongoDB cluster url>";
```
ccho-mongodb, line 6: I know this isn't part of your PR, but if this variable isn't used, can you remove it?

ccho-mongodb (changes requested):
Looks pretty good, thanks for correcting the punctuation and capitalization. Could you make the requested changes and send it back to me once you're done? Thanks.

source/code-snippets/authentication/aws.js
rustagir: accepted. changed to reflect that concat does not change strings in place. I also had to use "let" to initialize because const types are not editable.

source/code-snippets/authentication/aws.js
rustagir: accepted

source/fundamentals/authentication/mechanisms.txt
rustagir: accepted

source/code-snippets/authentication/aws.js
rustagir: it looks like all of the variables are being used

source/code-snippets/authentication/aws.js
```diff
@@ -7,24 +7,26 @@ const clusterUrl = "<MongoDB cluster url>";
...
 
-// Replace the following with your MongoDB deployment's connection string.
-const uri =
+let uri =
   `mongodb+srv://${accessKeyId}:${secretAccessKey}@${clusterUrl}/?authMechanism=${authMechanism}`;
 
-// Create a new MongoClient
+// Uncomment the following lines if your AWS authentication setup requires a session token.
+// const sessionToken = encodeURIComponent("<AWS_SESSION_TOKEN>");
+// uri = uri.concat(`&authMechanismProperties=${sessionToken}`);
```
ccho-mongodb, line 15: I don't have a strong preference because it's correct and (I think) idiomatic JS, but another common way to accomplish this is:
```
uri += `&authMechanismProperties=${sessionToken}`;
```

source/code-snippets/authentication/aws.js
```diff
@@ -7,24 +7,26 @@ const clusterUrl = "<MongoDB cluster url>";
...
     await client.connect();
 
-    // Establish and verify connection
+    // Establish and verify connection.
     await client.db("admin").command({ ping: 1 });
-    console.log("Connected successfully to server");
+    console.log("Connected successfully to server.");
   } finally {
-    // Ensures that the client will close when you finish/error
+    // Ensure that the client will close when you finish/error.
```
ccho-mongodb, line 29: Nitpick: I think we can improve the prior comment by properly identifying the object. In this case, "you" is used rather than the "client" which is actually taking the action.  Also, prefer present tense to future for clarity. E.g.
"Ensure the client closes when it finishes/errors"

ccho-mongodb (approved):
LGTM, just one nitpick.

source/code-snippets/authentication/aws.js
rustagir: accepted

PR #223, opened 2021-08-05
(DOCSP-17936) FindOne TypeScript example

biniona-mongodb:
## Pull Request Info
Add typescript example for Find One. 

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-15236

https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-17936-Find-One/usage-examples/findOne/
### Self-Review Checklist

### DIFF
--- a/source/code-snippets/usage-examples/findOne.ts
+++ b/source/code-snippets/usage-examples/findOne.ts
@@ -23,12 +23,12 @@ async function run(): Promise<void> {
 
     const database = client.db("sample_mflix");
     const movies = database.collection<MovieSchema>("movies");
-    const query = { title: "The Room" } as Document;
+    const query: object = { title: "The Room" };
 
-    const options = {
+    const options: FindOptions = {
       sort: { rating: -1 },
       projection: { _id: 0, title: 1, imdb: 1 },
-    } as FindOptions;
+    };
 
     const movie = await movies.findOne(query, options);
     console.log(movie);
--- a/source/code-snippets/usage-examples/findOne.ts
+++ b/source/code-snippets/usage-examples/findOne.ts
@@ -26,7 +26,8 @@ async function run(): Promise<void> {
     await client.connect();
 
     const database = client.db("sample_mflix");
-    // Specifying a Schema is always optional, but it enables type hinting on finds and inserts
+    // Specifying a Schema is always optional, but it enables type hinting on
+    // finds and inserts
     const movies = database.collection<Movie>("movies");
 
     const movie = await movies.findOne({ title: "The Room" }, {
--- a/source/code-snippets/usage-examples/findOne.ts
+++ b/source/code-snippets/usage-examples/findOne.ts
@@ -12,9 +12,13 @@ interface IMDB {
   id: number;
 }
 
-interface MovieSchema extends Document {
+export interface Movie {
   title: string;
-  imdb: imdb;
+  year: number;
+  released: Date;
+  plot: string;
+  type: 'movie' | 'tv' | 'web series';
+  imdb: IMDB;
 }
 
 async function run(): Promise<void> {
--- a/source/code-snippets/usage-examples/findOne.ts
+++ b/source/code-snippets/usage-examples/findOne.ts
@@ -6,11 +6,11 @@ const uri =
 
 const client = new MongoClient(uri);
 
-type imdb = {
+interface IMDB {
   rating: number;
   votes: number;
   id: number;
-};
+}
 
 interface MovieSchema extends Document {
   title: string;
--- a/source/code-snippets/usage-examples/findOne.ts
+++ b/source/code-snippets/usage-examples/findOne.ts
@@ -17,7 +17,7 @@ export interface Movie {
   year: number;
   released: Date;
   plot: string;
-  type: 'movie' | 'tv' | 'web series';
+  type: 'movie' | 'series';
   imdb: IMDB;
 }
 
--- a/source/code-snippets/usage-examples/findOne.ts
+++ b/source/code-snippets/usage-examples/findOne.ts
@@ -1,0 +1,39 @@
+import { FindOptions, MongoClient } from "mongodb";
+
+// Replace the uri string with your MongoDB deployment's connection string.
+const uri =
+  "mongodb+srv://<user>:<password>@<cluster-url>?writeConcern=majority";
+
+const client = new MongoClient(uri);
+
+type imdb = {
+  rating:number,
+  votes:number,
+  id:number
+}
+
+interface MovieSchema extends Document {
+  title:string,
+  imdb:imdb
+}
+
+async function run(): Promise<void> {
+  try {
+    await client.connect();
+
+    const database = client.db("sample_mflix");
+    const movies = database.collection<MovieSchema>("movies");
+    const query = { title: "The Room" } as Document;
+
+    const options = {
+      sort: { rating: -1 },
+      projection: { _id: 0, title: 1, imdb: 1 },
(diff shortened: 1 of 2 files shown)


sarahsimpers (approved):
LGTM

biniona-mongodb:
Hey @nbbeeken and @emadum! This PR is for adding a TypeScript equivalent to the FindOne usage example. Apologies in advance, I am pretty new to TypeScript, so please let me know if I missed the mark on anything.

source/code-snippets/usage-examples/findOne.ts
```diff
@@ -0,0 +1,40 @@
...
+const uri =
+  "mongodb+srv://<user>:<password>@<cluster-url>?writeConcern=majority";
+
+const client = new MongoClient(uri);
+
+type imdb = {
+  rating: number;
+  votes: number;
+  id: number;
+};
```
nbbeeken, line 13:
```suggestion
interface IMDB {
  rating: number;
  votes: number;
  id: number;
}
```

Nothing wrong with using `type` here but might as well be consistent and use an `interface`. Also regardless is common that types start with a capital letter.

source/code-snippets/usage-examples/findOne.ts
```diff
@@ -0,0 +1,40 @@
...
+
+interface MovieSchema extends Document {
+  title: string;
+  imdb: imdb;
+}
```
nbbeeken, line 18:
```suggestion
export interface Movie {
  title: string;
  year: number;
  released: Date;
  plot: string;
  type: 'movie' | 'tv' | 'web series';
  imdb: IMDB;
}
```

I think showing more variety of types makes it a little more realistic but understandable if you want to keep the example terse.

You don't have to `extends Document` here, that makes the schema support any key, which may be desirable sometimes but not for this example.

source/code-snippets/usage-examples/findOne.ts
```diff
@@ -0,0 +1,40 @@
...
+
+async function run(): Promise<void> {
+  try {
+    await client.connect();
+
+    const database = client.db("sample_mflix");
+    const movies = database.collection<MovieSchema>("movies");
```
nbbeeken, line 25:
```suggestion
    // Specifying a Schema is always optional, but it enables type hinting on finds and inserts
    const movies = database.collection<MovieSchema>("movies");
```

I think its good that we'll be showing TS examples next to JS but it should be made clear that none of the TS features are requirements for TS users, the JS examples are valid TS examples.

source/code-snippets/usage-examples/findOne.ts
```diff
@@ -0,0 +1,40 @@
...
+    const query: object = { title: "The Room" };
```
nbbeeken, line 26:
```suggestion
    const query = { title: "The Room" };
```

No need for the `object` annotation.

source/code-snippets/usage-examples/findOne.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    const options: FindOptions = {
+      sort: { rating: -1 },
+      projection: { _id: 0, title: 1, imdb: 1 },
+    };
+
+    const movie = await movies.findOne(query, options);
```
nbbeeken, line 33:
```suggestion
    const movie = await movies.findOne({ title: "The Room" }, {
      sort: { rating: -1 },
      projection: { _id: 0, title: 1, imdb: 1 },
    });
```

I think we should inline the query and options in the TS example because it is how user's get type hinting working without much intervention (like importing FindOptions). Copying and pasting this ^ example and playing with the usage will reveal that `title` is forced to be a `string`. Or even that `title` can map to one of the many query operators.

source/code-snippets/usage-examples/findOne.ts
```diff
@@ -0,0 +1,40 @@
+import { FindOptions, MongoClient } from "mongodb";
```
nbbeeken, line 1:
```suggestion
import { MongoClient } from "mongodb";
```

If we go with my other suggestion of inlining the usage.

nbbeeken (changes requested):

source/code-snippets/usage-examples/findOne.ts
biniona-mongodb: Sounds good, will try to remember to capitalize types and be consistent with `type` and `interface` usage throughout these usage examples.

source/code-snippets/usage-examples/findOne.ts
biniona-mongodb: Sounds good, thanks for the note. I'll implement the suggested change.

The rest of this is purely for my own edification, feel free to ignore:

Sorry if this is a silly question, but I saw in the [collection method code](https://github.com/mongodb/node-mongodb-native/blob/b67af3cd/src/db.ts#L323) this generic type which made me think that the type passed to collection had to extend document:

>  <TSchema extends Document = Document>

Based off of this [stack overflow post](https://stackoverflow.com/questions/56843790/typescript-generic-type-with-equal-operator-means) I interpreted this as: The default type is document, and if you provide a type, it must extend document. I was surprised that you didn't need to extend `Document` in the `Movie` interface. Any tips on how I am misunderstanding this line of code?

source/code-snippets/usage-examples/findOne.ts
biniona-mongodb: Good point! I'll try and communicate that TS features are optional in future TS rewrites of usage examples.

source/code-snippets/usage-examples/findOne.ts
biniona-mongodb: Sounds good, I'll try to use inline queries as a means to get TS features working without much code intervention in future usage example rewrites.

source/code-snippets/usage-examples/findOne.ts
```diff
@@ -0,0 +1,42 @@
...
+    await client.connect();
+
+    const database = client.db("sample_mflix");
+    // Specifying a Schema is always optional, but it enables type hinting on
+    // finds and inserts
+    const movies = database.collection<Movie>("movies");
+
+    const movie = await movies.findOne({ title: "The Room" }, {
+      sort: { rating: -1 },
+      projection: { _id: 0, title: 1, imdb: 1 },
```
nbbeeken, line 35: Using a projection and getting accurate type information requires you to define the resulting type of your query. So we could simplify this example by dropping the projection or we can describe some expected/best practice here. 

Projections are often used to make a query directly return an edited document that's acceptable to send out over a REST api etc. So this is what someone could do:

```typescript
interface MovieSummary {
  title: string;
  imdb: IMDB;
}
// or 
type MovieSummary = Pick<Movie, 'title' | 'imdb'>

// ...

const movieSummary = await movies.findOne<MovieSummary>({ title: "The Room" }, {
  sort: { rating: -1 },
  projection: { _id: 0, title: 1, imdb: 1 },
}); // Returns FindCursor<MovieSummary>
```

The `Pick<>` is one of Typescript's utility types: https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys Pick and Omit lend themselves well to quickly defining an accurate projection type.

nbbeeken (changes requested):

source/code-snippets/usage-examples/findOne.ts
biniona-mongodb: Oh nice! That seems super helpful for projections. Adding the suggested change now using `Pick`.

nbbeeken (approved):
LGTM

PR #229, opened 2021-08-09
(DOCSP-17936) Bulk TS Example

biniona-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-17936

https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-17936-Bulk-TS/usage-examples/bulkWrite/

### Self-Review Checklist

### DIFF
--- a/source/code-snippets/usage-examples/bulkWrite.ts
+++ b/source/code-snippets/usage-examples/bulkWrite.ts
@@ -56,6 +56,7 @@ async function run() {
       },
       {
         updateMany: {
+          // Important: You lose type safety when you use dot notation in queries
           filter: { "location.address.zipcode": "44011" },
           update: { $set: { is_in_ohio: true } },
           upsert: true,
--- a/source/code-snippets/usage-examples/bulkWrite.ts
+++ b/source/code-snippets/usage-examples/bulkWrite.ts
@@ -7,14 +7,15 @@ const uri =
 const client = new MongoClient(uri);
 
 interface Address {
-  street1: string,
-  city: string,
-  state: string,
-  zipcode: string,
+  street1: string;
+  city: string;
+  state: string;
+  zipcode: string;
 }
 
 interface Theater {
-  location: {address: Address}
+  location: { address: Address };
+  is_in_ohio?: boolean;
 }
 
 async function run() {
@@ -56,12 +57,14 @@ async function run() {
       {
         updateMany: {
           filter: { "location.address.zipcode": "44011" },
-          update: { $set: { street2: "25th Floor" } },
+          update: { $set: { is_in_ohio: true } },
           upsert: true,
         },
       },
       {
-        deleteOne: { filter: { "location.address.street1": "221b Baker St" } },
+        deleteOne: {
+          filter: { "location.address.street1": "221b Baker St" },
+        },
       },
     ]);
 
--- a/source/code-snippets/usage-examples/bulkWrite.ts
+++ b/source/code-snippets/usage-examples/bulkWrite.ts
@@ -1,0 +1,53 @@
+import { MongoClient } from "mongodb";
+
+// Replace the uri string with your MongoDB deployment's connection string.
+const uri =
+  "mongodb+srv://<user>:<password>@<cluster-url>?writeConcern=majority";
+
+const client = new MongoClient(uri);
+
+async function run() {
+  try {
+    await client.connect();
+
+    const database = client.db("sample_mflix");
+    const theaters = database.collection("theaters");
+
+    const result = await theaters.bulkWrite([
+      { insertOne:
+        {
+          "document": {
+            location: {
+              address: { street1: '3 Main St.', city: 'Anchorage', state: 'AK', zipcode: '99501' },
+            }
+          }
+        }
+      },
+      { insertOne:
+        {
+          "document": {
+            location: {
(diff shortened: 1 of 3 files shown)


terakilobyte (approved):
LGTM. Is it intentional the only change is the import statement?

biniona-mongodb:
Yes, the Intellisense (I think that is what the VSCode feature is called) Typing UX was not improved by adding any additional Typing code to the example, so felt best to not add anything. The import statement was changed as from what I can tell TypeScript requires the [@types/node](https://www.npmjs.com/package/@types/node) package in order to use the 

`const { MongoClient } = require("mongodb");`

So it seemed easier for readers to change the import statment so they would not need to install another package.

biniona-mongodb:
Hey @mongodb/dbx-node, could I have a review for this TypeScript usage example?

nbbeeken:
If someone is using TS and working on a nodejs project then it is an implicit requirement to have `@types/node` installed. So you can assume it is present, we don't make it a dependency of the driver because the only place for it would be in production dependencies which is never the case for type definitions.
Maybe this is something we can callout in documentation to be explicit. (`npm i @types/node` is assumed etc...)

---

As for the import style we put at the top of the examples (maybe only on some top-level page?), I think we can maybe also hoist that to a parenthetical something along the lines of:

> When using typescript or ES modules you can import from the driver like so:
> ```ts
> import { MongoClient } from 'mongodb'
> ```
> 
> When using javascript or CommonJS modules you can import from the driver like so:
> ```js
> const { MongoClient } = require('mongodb')
> ```
> 
> All examples use the es module style.

And then, as the last statement implies, we can omit explicit TS examples where the only change would be the import.

source/code-snippets/usage-examples/bulkWrite.ts
```diff
@@ -0,0 +1,53 @@
...
+          "document": {
+            location: {
+              address: { street1: '75 Penn Plaza', city: 'New York', state: 'NY', zipcode: '10001' },
+            }
+          }
+        }
+      },
+      { updateMany:
+        {
+          "filter": { "location.address.zipcode" : "44011" },
```
nbbeeken, line 37: nit: Wanna run this through the formatter? the stringified keys are just not necessary

source/code-snippets/usage-examples/bulkWrite.ts
```diff
@@ -0,0 +1,53 @@
...
+  "mongodb+srv://<user>:<password>@<cluster-url>?writeConcern=majority";
+
+const client = new MongoClient(uri);
+
+async function run() {
+  try {
+    await client.connect();
+
+    const database = client.db("sample_mflix");
+    const theaters = database.collection("theaters");
```
dariakp, line 14: I think it may also be worthwhile to add a type to this collection, because the insert/update operations is where typescript is actually helpful in terms of hinting which properties are expected on an object

dariakp (changes requested):

source/code-snippets/usage-examples/bulkWrite.ts
biniona-mongodb: Ran JS and TS examples through `prettier`!

source/code-snippets/usage-examples/bulkWrite.ts
biniona-mongodb: Added  a type to the collection and I think it improves the UX! Good catch. 

I also change fields specified like:

> `field.subfield.another_subfield : "value"`

To:

> `{field : {subfield : { another_subfield : "value"}}}`

As typescript did not seem to able to determine the correct type of the field with the first syntax.

source/code-snippets/usage-examples/bulkWrite.ts
nbbeeken: Nice! If you wanted those address objects to still be on one line you can increase the `--print-width` of prettier. But I think it looks fine either way

biniona-mongodb:
Good idea @nbbeeken! I added the following items to a[ Cleanup ticket](https://jira.mongodb.org/browse/DOCSP-18084) to be complete once all TypeScript examples have been merged in:

- Change All Node Usage Examples (TS and JS) to use ES Module Import Syntax
- Add note to "Usage Examples" Page explaining that Imports in Usage Examples in line [with this Github Review Comment](https://github.com/mongodb/docs-node/pull/229#issuecomment-896298377)

source/code-snippets/usage-examples/bulkWrite.ts
biniona-mongodb: Thanks for the tip! Will keep the `--print-width` flag in mind in the future

source/code-snippets/usage-examples/bulkWrite.ts
dariakp: Unfortunately there are no typescript solutions for this that would not result in unacceptably long compilation times for our users' projects. We should continue advocating for using dot notation in queries, just warn about the inevitable loss of type safety.

dariakp (approved):
Updated examples look good, thank you!

source/code-snippets/usage-examples/bulkWrite.ts
biniona-mongodb: Sounds good, thanks for looking into this. I'll add a note under the example warning about loss of type safety with dot notation.

nbbeeken (approved):

PR #233, opened 2021-08-09
(DOCSP-17936) Delete One TS Example

biniona-mongodb:
## Pull Request Info

Add TS Variant for Delete One Usage Example

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-17936

https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-17936-Delete-One-TS/usage-examples/deleteOne/

### Self-Review Checklist


terakilobyte (approved):
LGTM, though this is a really strange operation overall since it'd be pretty hard to know which movie it was deleting.

biniona-mongodb:
@terakilobyte I agree. I'll make a ticket to update this usage example to do something that is easier to predict.

biniona-mongodb:
Hey @mongodb/dbx-node, could I have a review for this TypeScript usage example?

source/code-snippets/usage-examples/deleteOne.ts
```diff
@@ -0,0 +1,26 @@
...
+const client = new MongoClient(uri);
+
+async function run() {
+  try {
+    await client.connect();
+
+    const database = client.db("sample_mflix");
+    const movies = database.collection("movies");
+
+    const result = await movies.deleteOne({ title: { $type: "string" } });
```
dariakp, line 16: we don't think this TS example is necessary since it is identical to the `.js` one and there is no added value from TS over plain JS for this use case

dariakp (changes requested):

source/code-snippets/usage-examples/deleteOne.ts
biniona-mongodb: Thanks for the review! While I agree using TypeScript for this use case does not provide much value for readers, I think it still makes sense to include TS (or a note about TS) for this example. Here are my reasons for including it:

- Including TS communicates explicitly rather than implicitly that the TS is the same as the JS for this use case. I think if the TS example is removed readers will have to infer that a lack of TS means that there is no benefit to using TS here, which I think may be difficult for Jasmine persona (early developer) readers.
- Including TS provides a consistent experience across usage examples (all examples have a JS and TS component to them). I think the usage examples could look incomplete if some examples have a TS variant and some do not.

I agree that right now it is not clear that the TS and JS are identical and that readers receive no substantive benefits from using TS here. Do any of these options address this concern?

1. Include a comment in the TS example saying that the "TS and JS snippets are identical for this use case"
2. Add a note admonition above the code snippet saying "The TS and JS Snippets are Identical For this Use Case"
3. Include an admonition below examples where TS does not benefit readers saying, "TS does not provide significant value in this use case". 

Let me know if you disagree or would like to discuss this over slack!

source/usage-examples/deleteOne.txt
```diff
@@ -45,5 +45,30 @@ to match only movies with a title of type ``string``.
...
+     :tabid: typescript
+
+     .. literalinclude:: /code-snippets/usage-examples/deleteOne.js
+        :language: javascript
+        :linenos:
+
+.. note:: Identical Code Snippets
+
+   The JavaScript and TypeScript code snippets above are identical. There is no
+   added value from using TypeScript in this use case.
```
dariakp, line 67: I would suggest that instead of saying "There is no added value from using TypeScript" we say "There are no specific TypeScript features relevant to this use case", or something along those lines. (TypeScript will still provide type hinting in the editor, just nothing that the developers need to actively think about in this example)

dariakp (changes requested):

source/usage-examples/deleteOne.txt
biniona-mongodb: Good rephrase! How is "There are no TypeScript specific features of the driver relevant to this use case"?

dariakp: Sure, sounds good to me :)

dariakp (approved):

PR #243, opened 2021-08-24
DOCSP-17991 4.0-Code-Snippets

kyuan-mongodb:
## Pull Request Info

- Tested all code snippets. No changes were needed, but I removed information about the insertCount for insertOne since it's not longer a property.
- Made all output not copyable.

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-17991

https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/4.0-Code-Snippets/

### Self-Review Checklist

### DIFF
--- a/source/fundamentals/monitoring/cluster-monitoring.txt
+++ b/source/fundamentals/monitoring/cluster-monitoring.txt
@@ -85,7 +85,8 @@ The following sections show sample output for each type of SDAM event.
 serverDescriptionChanged
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
-.. code-block:: js
+.. code-block:: json
+   :copyable: false
 
   ServerDescriptionChangedEvent {
      topologyId: 0,
@@ -166,7 +167,8 @@ one of the following possible values:
 serverHeartbeatStarted
 ^^^^^^^^^^^^^^^^^^^^^^
 
-.. code-block:: js
+.. code-block:: json
+   :copyable: false
 
    ServerHeartbeatStartedEvent {
      connectionId: 'localhost:27017'
@@ -175,7 +177,8 @@ serverHeartbeatStarted
 serverHeartbeatSucceeded
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
-.. code-block:: js
+.. code-block:: json
+   :copyable: false
 
    ServerHeartbeatSucceededEvent {
      duration: 1.939997,
@@ -212,7 +215,8 @@ serverHeartbeatSucceeded
 serverHeartbeatFailed
 ^^^^^^^^^^^^^^^^^^^^^
 
-.. code-block:: js
+.. code-block:: json
+   :copyable: false
 
    ServerHeartbeatFailed {
      duration: 20,
@@ -223,7 +227,8 @@ serverHeartbeatFailed
 serverOpening
 ^^^^^^^^^^^^^
 
-.. code-block:: js
+.. code-block:: json
+   :copyable: false
 
    ServerOpeningEvent {
      topologyId: 0,
@@ -233,7 +238,8 @@ serverOpening
 serverClosed
 ^^^^^^^^^^^^
 
-.. code-block:: js
+.. code-block:: json
+   :copyable: false
 
    ServerClosedEvent {
      topologyId: 0,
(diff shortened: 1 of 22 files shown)


source/fundamentals/monitoring/cluster-monitoring.txt
```diff
@@ -85,7 +85,8 @@ The following sections show sample output for each type of SDAM event.
 serverDescriptionChanged
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
-.. code-block:: js
+.. code-block:: json
```
ccho-mongodb, line 88: I don't think these are JSON, but rather Node objects being printed out.
One way you can tell is that "ServerDescriptionChangedEvent" isn't a key or part of a value. Everything in JSON needs to be a key or part of a value and paired with a ":". 

The pygments lexer (what parses the code block and guides the code that adds the highlighting) doesn't recognize this as JSON and therefore does not have any highlighting. The current version in which Javascript is identified as the code block language has highlighting.

Could you change all instances in which the output represents some output that isn't pure JSON back to Javascript so that it maintains the highlighting?

Note that as long as each preceding line has valid JSON, syntax highlighting continues to work; i.e. when you have JSON objects on separate lines not joined by commas, it still renders correctly.

source/usage-examples/insertOne.txt
```diff
@@ -25,11 +25,8 @@ can also pass a callback method as an optional third parameter.
 For more information on this method, see the
 :node-api-4.0:`insertOne() API documentation </classes/collection.html#insertone>`.
 
-The ``insertedId`` field of this object is the ``_id`` of the
-inserted document. If the operation fails to create a document, the
-``insertedCount`` field of this object has a value of ``0``. If the
-operation creates a document, the ``insertedCount`` field of this object
-has a value of ``1``.
+If the operation creates a document, the ``insertedId`` field of this
+object is the ``_id`` of the inserted document.
```
ccho-mongodb, line 29: I think there's a grammatical issue here. The operation appends _id value of the inserted document to the object as the insertedId field, right?
Currently, the effect of this sentence doesn't seem connected to the cause. Here's something similar in structure to this sentence:
"If Chris goes to the grocery store, the bag is full of chips and salsa."
Instead, you might say:
"If Chris goes to the grocery store, he purchases a bag full of chips and salsa."

ccho-mongodb (changes requested):
Mostly looks good. One grammar issue and a code block syntax highlighting issue.

source/fundamentals/monitoring/cluster-monitoring.txt
```diff
@@ -85,7 +85,7 @@ The following sections show sample output for each type of SDAM event.
 serverDescriptionChanged
 ^^^^^^^^^^^^^^^^^^^^^^^^
 
-.. code-block:: json
+.. code-block:: js
```
ccho-mongodb, line 88: A couple items here:
- I think "js" might be equivalent to "javascript" (need to check pygments reference docs), but I noticed that the old version of the docs has proper syntax highlighting and this one does not:
https://docs.mongodb.com/drivers/node/current/fundamentals/monitoring/cluster-monitoring/#serverdescriptionchanged

I think part of the goal should not be to remove syntax highlighting where it existed before unless it was incorrect. Could you check the current documentation and compare them with all the instances you've changed to make sure this goal is met?

source/usage-examples/insertOne.txt
```diff
@@ -25,8 +25,9 @@ can also pass a callback method as an optional third parameter.
 For more information on this method, see the
 :node-api-4.0:`insertOne() API documentation </classes/collection.html#insertone>`.
 
-If the operation creates a document, the ``insertedId`` field of this
-object is the ``_id`` of the inserted document. 
+If the operation creates a document, the object appends the ``_id``
+of the inserted document and references it as the objects ``insertedId``
+field.
```
ccho-mongodb, line 30: A few items here:
- The object does not do the appending. The operation, when successful, appends the field and value to the object.
- When using possessives, make sure to use apostrophe. E.g. "object's".
- Make sure to clearly identify to the reader that "creates a document" is the successful case and that it's referred to by the specific article ("the") rather than the unspecific one ("a"). Also, we might consider using "inserts" rather than "creates". E.g. "If the operation successfully inserts the document".

ccho-mongodb (changes requested):
One additional item I was thinking about this morning is how you might cherry-pick this. I think the insertOne changes do not apply to 3.6, only 4.x/master. I think you'll need to do a separate commit to revert the change on the 3.6 branch if you cherry-pick. Just something to keep in mind.

PR #247, opened 2021-08-30
(DOCSP-17935) Fundamentals > TypeScript

biniona-mongodb:
## Pull Request Info

Node TypeScript Page

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-17935

### Snooty build log:
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?jobId=612c36b37dcf012930f34ccf

https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-17935-Typescript-Page/fundamentals/typescript/

### Self-Review Checklist


biniona-mongodb:
There is currently an unrelated issue with the Node Driver Docs causing API links to not work. This will be resolved separately, and has to do with a case change in API link URLs.

source/usage-examples/bulkWrite.txt
```diff
@@ -87,6 +87,12 @@ to ``bulkWrite()`` includes examples of ``insertOne``, ``updateMany``, and
          :language: typescript
          :linenos:
 
+      .. important:: Dot Notation Loses Type Safety
+         
+         You lose type-safety for values when you use dot notation in keys. For
+         more information, see our guide on
```
jason-price-mongodb, line 93: Suggestion: "For more information, see :ref:`TypeScript ...`"

source/fundamentals/typescript.txt
```diff
@@ -0,0 +1,97 @@
...
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecols
+
+Overview
+--------
+
+In this guide, you can learn about the **TypeScript** specific features and limitiations
```
jason-price-mongodb, line 16: Suggestion: "In this guide, you learn learn about the **TypeScript** features and ..."

source/fundamentals/typescript.txt
```diff
@@ -0,0 +1,97 @@
...
+of the MongoDB Node.js driver. TypeScript is a strongly typed programming
+language that compiles to JavaScript.
+
+All TypeScript features of the driver are optional, and all valid JavaScript
```
jason-price-mongodb, line 20: Suggestion: "... optional. All valid ..."

source/fundamentals/typescript.txt
```diff
@@ -0,0 +1,97 @@
...
+code written with the driver is also valid TypeScript code. 
+
+For more information on TypeScript, see 
+`the official TypeScript website <https://www.typescriptlang.org/>`__.
```
jason-price-mongodb, line 24: Suggestion: "... see the `TypeScript ..."

source/fundamentals/typescript.txt
```diff
@@ -0,0 +1,97 @@
...
+All classes that accept a type paramater in the driver have the default type
+:node-api-4.0:`Document <interfaces/document.html>`.
+
+The following classes accept any type that extends the ``Document``
+interface:
+
+- :node-api-4.0:`Collection <classes/collection.html>`
+- :node-api-4.0:`ChangeStream <classes/changestream.html>`
+
+You can pass a type parameter that extends the ``Document`` interface like this:
```
jason-price-mongodb, line 39: Suggestion: "To pass a type ... interface:"

source/fundamentals/typescript.txt
```diff
@@ -0,0 +1,97 @@
...
+.. literalinclude:: /code-snippets/typescript/extend-document.ts
+   :language: typescript
+   :linenos:
+
+The following classes accept any type parameter:
+
+- :node-api-4.0:`FindCursor <classes/findcursor.html>`
+- :node-api-4.0:`AggregationCursor <classes/aggregationcursor.html>`
+
+You can find an example that shows how to specify a type for the ``FindCursor``
```
jason-price-mongodb, line 50: Suggestion: "To specify a type for the ``FindCursor`` class, see :ref:... "

source/fundamentals/typescript.txt
```diff
@@ -0,0 +1,97 @@
...
+class in our 
+:ref:`Find a Document Usage Example <node-driver-findone-usage-example-code-snippet>`.
+
+Limitations
+-----------
+
+.. _node-driver-typescript-limitations-dot-notation:
+
+The driver cannot infer the type of values with keys containing **dot
```
jason-price-mongodb, line 59: Suggestion:" The driver cannot determine ..."

source/fundamentals/typescript.txt
```diff
@@ -0,0 +1,97 @@
...
+        :start-after: start-error
+        :end-before: end-error
+
+     This is the error:
+
+     .. code-block:: text
+
+        Type 'string' is not assignable to type 'number'.
+
+You should still use dot notation to access nested fields in query and update documents
```
jason-price-mongodb, line 93: Suggestion: "You use dot ... TypeScript. Ensure your nested field values have your intended type."

source/fundamentals/typescript.txt
biniona-mongodb: I agree that in most cases determine would be a better word, but [type inference](https://www.typescriptlang.org/docs/handbook/type-inference.html) is a specific feature of the typescript language, so I think it is better to use "infer" here as it has a precise meaning. Open to further discussion if you disagree!

source/fundamentals/typescript.txt
biniona-mongodb: We use the "you can pass... like this:" style for introducing code snippets in drivers documentation pretty frequently and it think it is good for talking about something that demonstrates the look and feel of a feature that could be used in lots of different ways. 

"To pass a type... interface" feels a little too restrictive, as the code snippet is not the be all end all way to do this, just highlights some of the syntax. Open to further discussion if you disagree!

source/fundamentals/typescript.txt
biniona-mongodb: I think the edit loses some meaning from the original sentence. I am trying to say:

- Even though you lose a feature of TypeScript which caused you to use TypeScript when you use dot notation, you should still use dot notation (I found this surprising). 
- You can't ensure through the TS language that your type is correct, so you need to manually check with your eyes and trace the code yourself. There is not a way at present to "ensure", so I think "check" is a better word here.

Tweaked the sentence to try and highlight these points: 

Despite the lack of type safety, you should still use dot notation to access nested
fields in query and update documents when you use TypeScript. You must
manually check that your nested field values have your intended type.

source/fundamentals/typescript.txt
biniona-mongodb: Similar response [as above](https://github.com/mongodb/docs-node/pull/247#discussion_r698669422), I think this edit is a little too restrictive (this isn't the be all end all it just highlights syntax). Open to further discussion if you disagree!

jason-price-mongodb (approved):

biniona-mongodb:
Hi @mongodb/dbx-node, could I have a review of this page?

source/fundamentals/typescript.txt
```diff
@@ -0,0 +1,97 @@
...
+
+For more information, see the
+`TypeScript website <https://www.typescriptlang.org/>`__.
+
+Features
+--------
+
+If you use TypeScript, you can specify a type for some classes in the driver.
+All classes that accept a type paramater in the driver have the default type
+:node-api-4.0:`Document <interfaces/document.html>`.
```
nbbeeken, line 31: It isn't that useful to link to the `Document` API docs since it doesn't display the definition. Can you inline:

```typescript
interface Document {
    [key: string]: any;
}
```

This is what Document will always be defined as, the minimum constraint it enforces is that keys must be strings and `T extends Document` enforces T must be an object (as opposed to some primitive like `number`, since all things in mongodb must be objects)

source/code-snippets/typescript/extend-document.ts
```diff
@@ -0,0 +1,7 @@
+interface YourInterface {
```
nbbeeken, line 1: Disregard if this is consistent with other docs but I think an example would work better here:

```typescript
interface Pet {
  name: string;
  age: number;
  cute: true;
}

const database = client.db("<your database>");
const collection = database.collection<Pet>("<your collection>");
```

Using sentences in the strings as types does work, but the example to a potential newcomer to typescript feels confusing. Again no worries if that's not what we're going for in this particular example.

source/fundamentals/typescript.txt
```diff
@@ -0,0 +1,97 @@
...
+        :start-after: start-error
+        :end-before: end-error
+
+     This is the error:
+
+     .. code-block:: text
+
+        Type 'string' is not assignable to type 'number'.
+
+Despite the lack of type safety, you should still use dot notation to access nested
```
nbbeeken, line 93:
```suggestion
Despite the lack of type safety, we still recommend using dot notation to access nested
```

source/fundamentals/typescript.txt
```diff
@@ -0,0 +1,97 @@
...
+:ref:`Find a Document Usage Example <node-driver-findone-usage-example-code-snippet>`.
+
+Limitations
+-----------
+
+.. _node-driver-typescript-limitations-dot-notation:
+
+The driver cannot infer the type of values with keys containing **dot
+notation**. Dot notation is a property access syntax for navigating BSON objects.
+Click on the tabs to see code snippets that highlight this behavior:
```
nbbeeken, line 61: Add here something along the lines of:
This has an additional effect of any keys that are not explicitly listed in your Schema are permitted and typed as any.

i.e.

```typescript
const c = db.collection<{ someKey: number }>('...')
c.find({ someRANDOMKey: 'whatever type I want!' })
```

nbbeeken (changes requested):

source/fundamentals/typescript.txt
biniona-mongodb: Sounds good! One thing to be aware of for saying "we recommend" in docs is that this becomes an opinion on behalf of MongoDB the company. Will make the edit, just giving context around the specific meaning of "we recommend".

[Passage from style guide](https://docs.mongodb.com/meta/style-guide/writing/write-to-the-user/#id1):

>  Use we recommend rather than it’s recommended or MongoDB recommends. Limit how often you state recommendations. This expresses an opinion on behalf of MongoDB. Make sure that the engineering team or TSE team has expressed this opinion before stating anything as a recommendation. You should be prepared to defend any recommendations you publish with a ticket number from engineering or the TSE organization.

source/code-snippets/typescript/extend-document.ts
biniona-mongodb: Your edit is not inconsistent with other docs at all. If what I currently have is something that will confuse a TypeScript reader than best not to do it. Thanks for the note!

source/fundamentals/typescript.txt
biniona-mongodb: Added an "Important" note describing this behavior!

Is this intended behavior or a limitation? I thought this was intended behavior modeling MongoDBs acceptance of any key in documents?

I added an "Important" note to the "Features" section clarifying this behavior. If you believe this should be described as a "Limitation" (something that in a perfect world should be there but due to a technical or resource constraint is not there at this time), let me know and I'll move the note!

source/fundamentals/typescript.txt
nbbeeken: Saying you "should use" seems like a recommendation without saying the word so we need to make this defensible regardless.

Maybe we can just add why we recommend it here. Using dot notation vs nested documents leads to smaller BSON payloads over the network. There may be more reasons but at least this one is a guarantee. There's no ticket number for that but its inherent to the BSON format.

Typescript folks will likely want to spell out their nested documents for the sake of type safety so I also don't want to scare people off from that because while there is overhead it may be negligible and worth the downside to maintain developer productivity.

source/fundamentals/typescript.txt
nbbeeken: Interesting point. In a perfect world where TS is tailored to representing MongoDB query syntax I suppose I don't think we would permit any key in a `Filter<Schema>` because if you want to be able to query a document with a key that isn't in `Schema` you should add it to `Schema` or redeclare the collection you are querying with a new `Schema2`. Also in a perfect world, I also think this would be configurable, user's would opt-in to strict `Filter<X>` vs something more flexible knowing the risks.

But focusing on the docs at hand, I would say this is working as designed then since we do want the flexibility of supporting any key in a query, it wasn't exactly planned that way 😅 the "any key" functionality is specifically to support dot and array notation, but it does work out to be a Feature.

source/fundamentals/typescript.txt
biniona-mongodb: Added a note on why we recommend it! Lead to a large internal DevEd team thread, and it is fairly difficult to mimic the behavior or dot notation using query expressions. You have  the ENTIRE subdocument in a query must match, making not possible to query on a single field reliably (you can't return documents with more or less fields). Nathan Leniz came up with a query that behaves like dot notation and it is hideous (but impressive that he came up with it all!) that I think demonstrates why dot notation is preferable. 

Query:

 collection.find({"$jsonSchema": { required: [ "field" ], properties: { field: { bsonType: "object", properties: { s1: { enum: ["hi"] }}}}}})

source/fundamentals/typescript.txt
biniona-mongodb: Woah thanks for the detailed backstory! Super interesting design decisions. Would be super cool to see opt in behavior for schema typing in the future.

source/fundamentals/typescript.txt
nbbeeken: That's quite the query! I did forget that the whole subdocument must match so yea dot notation is certainly a must. changes LGTM

source/code-snippets/typescript/dot-notation.ts
```diff
@@ -0,0 +1,31 @@
...
+  {},
+  {$set: { field : {nested : "A string"}}}
+);
+// end-error
+// start-no-key
+interface TestNumber {
+  myNumber : number
+}
+
+const database = client.db("<your db>");
```
dariakp, line 28: I think we should use consistent quotes (either all single or all double) for strings, I see both in this file

source/code-snippets/typescript/dot-notation.ts
```diff
@@ -0,0 +1,31 @@
+// start-no-error
+interface TestType {
+  field: {nested : number}; 
+}
+const database = client.db("<your db>");
+const collection = database.collection<TestType>("<your collection>");
+await collection.updateOne(
+  {},
+  {$set: { "field.nested" : "A string"}}
```
dariakp, line 9: it would be good to use consistent spacing for curly braces, e.g. `{ key: value }`

source/fundamentals/typescript.txt
```diff
@@ -0,0 +1,160 @@
...
+   In the MongoDB Query Language, you must match a subdocument exactly
+   when specifying subdocuments in a query. Dot notation allows you to query
+   nested fields without matching subdocuments exactly.  
+
+   To show this behavior, lets say you have a collection containing
+   only the following document:
+
+   .. code-block:: json
+
+      {field: {s1: "hi", s2: "bye"}}
```
dariakp, line 136: let's use consistent quotes and curly brace spacing in the inlined examples here

dariakp (changes requested):

source/code-snippets/typescript/dot-notation.ts
biniona-mongodb: Apologies, I forgot to run all code snippets through a formatter. Ran snippets through prettier formatter.

source/code-snippets/typescript/dot-notation.ts
biniona-mongodb: Ran through formatter!

source/fundamentals/typescript.txt
biniona-mongodb: Ran through formatter!

dariakp (approved):
LGTM!

PR #249, opened 2021-08-31
(DOCSP-18084) type script cleanup

biniona-mongodb:
## Pull Request Info

Miscellaneous clean-up tasks that arose while adding TypeScript content to the Node Docs. The tasks are as follows:

- [x] Make sure all usage example code is correct and URIs conform to the Spec or use a placeholder value.
- [x] Use a standard sample database name across examples that do not use the sample dataset. 
- [x] Use ES Module Import Style for All Examples 
- [x] Add note to "Usage Examples" page discussing imports in Usage Examples in line with [this Github Review Comment](https://github.com/mongodb/docs-node/pull/229#issuecomment-896298377)
- [x]  "Above Example" to "Preceding Example"
- [x]  Make Interface Names Singular when appropriate

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-18084

### Snooty build log:
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?jobId=612ea5f77dcf01293080f1d9

https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-18084-TypeScript-Cleanup/usage-examples/

### Self-Review Checklist

### DIFF
--- a/source/usage-examples.txt
+++ b/source/usage-examples.txt
@@ -63,14 +63,12 @@ with your instance of MongoDB:
 
 All examples use ES module imports. You 
 `must enable ES module imports <https://nodejs.org/api/esm.html#esm_enabling>`__
-by adding the following to your package.json file:
+by adding the following key-value pair to your package.json file:
 
 .. code-block:: json
 
-   // package.json
-   {
-      "type": "module"
-   }
+   "type": "module"
+   
 
 .. note:: Different Import Styles
 
--- a/source/usage-examples.txt
+++ b/source/usage-examples.txt
@@ -72,6 +72,27 @@ address or URL of your instance. Consult the
 :doc:`Connection Guide </fundamentals/connection>` for more information
 about getting connected to your MongoDB instance.
 
+.. note:: Examples Use ES Module Import Style
+
+   All usage examples use the ES module import style. Click on the tabs to see how to
+   import from the driver with the ES module and CommonJS module import styles:
+
+   .. tabs::
+
+      .. tab:: ES Module
+         :tabid: es-module
+
+         .. code-block:: javascript
+
+            import { MongoClient } from 'mongodb'
+
+      .. tab:: CommonJS Module
+         :tabid: commonjs-module
+
+         .. code-block:: javascript
+
+            const { MongoClient } = require('mongodb')
+
 Available Usage Examples
 ------------------------
 
--- a/source/usage-examples.txt
+++ b/source/usage-examples.txt
@@ -62,7 +62,7 @@ with your instance of MongoDB:
       "mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&writeConcern=majority";
 
 All examples use ES module imports. You 
-`must enable ES module imports <https://nodejs.org/api/esm.html#esm_enabling>`__
+`can enable ES module imports <https://nodejs.org/api/esm.html#esm_enabling>`__
 by adding the following key-value pair to your package.json file:
 
 .. code-block:: json
@@ -70,10 +70,14 @@ by adding the following key-value pair to your package.json file:
    "type": "module"
    
 
-.. note:: Different Import Styles
+.. note:: CommonJS
 
-   Click on the tabs to see how to import from the driver with the ES module and
-   CommonJS module import systems:
+   You can use any usage example with CommonJS ``require``. To use CommonJS ``require``, you
+   must swap out the ES module ``import`` statement for your CommonJS ``require``
+   statement.
+   
+   Click on the tabs to see the syntax for importing the driver with ES module
+   ``import`` and CommonJS ``require``:
 
    .. tabs::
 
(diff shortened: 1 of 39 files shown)


corryroot (changes requested):
Hi Alek, I've noted one [Style Guide](https://docs.mongodb.com/meta/style-guide/writing/use-pronouns-carefully/#this) issue with some of your edits. Otherwise, your changes look good.

source/usage-examples/changeStream.txt
```diff
@@ -127,8 +127,7 @@ callback process the event before exiting.
    The JavaScript and TypeScript code snippets above are identical. There are no
    TypeScript specific features of the driver relevant to this use case.
 
-If you run the example above, you should see output similar to the
-following:
+If you run the preceding example, you should see output that looks like this:
```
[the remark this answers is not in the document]
biniona-mongodb: I think the style guide "[this](https://docs.mongodb.com/meta/style-guide/writing/use-pronouns-carefully/#this)" guideline is only for beginning a sentence? 

>  Avoid beginning a sentence with the pronoun this, unless you follow this with a noun to clarify its meaning.

[Here is the initial thread ](https://github.com/mongodb/docs-java/pull/57#discussion_r579507100) where we arrived at the "looks like this" style for examples. I don't have a preference either way, but this phrasing is used a fair bit throughout the Java Driver documentation, so would like to make sure this phrasing is outside guidelines before accepting note as it could alter the default format for how we discuss output.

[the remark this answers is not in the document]
corryroot: @biniona-mongodb, the style guide does note the beginning of a sentence, but it creates the same issues at the beginning or the end of a sentence. Using just "the following" as noted in the other thread creates the same issues as well. From what I understand, it's not a great experience for someone using a screen reader and can create machine translation issues down the line. I'll provide this feedback on the other thread as well.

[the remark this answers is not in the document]
biniona-mongodb: I'm convinced enough to make the edits! I like your proposed phrasing. Hopefully we will all arrive at a consensus on what the best way to say "your output is the next thing on this page". I agree with you at present  that "you should see the following output" seems to be stronger than "looks like this" though.

source/usage-examples.txt
```diff
@@ -61,6 +61,38 @@ with your instance of MongoDB:
    const uri =
       "mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&writeConcern=majority";
 
+All examples use ES module imports. You 
+`must enable ES module imports <https://nodejs.org/api/esm.html#esm_enabling>`__
+by adding the following to your package.json file:
```
corryroot, line 66:
```suggestion
by adding the following code to your package.json file:
```

corryroot (approved):
LGTM with one nit!

source/usage-examples.txt
biniona-mongodb: Interesting comment! I'm not sure if the contents of a json file can reasonably be called "code" (I don't think it is executable if that is a good definition of code?), but you are right that it is ambiguous what is being referred to. Adding "you must add the following key-value pair to your..."

source/usage-examples.txt
```diff
@@ -61,6 +61,36 @@ with your instance of MongoDB:
    const uri =
       "mongodb+srv://<user>:<password>@<cluster-url>?retryWrites=true&writeConcern=majority";
 
+All examples use ES module imports. You 
+`must enable ES module imports <https://nodejs.org/api/esm.html#esm_enabling>`__
+by adding the following key-value pair to your package.json file:
```
nbbeeken, line 66: Maybe just be more clear about what the user can do:


> All examples use ES module imports, however any example can be used with a commonJS require. Copy the import statement below that matches your environment. If you want to use ES modules in Node.js see the documentation here: [LINK to Node.js docs].


The `"type": "module"` is only one way, and the other would be to use the `.mjs` extension, since there are multiple ways and other implications that come with adopting ES modules, we should leave it to the Node.js docs entirely to cover.

nbbeeken (changes requested):
Examples all look good, but a note about the import usage docs:

source/usage-examples.txt
biniona-mongodb: I agree that the current language is too restrictive, as there are multiple ways, however I think the docs should give enough guidance to get the examples running without readers having to leave the page.

I think if we present steps to make the example work we lift the burden of making readers have to become familiar with the node.js docs or consider whether to use `.mjs` extensions or update package.json (both work, so I think if we just pick one I think it makes it easier for them).

How is "You can enable ES modules..."?

I think in the docs we try to put alternative paths in notes so that readers have a clear idea of what they need to do to get to a base working state (even if both choices give the same result, it makes it harder for readers to get to a base working state if they have to make a choice). 

How is updating the note to communicate the information above? Update: 


>    You can use any usage example with CommonJS require. To use CommonJS require, you
>    must swap out the ES module import statement for your CommonJS require
>    statement.
>    
>    Click on the tabs to see the syntax for importing the driver with ES module
>    imports and CommonJS requires:

nbbeeken: I agree with this approach, thanks for taking the time to rephrase it!

nbbeeken (approved):
LGTM

PR #228, opened 2021-08-06
Docsp-17936 Watch For Changes Typescript

biniona-mongodb:
## Pull Request Info
Add typescript example for change streams

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-15236

### Docs staging link (requires sign-in on MongoDB Corp SSO):
https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-17936-Watch-For-Changes/

### Self-Review Checklist

- [ ] Is this free of any warnings or errors in the RST?
- [ ] Did you run a spell-check?
- [ ] Did you run a grammar-check?
- [ ] Does it render on staging correctly?
- [ ] Are all the links working?
- [ ] Are the staging links in the PR description updated?

PR #256, opened 2021-09-09
DOCSP-18158 updated change stream example

kyuan-mongodb:
## Pull Request Info
Updated the watch for changes usage example to be more natural -- I spoke to Daria and she provided me a snippet to replace the current one with.

This will be cherry-picked across all the pages since its the same page for all.

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-18158

### Snooty build log:
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?jobId=61410b4dc608bec57ff3c51a

https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-18158-WatchUsageExampleUpdate/usage-examples/changeStream/

### Self-Review Checklist

### DIFF
--- a/source/code-snippets/usage-examples/changeStream.js
+++ b/source/code-snippets/usage-examples/changeStream.js
@@ -14,11 +14,11 @@ let changeStream;
 async function run() {
   try {
     await client.connect();
-    const database = client.db("sample_mflix");
-    const movies = database.collection("movies");
+    const database = client.db("insertDB");
+    const collection = database.collection("haikus");
 
-    // open a Change Stream on the "movies" collection
-    changeStream = movies.watch();
+    // open a Change Stream on the "haikus" collection
+    changeStream = collection.watch();
 
     // set up a listener when change events are emitted
     changeStream.on("change", next => {
@@ -28,13 +28,16 @@ async function run() {
 
     await simulateAsyncPause();
 
-    await movies.insertOne({
-      test: "sample movie document"
+    await collection.insertOne({
+      title: "Record of a Shriveled Datum",
+      content: "No bytes, no problem. Just insert a document, in MongoDB",
     });
 
     await simulateAsyncPause();
 
-    await changeStream.close(() => console.log("closed the change stream"));
+    await changeStream.close();
+    
+    console.log("closed the change stream");
   } finally {
     await client.close();
   }
--- a/source/code-snippets/usage-examples/changeStream.js
+++ b/source/code-snippets/usage-examples/changeStream.js
@@ -14,27 +14,27 @@ let changeStream;
 async function run() {
   try {
     await client.connect();
-    const database = client.db('sample_mflix');
-    const movies = database.collection('movies');
+    const database = client.db("sample_mflix");
+    const movies = database.collection("movies");
 
     // open a Change Stream on the "movies" collection
     changeStream = movies.watch();
 
     // set up a listener when change events are emitted
-    changeStream.on('change', next => {
+    changeStream.on("change", next => {
       // process any change event
-      console.log('received a change to the collection: \t', next);
+      console.log("received a change to the collection: \t", next);
     });
 
     await simulateAsyncPause();
 
     await movies.insertOne({
-      test: 'sample movie document'
+      test: "sample movie document"
     });
 
     await simulateAsyncPause();
 
-    await changeStream.close(() => console.log('closed the change stream'));
+    await changeStream.close(() => console.log("closed the change stream"));
   } finally {
     await client.close();
   }
--- a/source/code-snippets/usage-examples/changeStream.js
+++ b/source/code-snippets/usage-examples/changeStream.js
@@ -7,7 +7,7 @@ const client = new MongoClient(uri);
 
 const simulateAsyncPause = () =>
   new Promise(resolve => {
-    setTimeout(() => resolve(), 100);
+    setTimeout(() => resolve(), 1000);
   });
 
 let changeStream;
--- a/source/code-snippets/usage-examples/changeStream.js
+++ b/source/code-snippets/usage-examples/changeStream.js
@@ -5,37 +5,30 @@ const uri = "<connection string uri>";
 
 const client = new MongoClient(uri);
 
+const simulateAsyncPause = () =>
+  new Promise(resolve => {
+    setTimeout(() => resolve(), 20000);
+  });
+
 let changeStream;
 async function run() {
   try {
     await client.connect();
-    const database = client.db("sample_mflix");
-    const movies = database.collection("movies");
+    const database = client.db('insert_db');
+    const movies = database.collection('haiku');
 
-    // open a Change Stream on the "movies" collection
+    // open a Change Stream on the "haiku" collection
     changeStream = movies.watch();
 
     // set up a listener when change events are emitted
-    changeStream.on("change", next => {
+    changeStream.on('change', next => {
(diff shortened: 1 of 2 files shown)


source/code-snippets/usage-examples/changeStream.js
```diff
@@ -21,21 +26,15 @@ async function run() {
...
-          test: "sample movie document",
-        });
-        // wait to close `changeStream` after the listener receives the event
-        setTimeout(async () => {
-          resolve(
-            await changeStream.close(() => console.log("closed the change stream"))
-          );
-        }, 1000);
-      }, 1000);
+    await simulateAsyncPause();
```
biniona-mongodb, line 29: While this code snippet is definitely more readable than what was here before (great job!), I think this doesn't substantively change the nature of the example (it still waits, inserts a document, waits, closes the change stream). This [is the comment](https://github.com/mongodb/docs-node/pull/228#discussion_r686283833) that initiated this ticket. I think this is mainly about how the example seems largely artificial in addition to how the example doesn't take advantage of helpful promise syntax.

Could this code snippet be updated to more closely match a real world use case for change streams? [I think you did this beautifully in the go docs](https://github.com/mongodb/docs-golang/blob/master/source/usage-examples/watch.txt#L30-L33).

biniona-mongodb: I think this code snippet could be used to mirror what is in the go docs: 

    // Replace the uri string with your MongoDB deployment's connection string.
    const uri = "<your uri>";
    
    var changeStream;
    async function run() {
      try {
        await client.connect();
        const database = client.db("insertDB");
        const movies = database.collection("foods");
    
        // open a Change Stream on the "movies" collection
        changeStream = movies.watch();
    
        console.log("Waiting for changes...");
        while (changeStream.hasNext()) {
          const change = await changeStream.next();
          console.log("received a change to the collection: \t", change);
        }
      } finally {
        await changeStream.close(() => console.log("closed the change stream"));
        await client.close();
      }
    }
    run().catch(console.dir);

source/usage-examples/changeStream.txt
```diff
@@ -88,17 +88,17 @@ print change events that occur on the collection.
...
+listener function can register, we use a timer, declared with
+``simulateAsyncPause`` to wait 100 milliseconds before executing the insert.
 
-We also use a second timer to wait an additional second after the insertion of
-the document to provide ample time for the change event to be received and
-the for the callback to complete its execution before closing the
-``ChangeStream`` instance using the ``close()`` method.
+We also use ``simulateAsyncPause`` after the insertion of the document
+to provide ample time for the listener function to receive the change
+event and the for the callback to complete its execution before
```
biniona-mongodb, line 100:
```suggestion
event and for the callback to complete its execution before 
```

source/usage-examples/changeStream.txt
```diff
@@ -88,17 +88,17 @@ print change events that occur on the collection.
 
 First, open the change stream on the collection and then define a callback
 on the change stream using the ``on()`` method. Once set, generate a change
-event to be emitted by performing a change to the collection.
+event to emit by performing a change to the collection.
```
biniona-mongodb, line 91:
```suggestion
event by performing a change to the collection.
```

Suggestion: Im not sure "to emit" is needed here, and it sounds odd that the change event was generated be "emitted", but rathe it was just generated.

source/usage-examples/changeStream.txt
```diff
@@ -88,17 +88,17 @@ print change events that occur on the collection.
...
 
 To generate the change event on the collection, let's use ``insertOne()``
 method to add a new document. Since the ``insertOne()`` may run before the
-listener function can register, we use a timer, declared with ``setTimeout()``
-to wait one second before executing the insert.
+listener function can register, we use a timer, declared with
+``simulateAsyncPause`` to wait 100 milliseconds before executing the insert.
```
biniona-mongodb, line 96:
```suggestion
``simulateAsyncPause`` to wait 1 second before executing the insert.
```
Suggestion: I think seconds is clearer than milliseconds (I don't intuitively know how long 100 milliseconds is). 1 second only works though if the time interval is raised.

source/code-snippets/usage-examples/changeStream.js
```diff
@@ -21,21 +26,15 @@ async function run() {
...
+
+    await movies.insertOne({
```
biniona-mongodb, line 31: Suggestion: Could the database be changed to use the `insertDB` database? I think this is used as the[ insert database on Node insert pages](https://docs.mongodb.com/drivers/node/current/usage-examples/insertOne/#example).

source/usage-examples/changeStream.txt
biniona-mongodb, line 96:
```suggestion
listener function can register, we use a timer to wait 100 milliseconds before executing the insert.
```
Suggestion: I'm not sure "declared with" is right here. I think you could say "defined as", or I think it could just be called a timer.

source/code-snippets/usage-examples/changeStream.js
```diff
@@ -5,6 +5,11 @@ const uri = "<connection string uri>";
 
 const client = new MongoClient(uri);
 
+const simulateAsyncPause = () =>
+  new Promise(resolve => {
+    setTimeout(() => resolve(), 100);
```
biniona-mongodb, line 10:
```suggestion
    setTimeout(() => resolve(), 1000);
```
Question: Why was the time lowered? Is that enough time? I think this may make this harder to talk about later on.

source/code-snippets/usage-examples/changeStream.js
```diff
@@ -5,6 +5,11 @@ const uri = "<connection string uri>";
 
 const client = new MongoClient(uri);
 
+const simulateAsyncPause = () =>
```
biniona-mongodb, line 8:
```suggestion
const timer = () =>
```

Suggestion: As this causes the function to pause, I think it could better be called "Pause" (not sure what is meant by simulate a pause)

biniona-mongodb (approved):
LGTM with suggestions

source/code-snippets/usage-examples/changeStream.js
kyuan-mongodb: I think the name is fine as is since it is stimulating  a pause -- which means making it look like a pause

source/code-snippets/usage-examples/changeStream.js
kyuan-mongodb: It can, but I don't see it adding additional value so I will leave it as is.

kyuan-mongodb:
Hi @mongodb/dbx-node , I  want to verify updates to this page are is correct.

source/code-snippets/usage-examples/changeStream.js
```diff
@@ -21,21 +26,15 @@ async function run() {
...
-      }, 1000);
+    await simulateAsyncPause();
+
+    await movies.insertOne({
+      test: "sample movie document"
     });
+
+    await simulateAsyncPause();
+
+    await changeStream.close(() => console.log("closed the change stream"));
```
nbbeeken, line 37: You can either await close or pass a callback not both. We should just await it here, and move the console.log to right after the close call.

nbbeeken (changes requested):

nbbeeken (approved):

PR #271, opened 2021-12-22
DOCSP-19819: transactions

ccho-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-19819

### Snooty build log:
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=61eb3ccabcf82347e5671ab0

https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-19396-transactions/fundamentals/transactions/

### Self-Review Checklist


source/code-snippets/transactions/txn-core.js
```diff
@@ -0,0 +1,137 @@
...
+    if (error instanceof MongoError) {
+      if (err.hasErrorLabel('UnknownTransactionCommitResult')) {
+        // add your logic to retry or handle the error
+      }
+      else if (err.hasErrorLabel('TransientTransactionError')) {
+        // add your logic to retry or handle the error
+      }
+    }
+
+    console.log('An error occured in the transaction, performing a data rollback:' + error);
```
zach-carr, line 108: [issue]
```suggestion
    console.log('An error occurred in the transaction, performing a data rollback:' + error);
```

source/code-snippets/transactions/txn-core.js
```diff
@@ -0,0 +1,137 @@
...
+    );
+    await session.commitTransaction();
+    console.log('Transaction successfully committed.');
+
+  } catch (error) {
+    if (error instanceof MongoError) {
+      if (err.hasErrorLabel('UnknownTransactionCommitResult')) {
+        // add your logic to retry or handle the error
+      }
+      else if (err.hasErrorLabel('TransientTransactionError')) {
```
zach-carr, line 103: [comment, maybe an issue]

I realize the snippet that appears on the page is only part of this larger example, but when pulling just that part of it and running it on my own, I run into:
- `MongoError` is not defined (line 99), and then
- `err` is not defined (lines 100, 103)

I solved this by removing the `instanceof MongoError` check (now just `(if error)`) and changing `err` to `error`.

source/fundamentals/transactions.txt
```diff
@@ -0,0 +1,400 @@
...
+
+Callback API Implementation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The code examples in this section show how you can use the Callback API to
+run the multi-document payment transaction in a session.
+
+The following code example shows a function you can use to start the
+session and pass the callback function that contains the operations you want
+to run as a transaction.
```
zach-carr, line 308: [suggestion] I think it would be helpful to include your entire async `run` function here. I thought it was a little confusing to see some of the internal logic of the function without seeing the extra context/ seeing the logic wrapped in a function you can run.

I think if you include the entire function, the only thing missing from a runnable example is an instantiated `MongoClient`

zach-carr (commented):
Looks really good, some minor issues and a bunch of comments you can take or leave.

source/code-snippets/transactions/txn-core.js
ccho-mongodb: Whoops, good catch! Not immediately sure why it didn't error out when I ran it.

source/fundamentals/transactions.txt
```diff
@@ -0,0 +1,400 @@
...
+Since all write operations on a single document in MongoDB are atomic, you
+may benefit most from transactions when you must make an atomic change that
+modifies multiple documents, which is called a multi-document transaction.
+Similar to write operations on a single document, multi-document transactions
+are **ACID compliant**, which means MongoDB guarantees the data involved
+in your transaction operations remains consistent, even if it encounters
+unexpected errors. Learn more from this MongoDB article on
+`ACID transactions <https://www.mongodb.com/basics/acid-transactions>`__.
+
+You can use the driver's transaction API to perform multi-document
```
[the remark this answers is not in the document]
ccho-mongodb: Thanks, I'll probably need "one of the driver's transaction APIs" for specificity and avoid any confusion that you can use them simultaneously.

source/fundamentals/transactions.txt
ccho-mongodb: Do you mean defining placeOrder() as an anonymous function? If you just meant adding the MongoClient, I don't think that's necessary. We often leave that out of fundamentals pages (e.g. https://docs.mongodb.com/drivers/node/current/fundamentals/crud/write-operations/insert/) because we are helping them learn how to create the transaction rather prescribe the structure or produce a runnable example.

If you meant the former:
What I wanted to show is that you can use a function free of any transaction-specific syntax (unfortunately need the session), but have no idea whether that's an advantage.
I do think inlining would add a lot of clutter since it's not a very short function -- nor would any multi-document transaction -- and therefore draw attention away from the withTransaction parameter list.

source/fundamentals/transactions.txt
```diff
@@ -0,0 +1,400 @@
...
+
+If the payment fails, you can use a transaction to make sure that you
+avoid exposing any partial updates that might cause data consistency issues
+for other operations that depend on that data.
+
+Sample Data
+~~~~~~~~~~~
+
+The code examples require the following sample data in the ``testdb``
+database to successfully run the multi-document payment transaction:
```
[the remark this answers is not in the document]
ccho-mongodb: I wanted to use bullet points, but I don't think you can put code-blocks within them. Do you think sub-headers would  accomplish the same goal?

source/fundamentals/transactions.txt
```diff
@@ -0,0 +1,400 @@
...
+
+.. code-block:: javascript
+
+   const transactionOptions = {
+     readPreference: 'primary',
+     readConcern: { level: 'local' },
+     writeConcern: { w: 'majority' },
+     maxCommitTimeMS: 1000
+   };
+   await session.withTransaction(<callback function>, transactionOptions);
```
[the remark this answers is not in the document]
ccho-mongodb: I went through several iterations of this on my own, but not sure this version is better. I'll try to find other ways we use placeholders for functions across all the docs.

source/fundamentals/transactions.txt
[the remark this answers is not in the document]
zach-carr: Ah gotcha. I think subheaders might be excessive - the section is clear enough, I was just thinking bullets could make it a little more intuitive. 

No change necessary, one last idea - how about a bullet point, a colon at the end, and the code examples outside of the bullet where they are now? Like:

- A document that represents a customer and associated orders in a collection called `customers`:

`code example`

source/fundamentals/transactions.txt
zach-carr: No, I think placeOrder() is good as-is, in its own code block below. I don't mean adding the MongoClient either, I agree that's not necessary.

I'm suggesting wrapping the snippet that's there in an async function, and maybe including client connection and disconnection. In your code, it's called `run()` and includes consts `cart` and `payment`, as well as the methods to connect and disconnect from the client (you definitely don't need `cart` and `payment` since those are established in the sample data intro. I think connecting/disconnecting could be helpful, but not necessary).

Initially my thinking was to include as much working code as possible to get a user to a working example quickly. Now that I'm reconsidering, I see more value in keeping it as concise as possible to illustrate the topic. Leaving it to you - my thought is that simply wrapping the existing snippet in an async function will add a bit of clarity.

source/fundamentals/transactions.txt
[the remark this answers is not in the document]
ccho-mongodb: Decided to leave out the "transaction API" terminology as it hasn't been explained by this point on the page.

source/fundamentals/transactions.txt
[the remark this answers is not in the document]
ccho-mongodb: Changed to myCallbackFunction and added explanation after the code block.

source/fundamentals/transactions.txt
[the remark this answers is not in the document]
ccho-mongodb: Added bullet points prior and then showed the code blocks later.

source/code-snippets/transactions/txn-core.js
ccho-mongodb: Changed this around a bit to ensure the reader understands to import MongoError and to pair the check that it's a MongoError (to guarantee that there's a function on it "hasErrorLabel").

source/fundamentals/transactions.txt
ccho-mongodb: Decided not to include connection code. If we find readers request runnable code examples for fundamentals pages, we can address creating a common format then.

source/fundamentals/transactions.txt
```diff
@@ -0,0 +1,417 @@
...
+
+.. _nodejs-transaction-callback-api-example:
+
+Callback API Implementation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The code examples in this section show how you can use the Callback API to
+run the multi-document payment transaction in a session.
+
+The following code example shows a function you can use to start the
```
zach-carr, line 320: [suggestion] very minor, you can take or leave this - the following example doesn't show a function in the same sense that the other examples do, it's more like the contents of a function that the user could write.

I think this line could be reduced to something like "The following code example shows how you can start the session and pass the callback..."

zach-carr (approved):
🎉 LGTM, left one more small suggestion

ccho-mongodb:
Hi @dariakp and @nbbeeken, this page on Node.js Driver Transactions is ready for your review!

source/code-snippets/transactions/txn-callback.js
```diff
@@ -0,0 +1,129 @@
...
+    { session }
+  );
+}
+// end callback
+
+const uri = process.env.MONGODB_URI;
+const client = new MongoClient(uri, { useUnifiedTopology: true });
+
+async function run() {
+  //await client.connect();
```
dariakp, line 98: why is there commented out code?

source/fundamentals/transactions.txt
```diff
@@ -0,0 +1,417 @@
...
+
+.. code-block:: javascript
+
+   const transactionOptions = {
+     readPreference: 'primary',
+     readConcern: { level: 'local' },
+     writeConcern: { w: 'majority' },
+     maxCommitTimeMS: 1000
+   };
+   await session.withTransaction(myCallbackFunction(session), transactionOptions);
```
dariakp, line 179: I don't think this syntax and the explanation below is quite correct, since `myCallbackFunction(session)` implies that you need to pass in the session and the text below could be interpreted to mean that there are other variables that could be defined as parameters of the function. It would be more correct to just have `myCallbackFunction` in this line, and then explain in the lines below that it can use `session` as a parameter (i.e., it will have access to `session` when it is executing). It's also probably important to mention that this function *must* return a promise to work properly.

dariakp (changes requested):
Here are a few things I noticed:

source/fundamentals/transactions.txt
nbbeeken: Just to add on a bit, I think this can still be a one liner example like so: 
```suggestion
   await session.withTransaction(async (session) => { /* code here */ }, transactionOptions);
```
And the text below could reference replacing "code here" or however the final formatting ends up.

source/code-snippets/transactions/txn-callback.js
ccho-mongodb: It's not code we're including in the documentation, just for testing purposes (related to "cleanUp" and "setup" methods). I can provide more hints to the technical writers that may look at it in the future.

source/fundamentals/transactions.txt
ccho-mongodb: Ok sounds good. I'll use Neal's syntax and explain that 1) you should pass session as a parameter if you want to declare a separate function and 2) it needs to return a Promise.

ccho-mongodb:
Hi @dariakp and @nbbeeken, thanks for your feedback. I've incorporated it in the latest commits, and it's ready for re-review whenever you get the chance.

source/code-snippets/transactions/txn-callback.js
```diff
@@ -0,0 +1,147 @@
...
+    { session }
+  );
+}
+// end callback
+
+const uri = process.env.MONGODB_URI;
+const client = new MongoClient(uri, { useUnifiedTopology: true });
+
+async function run() {
+  /* Test code: uncomment block and commentto run
```
dariakp, line 98: Is. This meant to be commented out? If so, just need to fix the typo -> "commentto"

source/fundamentals/transactions.txt
dariakp: It doesn't look like this feedback has been addressed

ccho-mongodb: Sorry, I changed it in one location but not this one. I'll check more carefully in the future!

source/code-snippets/transactions/txn-callback.js
ccho-mongodb: Thanks. 
Side comment: Admittedly, this code isn't structured in a way that is easy to either demonstrate the example code nor run the example. Since we never surface it to the user, I think the structure is ok for now.

PR #273, opened 2022-01-06
(DOCSP-19330) Update source constants in Node v3.6 branch

zach-carr:
## Pull Request Info

- Fixed version constant
- Added constants from master
- Updated the API link to use new constant

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-19330

### Snooty build log:
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=61d75d7f41accb36337011e7
Snooty build logs aren't working at the moment

https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-19330/

### Self-Review Checklist

### DIFF
--- a/snooty.toml
+++ b/snooty.toml
@@ -1,6 +1,5 @@
 name = "node"
 title = "Node.js"
-
 toc_landing_pages = ["/fundamentals/authentication", "/fundamentals", "/fundamentals/crud", "/usage-examples"]
 
 [constants]
--- a/snooty.toml
+++ b/snooty.toml
@@ -4,6 +4,9 @@ title = "Node.js"
 toc_landing_pages = ["/fundamentals/authentication", "/fundamentals", "/fundamentals/crud", "/usage-examples"]
 
 [constants]
-version = 4.0
+version = 3.6
 package-name-org = "mongodb-org"
 pgp-version = "{+version+}"
+api = "https://mongodb.github.io/node-mongodb-native/{+version+}"
+mongosh = "``mongosh``"
+driver = "node"
--- a/snooty.toml
+++ b/snooty.toml
@@ -3,9 +3,7 @@ title = "Node.js"
 toc_landing_pages = ["/fundamentals/authentication", "/fundamentals", "/fundamentals/crud", "/usage-examples"]
 
 [constants]
-version = 3.6
+version = "3.6"
 package-name-org = "mongodb-org"
 pgp-version = "{+version+}"
 api = "https://mongodb.github.io/node-mongodb-native/{+version+}"
-mongosh = "``mongosh``"
-driver = "node"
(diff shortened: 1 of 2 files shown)


snooty.toml
```diff
@@ -1,9 +1,11 @@
...
 toc_landing_pages = ["/fundamentals/authentication", "/fundamentals", "/fundamentals/crud", "/usage-examples"]
 
 [constants]
-version = 4.0
+version = 3.6
 package-name-org = "mongodb-org"
 pgp-version = "{+version+}"
+api = "https://mongodb.github.io/node-mongodb-native/{+version+}"
+mongosh = "``mongosh``"
+driver = "node"
```
biniona-mongodb, line 11: Comment: As these source constants are not used, should these be removed? This is sort of related to this [PR discussion around constants](https://github.com/mongodb/docs-visual-studio-extension/pull/2#discussion_r761268859) (may fall under YAGNI).

source/index.txt
```diff
@@ -64,8 +64,8 @@ API
 ---
 
 For detailed information about classes and methods in the MongoDB
-Node.js driver, see the :node-api-3.6:`MongoDB Node.js driver API
-documentation <>`.
+Node.js driver, see the `MongoDB Node.js driver API documentation
+<{+api+}>`__ .
```
biniona-mongodb, line 68: Question: The :node-api-3.6: constant is seems to still be used a good bit throughout the docs. It seems like it could cause issues to have some links use :node-api-3.6: and some links use the constant, but I could be incorrect. Should all :node-api-3.6: links be updated? If not, would it make sense to make another ticket to perform this replacement?

I read the A/C for the ticket and it seemed sparse (did not include TOC update), so not sure if this should or should not be a part of this ticket.

biniona-mongodb (commented):
Looks good overall. One bigger question around constant replacement.

snooty.toml
zach-carr: Sounds good - removing them. Thanks for the context 👍

source/index.txt
zach-carr: Good catch, there are many more uses of :node-api-3.6: and I do think we should keep it consistent.

I tested a handful of the :node-api-3.6: links and they all work, which raises the YAGNI-esque question for me - why replace these with a new api constant?

I reverted the change on the index page back to using :node-api-3.6: - this seems like the simplest solution. I'll bring it up in standup and please let me know what you think.

snooty.toml
terakilobyte: This needs to be quoted.

biniona-mongodb (approved):

zach-carr:
Created a ticket to swap instances of the Snooty directive :node-api-3.6: with the newer api source constant:
https://jira.mongodb.org/browse/DOCSP-20263

PR #275, opened 2022-01-11
(DOCSP-20071) Update TS Limitations

biniona-mongodb:
## Pull Request Info

Remove limitation around dot notation. Add warning around recursive types.

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-20071
+
https://jira.mongodb.org/browse/DOCSP-20417

### Snooty build log:

https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=61f81a59382d7fe4ee46b528

https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-20071-Support-Dot-Notation-TS/fundamentals/typescript/

### Self-Review Checklist


source/fundamentals/typescript.txt
```diff
@@ -79,81 +80,51 @@ You can find a code snippet that shows how to specify a type for the ``FindCurso
...
 
-.. _node-driver-typescript-limitations-dot-notation:
+.. _node-driver-recursive-type:
 
-The driver cannot infer the type of values with keys containing **dot
-notation**. Dot notation is a property access syntax for navigating BSON objects.
-Click on the tabs to see code snippets that highlight this behavior:
+You cannot specify a recursive type as a type parameter for {+driver-long+}
+classes. Click the following tabs to see a recursive type and a code snippet that
+uses the recursive type and raises an error:
```
ccho-mongodb, line 92: There are too many conjunctions in this sentence. Could you split it into two? On a related note:

I think we should refrain from using tabs (and possibly expecting readers to identify them as such) to show two or more disparate items (in this case, a type declaration and code snippet that implements the declaration). Throughout our docs, we generally only use them to  show how to accomplish one task using different tools or drivers because only one of them applies to their use case.
Not only are they different items, they cannot view both at the same time (without opening another browser) which might be easier to read. Keep in mind that some readers print out the documentation, so it's a disadvantage not to be able to see some of the information. I see some examples of this disparity in MongoDB Analyzer docs as well (e.g. "Code Snippet" and "Warning" which you might consider revisiting.

source/fundamentals/typescript.txt
```diff
@@ -79,81 +80,51 @@ You can find a code snippet that shows how to specify a type for the ``FindCurso
...
+         error TS2615: Type of property 'r' circularly references itself in mapped type '{ [Key in keyof RecursiveType]: [Key, ...NestedPaths<RecursiveType[Key]>]; }'.
 
-   The syntax of the query that does not use dot notation is cumbersome and hard
-   to understand, and may not be worth the type safety obtained from
-   avoiding dot notation.
+If you must use a recursive type to specify a schema for your documents,
+use version 4.2 of the {+driver-long+}.
 
-For more information on dot notation, see :manual:`the MongoDB Manual </core/document/#dot-notation>`.
+To track the progress of mitigating this limitation in version {+version+} of
```
ccho-mongodb, line 128: This sentence needs a more clear subject. E.g.
See the NODE-3852 JIRA ticket to track the fix for this limitation.

source/fundamentals/typescript.txt
```diff
@@ -79,81 +80,51 @@ You can find a code snippet that shows how to specify a type for the ``FindCurso
...
-      :language: typescript
-      :linenos:
-      :start-after: start-doc
-      :end-before: end-doc
+         error TS2615: Type of property 'r' circularly references itself in mapped type '{ [Key in keyof RecursiveType]: [Key, ...NestedPaths<RecursiveType[Key]>]; }'.
 
-   The syntax of the query that does not use dot notation is cumbersome and hard
-   to understand, and may not be worth the type safety obtained from
-   avoiding dot notation.
+If you must use a recursive type to specify a schema for your documents,
```
ccho-mongodb, line 125: Simplify the sentence for clarity:
"If you must use a recursive type in your document schema..."

source/fundamentals/typescript.txt
```diff
@@ -44,8 +44,8 @@ For more information on object types, see the
 Extend Document
 ~~~~~~~~~~~~~~~
 
-The following classes accept any type that extends the ``Document``
-interface:
+The following classes accept any :ref:`non-recursive <node-driver-recursive-type>`
```
ccho-mongodb, line 47: How about "The following classes accept all non-recursive types that extend ..." ?
This could avoid confusion with the "any" typescript type and "all" is a stronger specifier.

ccho-mongodb: Does it make sense to link non-recursive type to the examples that show a recursive type? Instead of this link, how about adding a sentence here that introduces the recursive type example for anyone not familiar?

source/fundamentals/typescript.txt
```diff
@@ -79,81 +80,51 @@ You can find a code snippet that shows how to specify a type for the ``FindCurso
...
-
-     .. literalinclude:: /code-snippets/typescript/dot-notation.ts
-        :language: typescript
-        :linenos:
-        :start-after: start-error
-        :end-before: end-error
-
-     This is the error:
+   .. tab:: Recursive Type
+      :tabid: recursive-type
```
ccho-mongodb, line 97: Suggestion: I wonder if it could be useful to add a sentence that explains when you might use a recursive type in MongoDB and for the example to demonstrate that.

source/fundamentals/typescript.txt
```diff
@@ -79,81 +80,51 @@ You can find a code snippet that shows how to specify a type for the ``FindCurso
...
-   nested fields without matching subdocuments exactly.
+      The following code snippet uses a recursive type:
 
-   To show this behavior, lets say you have a collection containing
-   only the following document:
+      .. code-block:: typescript
 
-   .. code-block:: json
+         const database = client.db("test-db");
+         const col = database.collection<RecursiveType>("test-collection");
```
ccho-mongodb, line 116: Should there be some consistency in the code examples within the page? This placeholder is "test-collection", but the example in Extend Document is `"<your collection>"`.

snooty.toml
```diff
@@ -11,3 +11,5 @@ pgp-version = "{+version+}"
 api = "https://mongodb.github.io/node-mongodb-native/{+version+}"
 mongosh = "``mongosh``"
 driver = "node"
+driver-long = "MongoDB Node.js driver"
+driver-short = "Node.js driver"
```
ccho-mongodb, line 15: Is this PR only for a single version (no cherry-pick)? If so, I wonder if you might consider separating it out so that you can make this specific change available to all versions.

ccho-mongodb (changes requested):
A few minor issues, would like to take another look after changes!

source/fundamentals/typescript.txt
biniona-mongodb: I removed the tabs and updated the sentence. I am open to updating other content that includes tabs, but I think it might be good to have a concrete guideline accepted across docs teams to make the update against before updates are made.

I do not believe the style guide takes a stance on the usage of tabs. The only content I could find [dealing with tabs was a tutorial describing their usage](https://docs.mongodb.com/meta/tutorials/tabbed-content/). Would it make sense to add a "Tabs" entry to the style guide describing proper usage, and then file tickets according to the guidelines arrived at?

source/fundamentals/typescript.txt
biniona-mongodb: I'm not sure I agree with this suggestion. My reasons are as follows:

- You would use a recursive type for a collection that contains a recursive structure (such as a phone tree). I'm not sure describing the real world example would contribute to the message of the section (don't use a recursive type), but would increase the length (and most likely complexity) of the page.
- This page is for [Lucas level readers](https://github.com/mongodb/docs-node/pull/247) who are familiar with recursion and recursive data structures. I'm not sure a real world example will provide much value for these readers, as they likely are able to relate recursive data structures to real world examples independently.

snooty.toml
biniona-mongodb: Good point. [I've broke this into a separate PR](https://github.com/mongodb/docs-node/pull/276). Could you review it when you get the chance?

ccho-mongodb (changes requested):
Some issues I commented on might not be directly related to this PR, so feel free to create JIRA tickets if they are not trivial.

Responding to comments here since I'm unable to add to the outdated threads at the moment:

Re: https://github.com/mongodb/docs-node/pull/275#discussion_r783187177
> I do not believe the style guide takes a stance on how tabs should be used. 

I provided an explanation as to why it doesn't make sense. There isn't guidance on everything you shouldn't do in the style guide, and sometimes there are exceptions to items listed in the style guide. If you disagree with my reasoning, we can discuss as a team. I can ticket the work to update the style guide (or you can) -- let me know what you prefer to do here.

Re: https://github.com/mongodb/docs-node/pull/275#discussion_r783209044 
> I'm not sure a real world example will provide much value for these readers, as they likely are able to relate recursive data structures to real world examples independently.

I'm not sure it's a common use case and would be helpful to demonstrate what this is. Otherwise, I would link to an example.

source/fundamentals/typescript.txt
```diff
@@ -70,90 +72,51 @@ You can pass a type parameter that extends the ``Document`` interface like this:
 Any Type
 ~~~~~~~~
 
-The following classes accept any type parameter:
+The following classes accept any type parameter that is not recursive:
```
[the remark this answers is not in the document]
biniona-mongodb: I don't believe so. The Collection type [parameter must extend `Document`](https://github.com/mongodb/node-mongodb-native/blob/00be68f/src/collection.ts#L166), where as this is [not true for the FindCursor type parameter ](https://github.com/mongodb/node-mongodb-native/blob/00be68f/src/cursor/find_cursor.ts#L34).

source/fundamentals/typescript.txt
```diff
@@ -70,90 +72,51 @@ You can pass a type parameter that extends the ``Document`` interface like this:
...
-      :language: typescript
-      :linenos:
-      :start-after: start-doc
-      :end-before: end-doc
+   error TS2615: Type of property 'r' circularly references itself in mapped type '{ [Key in keyof RecursiveType]: [Key, ...NestedPaths<RecursiveType[Key]>]; }'.
 
-   The syntax of the query that does not use dot notation is cumbersome and hard
-   to understand, and may not be worth the type safety obtained from
-   avoiding dot notation.
+If you must use a recursive type as your document schema, use version 4.2 of
```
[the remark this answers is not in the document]
biniona-mongodb: Updated to "if you must apply a recursive type to your documents"

Context: The Node docs call type arguments for documents 'TSchema' as it raises a type error if your document doesn't conform to the type (behaving like a schema in systems like Kafka). This isn't explained on the page, however so it has been removed.

source/fundamentals/typescript.txt
```diff
@@ -44,12 +44,14 @@ For more information on object types, see the
 Extend Document
 ~~~~~~~~~~~~~~~
 
-The following classes accept any type that extends the ``Document``
-interface:
+The following classes accept all types that extends the ``Document`` interface
+and that are not recursive:
```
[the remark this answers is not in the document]
biniona-mongodb: I think this changes the meaning of the sentence from "all non-recursive types that both extend Document" to "all types that are either non-recursive or extend document" (which is not a correct statement).

Based off of this, I think this edit should not be taken.

source/fundamentals/typescript.txt
```diff
@@ -70,90 +72,51 @@ You can pass a type parameter that extends the ``Document`` interface like this:
...
-     :tabid: nested-objects
-
-     The following code snippet raises a type error:
-
-     .. literalinclude:: /code-snippets/typescript/dot-notation.ts
-        :language: typescript
-        :linenos:
-        :start-after: start-error
-        :end-before: end-error
+You cannot specify a recursive type as a type parameter for {+driver-long+}
```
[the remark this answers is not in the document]
biniona-mongodb: Responding by paragraph:

- listed out the classes individually to try and clarify this
- I don't believe the suggested edit is accurate. It is not "classes that extend document", but rather some classes have type parameters that extend document.
- TODO: respond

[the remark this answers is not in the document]
ccho-mongodb: > I don't believe the suggested edit is accurate. It is not "classes that extend document", but rather some classes have type parameters that extend document.

Good to know -- why don't we workshop this together so that we can come up with a description that the reader and I can more quickly understand?

source/fundamentals/typescript.txt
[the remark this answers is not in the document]
ccho-mongodb: What I meant here is regarding my comment about  "all" vs "any". https://github.com/mongodb/docs-node/pull/275#discussion_r783258884

source/fundamentals/typescript.txt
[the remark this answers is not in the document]
ccho-mongodb: To clarify, I'm not recommending you take a suggestion, but rather there's an issue with the sentence. The two distinct verb phrases are:
1) "accept all types that [extend] the Document interface"
2) "[accept ?] that are not recursive"

I wonder if "accept" has subject agreement as the subject is not Typescript but rather the classes. Maybe talking about "assigning" a type to an interface (if that happens to be correct) is more direct and idiomatic?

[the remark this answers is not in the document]
biniona-mongodb: I've updated the sentence to the following : The following classes accept all non-recursive types that extend
the ``Document`` interface:

source/code-snippets/typescript/dot-notation.ts
```diff
@@ -19,7 +19,7 @@ interface TestNumber {
   myNumber: number;
 }
 
-const database = client.db("<your db>");
+const database = client.db("<your database>");
```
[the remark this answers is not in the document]
biniona-mongodb: 1. Sounds good
2. Tested this and I thought it was more readable as 2 lines than 1.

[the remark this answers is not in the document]
ccho-mongodb: Re 2: still not sure you need the database and collections variables unless you're targeting people new to the driver (I think you indicated to me in an earlier comment this is only for Lucas).

No change needed, but just wanted to mention that there are a few different representations that are perfectly readable used throughout the docs. For example, the Node.js engineers wrote this snippet in which both the references to the database/collection and the CRUD call are chained to the MongoClient: https://docs.mongodb.com/manual/core/transactions-in-applications/#example

source/fundamentals/typescript.txt
```diff
@@ -41,17 +41,21 @@ Any object type can extend the ``Document`` interface.
...
 
-The following classes accept any type that extends the ``Document``
-interface:
+The following classes accept all non-recursive types that extend
+the ``Document`` interface:
 
 - `Collection <{+api+}/classes/Collection.html>`__
 - `ChangeStream <{+api+}/classes/ChangeStream.html>`__
 
+To view an example of a recursive type, see the :ref:`<node-driver-limitations>` section.
```
ccho-mongodb, line 53: When reading through this time, I wonder if my comment during our discussion about ordering information might apply here. You're mentioning an example of something that won't work before you show an example of a type/interface that does work.

source/fundamentals/typescript.txt
```diff
@@ -67,93 +71,50 @@ You can pass a type parameter that extends the ``Document`` interface like this:
       :start-after: start-no-key
       :end-before: end-no-key
 
-Any Type
-~~~~~~~~
+Parameters of Any Type
+~~~~~~~~~~~~~~~~~~~~~~
 
-The following classes accept any type parameter:
+The following classes accept all type parameters that are not recursive:
```
ccho-mongodb, line 77: Would this be more uniform and therefore easier to read sequentially if it was the same format as the intro sentence in the previous section?
According to the style guide, we should avoid using negatives (e.g. "not") when possible.

source/fundamentals/typescript.txt
```diff
@@ -67,93 +71,50 @@ You can pass a type parameter that extends the ``Document`` interface like this:
...
-      :language: typescript
-      :linenos:
-      :start-after: start-doc
-      :end-before: end-doc
+   interface Pet {
+      pet: Pet | null;
+      name: string;
+      age: number;
+      cute: true;
+   }
```
ccho-mongodb, line 113: Do pets have pets? What purpose do the other fields serve, or can they be removed?

Naming aside, is this a more clear way to represent recursion for your example:
```
interface Pet {
  name: string;
  pet?: Pet[];
}
```
https://joshtronic.com/2020/04/19/recursive-types-and-interfaces-in-typescript/

Will leave this up to you (and external review if you seek one).

source/fundamentals/typescript.txt
```diff
@@ -67,93 +71,50 @@ You can pass a type parameter that extends the ``Document`` interface like this:
...
 
-   The syntax of the query that does not use dot notation is cumbersome and hard
-   to understand, and may not be worth the type safety obtained from
-   avoiding dot notation.
+To track the fix for this limitation, see 
+`NODE-3852 <https://jira.mongodb.org/browse/NODE-3852>`__
+in JIRA issue tracker.
```
ccho-mongodb, line 117:
```suggestion
in the JIRA issue tracker.
```

source/fundamentals/typescript.txt
```diff
@@ -67,93 +71,50 @@ You can pass a type parameter that extends the ``Document`` interface like this:
...
 
-For more information on dot notation, see :manual:`the MongoDB Manual </core/document/#dot-notation>`.
+If you must apply a recursive type to your documents, use version 4.2 of
```
ccho-mongodb, line 119: Should this be "classes" instead of "documents" to match the information presented in the earlier sections?

source/code-snippets/typescript/dot-notation.ts
```diff
@@ -19,7 +19,7 @@ interface TestNumber {
   myNumber: number;
 }
 
-const database = client.db("<your db>");
-const collection = db.collection<TestNumber>("...");
+const database = client.db("<your database>");
+const collection = db.collection<TestNumber>("<your collection>");
 collection.find({ someRandomKey: "Accepts any type!" });
```
ccho-mongodb, line 24: Oops, accidentally removed a comment I made here a few moments ago.
The comment was about whether the query relates to anything in the example. Would a document that includes a "TestNumber" contain a field called "someRandomKey"? Or maybe the example could be more relatable with field names that you might expect to find together.

ccho-mongodb (approved):
LGTM, a few minor issues.

source/fundamentals/typescript.txt
biniona-mongodb: Updated to explain that now a Pet can have a Pet.

The other fields are to match the earlier interface definition. I've removed the "cute" field from both in order to simplify this. I think having 2 fields in the earlier type definition makes sense, as it seems odd to specify an object type with 1 field (an object type must be specified to extend document).

I agree replacing `pet` with `pet?` increases clarity. I do not feel that making `pet` an array increases clarity (seems to just change the meaning of the class from a pet can have one pet to a pet can have many pets).

source/code-snippets/typescript/dot-notation.ts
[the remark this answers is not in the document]
biniona-mongodb: Thanks, I'll definitely consider this as an option going forward. I think it is more readable as two lines here (requires horizontal scroll with smaller window), and prettier (node teams formatter of choice) resizes this to 3 lines (each method gets its own line) rather than moves one method down.

I think two lines is best here, but if node team suggests otherwise I'll take the edit.

source/fundamentals/typescript.txt
biniona-mongodb: Updated to:

> The following classes accept all types that both extend
the ``Document`` interface and are recursive:

I think this is an equivalent sentence to:

> The following people sell cars that both drive well and are blue.

I have checked both of these sentences against multiple grammar checkers and cannot find any problems.

source/fundamentals/typescript.txt
biniona-mongodb: I'll change to "classes".

Note: I think both work here. I think "classes" ties better with what is earlier in the page, but "documents" is more in-line with the task the reader is performing (enforcing structure for documents).

source/code-snippets/typescript/dot-notation.ts
biniona-mongodb: Updated to have the example tie more to something in the world.

baileympearson:
Hey all - we received some feedback from users about the recursive type limitations and have a [PR](https://github.com/mongodb/node-mongodb-native/pull/3102) that removes some of the limitations around recursive types with the Node driver.  It doesn't fix the issue entirely but it does alleviate some of the previous limitations.

It might be worth waiting until this PR merges and is released before we update the docs so that we don't have to update them twice 😄

source/code-snippets/typescript/dot-notation.ts
```diff
@@ -15,11 +15,11 @@ const collection = database.collection<TestType>("<your collection>");
 await collection.updateOne({}, { $set: { field: { nested: "A string" } } });
 // end-error
 // start-no-key
-interface TestNumber {
-  myNumber: number;
+interface Furniture {
```
ccho-mongodb, line 18: Suggestion: choose an object that is a countable noun rather than an uncountable one. This can help a reader build an understanding of its contents more quickly as well as make it easier to translate. A simple countable object to understand might be:
```
interface User {
  email: string;
}
```

source/fundamentals/typescript.txt
```diff
@@ -79,81 +84,119 @@ You can find a code snippet that shows how to specify a type for the ``FindCurso
...
+.. _node-driver-limitations-mutual-recursion:
 
-   The syntax of the query that does not use dot notation is cumbersome and hard
-   to understand, and may not be worth the type safety obtained from
-   avoiding dot notation.
+Mutual Recursion
+~~~~~~~~~~~~~~~~
+
+You cannot specify a mutually recursive type as a type parameter in version
+{+version+} of the driver.
```
ccho-mongodb, line 171: Does this apply to all versions of the driver? If not, here's how I might structure this to prevent errors:

- Separate all text related to a specific version into a separate file
- Name the separate file "<identifying name>-4.3.txt" and include it
- It is then easier for anyone to identify that the text only belongs to v4.3.

This may become increasingly valuable when we need to apply changes to multiple driver versions of this page in the future.

source/fundamentals/typescript.txt
```diff
@@ -79,81 +84,119 @@ You can find a code snippet that shows how to specify a type for the ``FindCurso
...
+   }
+
+   interface Handler {
+      pet: MutuallyRecursivePet;
+      name: string;
+   }
 
-For more information on dot notation, see :manual:`the MongoDB Manual </core/document/#dot-notation>`.
+If you must apply a mutually recursive type to your classes, use version 4.2 of
+the {+driver-long+}.
```
ccho-mongodb, line 202: If I recall correctly, I think we decided as a team to only use the long version of the name if it both appears in an intro section and is the first mention of it.

ccho-mongodb: Similar comment for other instances.

source/fundamentals/typescript.txt
```diff
@@ -79,81 +84,119 @@ You can find a code snippet that shows how to specify a type for the ``FindCurso
...
+with an incorrect type using dot notation, but the TypeScript compiler
+does not raise an error:
 
-     .. code-block:: text
+.. code-block:: typescript
+   :emphasize-lines: 3
 
-        Type 'string' is not assignable to type 'number'.
+   database
+      .collection<RecursivePet>("your-collection")
```
ccho-mongodb, line 131: Could the placeholders for collection be consistent across examples?

source/fundamentals/typescript.txt
```diff
@@ -79,81 +84,119 @@ You can find a code snippet that shows how to specify a type for the ``FindCurso
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-     .. literalinclude:: /code-snippets/typescript/dot-notation.ts
-        :language: typescript
-        :linenos:
-        :start-after: start-error
-        :end-before: end-error
+The {+driver-short+} cannot provide type safety within nested instances of
+recursive types referenced through **dot notation**. Dot notation is a property
+access syntax for navigating BSON objects.
```
ccho-mongodb, line 120: Not sure "property access" is a common term, so maybe the sentence could be simplified:

```suggestion
recursive types referenced through **dot notation**. Dot notation is the syntax you can use to navigate nested BSON objects.
```

source/fundamentals/typescript.txt
```diff
@@ -79,81 +84,119 @@ You can find a code snippet that shows how to specify a type for the ``FindCurso
...
+.. _node-driver-limitations:
 
-The driver cannot infer the type of values with keys containing **dot
-notation**. Dot notation is a property access syntax for navigating BSON objects.
-Click on the tabs to see code snippets that highlight this behavior:
+Limitations For Driver Version {+version+}
+----------------------------------
 
-.. tabs::
+The following subsections describe the TypeScript specific limitations of
```
ccho-mongodb, line 95: I'm not sure we use "subsections" anywhere, just "sections". Readers aren't really able to differentiate this as it's more related to rst syntax, so there isn't much reason to include the specifier.

source/fundamentals/typescript.txt
```diff
@@ -79,81 +84,119 @@ You can find a code snippet that shows how to specify a type for the ``FindCurso
...
+version {+version+} of the {+driver-long+}. 
 
-  .. tab:: Dot Notation
-     :tabid: dot-notation
+Many limitations of the {+driver-short+} relate to **recursive types**.
```
ccho-mongodb, line 98: I think "Many limitations of the <driver>" isn't an accurate generalization. What are the other limitations of the driver as a whole? Why are most of them related to recursive types?

source/fundamentals/typescript.txt
ccho-mongodb: Applies to all instances.

source/fundamentals/typescript.txt
```diff
@@ -79,81 +84,119 @@ You can find a code snippet that shows how to specify a type for the ``FindCurso
...
-        :language: typescript
-        :linenos:
-        :start-after: start-error
-        :end-before: end-error
+The {+driver-short+} cannot provide type safety within nested instances of
+recursive types referenced through **dot notation**. Dot notation is a property
+access syntax for navigating BSON objects.
 
-     This is the error:
+For example, the following code snippet references a nested instance of the
```
ccho-mongodb, line 122: "For example" seems unnecessary and disconnected with the previous sentence (definition of dot syntax). Usually, you use "for example" to describe a concept from the previous sentence.

```suggestion
The following code snippet references a nested instance of the
```

source/fundamentals/typescript.txt
```diff
@@ -79,81 +84,119 @@ You can find a code snippet that shows how to specify a type for the ``FindCurso
...
+:ref:`RecursivePet <node-driver-limitations-recursive-pet>` interface
+with an incorrect type using dot notation, but the TypeScript compiler
+does not raise an error:
```
ccho-mongodb, line 125: I expected to find a description of why the TypeScript compiler does not raise an error, but couldn't find one. Could you include an explanation?

source/fundamentals/typescript.txt
```diff
@@ -79,81 +84,119 @@ You can find a code snippet that shows how to specify a type for the ``FindCurso
...
+      .findOne({ "pet.age": "Spot" });
 
-Despite the lack of type safety, we still recommend that you use dot notation to
-access nested fields in query and update documents when you use TypeScript. You
-must manually check that your nested field values have your intended type.
+The following code snippet references a top-level instance of the
+``RecursivePet`` interface with an incorrect type and raises a type error:
 
-.. note:: Reason To Use Dot Notation
+.. tabs::
```
ccho-mongodb, line 137: Wouldn't it be better to be able to show both code snippet and error without having to select the tab?

source/fundamentals/typescript.txt
```diff
@@ -67,10 +69,13 @@ You can pass a type parameter that extends the ``Document`` interface like this:
       :start-after: start-no-key
       :end-before: end-no-key
 
-Any Type
-~~~~~~~~
+To view an example of a mutually recursive type, which is not supported by the
+preceding classes, see the :ref:`<node-driver-limitations-mutual-recursion>` section.
```
ccho-mongodb, line 73: Since the classes aren't directly before this text, perhaps it's worth mentioning them again or linking back to it, which is how the style guide suggests using the term "preceding".

source/fundamentals/typescript.txt
```diff
@@ -79,81 +84,119 @@ You can find a code snippet that shows how to specify a type for the ``FindCurso
...
+
+You cannot specify a mutually recursive type as a type parameter in version
+{+version+} of the driver.
+
+If you specify a mutually recursive type, the TypeScript compiler raises the
+following error:
+
+.. code-block:: text
+
+   error TS2615: Type of property 'r' circularly references itself in mapped type '{ [Key in keyof MutuallyRecursive]: MutuallyRecursive[Key] extends MutuallyRecursive ? [Key] : MutuallyRecursive extends MutuallyRecursive[Key] ? [...] : MutuallyRecursive[Key] extends readonly (infer ArrayType)[] ? MutuallyRecursive extends ArrayType ? [...] : ArrayType extends MutuallyRecursive ? [...] : [...] : [...'.
```
ccho-mongodb, line 178: Could you truncate the full error without losing any important information? For example:

> error TS2615: Type of property 'r' circularly references itself in mapped type...

Or is there something in the error message that you want the reader to notice? If so, could you include an explanation?

source/fundamentals/typescript.txt
```diff
@@ -79,81 +84,119 @@ You can find a code snippet that shows how to specify a type for the ``FindCurso
...
+
+A mutually recursive type exists when two types define themselves relative
+to each other. You can update the
```
ccho-mongodb, line 181: I think this might require information about one of the types being a recursive type. I think if two types refer to each other and neither is recursive, that's not a mutually recursive type.

ccho-mongodb, line 181: Removed a comment because of my own misunderstanding.
I wonder if it could be helpful to clarify "define themselves relative to each other" as that is ambiguous. Maybe "mutual recursive types exist when two interfaces contain a property type of the other".

source/fundamentals/typescript.txt
```diff
@@ -79,81 +84,119 @@ You can find a code snippet that shows how to specify a type for the ``FindCurso
...
+To view an example of a mutually recursive type, which is not supported by the
+preceding classes, see the :ref:`<node-driver-limitations-mutual-recursion>` section.
 
-.. _node-driver-typescript-limitations-dot-notation:
+.. _node-driver-limitations:
 
-The driver cannot infer the type of values with keys containing **dot
-notation**. Dot notation is a property access syntax for navigating BSON objects.
-Click on the tabs to see code snippets that highlight this behavior:
+Limitations For Driver Version {+version+}
```
ccho-mongodb, line 92: While reading through this section, I wasn't able to find the information about what the limitations are quickly. I wonder if having a bulleted list of limitations that linked to the sections would be useful. The bulleted list should summarize what the limitations are rather than contain category names (current subtitles).

ccho-mongodb (changes requested):
Lots of new info here! Thanks for taking the time to create new sections and explain the concepts.

I'm not sure if I mentioned this before, but reading through the document, it seems to take a lot of time to fully grasp the advantages of TypeScript in the Node driver. In the Overview, it mentions that I'll learn about features and limitations and that the features are optional. Then, I'm presented with "Features" in which I'm expecting to see what I might use it for, but that's not clear in that section. I only learn about the syntax of the Document interface and that there are some (unidentified) classes that have the default type Document.

I think the page needs to explain why it might be useful. After reading the entire document, I deduced the following are benefits:

- I can specify interfaces to the Collection and ChangeStream type parameter to deserialize BSON data into my own interface instead of a Document.
- If I use interfaces, the driver errors out if it encounters the wrong type. 
- If there's no type specified in the interface that corresponds to a field, it's assigned the any type.

Similar request for the Limitations section. I think we discussed these changes could go in a separate ticket. Please address the PR comments here and create that ticket when you get the chance. Thanks!

source/code-snippets/typescript/dot-notation.ts
biniona-mongodb: Implemented the suggested change.

I'm not sure I see the benefit here of changing `Furniture` to `User` here ("pieces of furniture" and "users" are both able to be counted, and I would argue "Furniture" is more well known). If there is anything else suggesting `Furniture` was unacceptable here please let me know.

source/fundamentals/typescript.txt
```diff
@@ -79,81 +84,119 @@ You can find a code snippet that shows how to specify a type for the ``FindCurso
...
+         The last overload gave the following error.
+         Type 'string' is not assignable to type 'Condition<Pet>'.
 
-   .. literalinclude:: /code-snippets/typescript/note-on-dot-notation.ts
-      :language: typescript
-      :linenos:
-      :start-after: start-no-doc
-      :end-before: end-no-doc
+If you must have type safety within nested instances of recursive types,
+you must write your query or update without dot notation.
```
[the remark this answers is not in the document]
biniona-mongodb: This is a shortcoming of the MongoDB documentation that I'm not sure should be addressed here in the TS page.

I've [created a ticket to capture the unit of work of re-adding the "Query without Dot Notation" admonition that used to be on this page to somewhere in the MongoDB documentation](https://jira.mongodb.org/browse/DOCSP-20479) and linking this content to that section. This is a fairly lengthy admonition that should probably be tied to the server [Dot Notation page](https://docs.mongodb.com/manual/core/document/#dot-notation) rather that Node -> Fundamentals -> TypeScript.

source/fundamentals/typescript.txt
biniona-mongodb: Apologies, still trying to break my tabs habit.

source/fundamentals/typescript.txt
biniona-mongodb: Update to "Many" to "All" as the page only discusses limitations related to recursive types.

source/fundamentals/typescript.txt
biniona-mongodb: Added an admonition explaining why this is the case.

biniona-mongodb:
Hey Chris, I created [this ticket](https://jira.mongodb.org/browse/DOCSP-20482) to try and[ capture the unit of work described in this message](https://github.com/mongodb/docs-node/pull/275#pullrequestreview-857326680). Feel free to edit the ticket if you think some parts of the suggested desired changes are not being accurately captured.

source/fundamentals/typescript.txt
biniona-mongodb: Truncated the error message as suggested.

source/fundamentals/typescript.txt
ccho-mongodb: 1. "For" is a preposition which shouldn't be capitalized in titles unless it's the first word
2. The preposition associated with "Limitations" is usually "of". Let me know if there's some specific exception in which "for" is more appropriate.

source/fundamentals/typescript.txt
ccho-mongodb: Just for clarification, I used "property type", which seems consistent with TypeScript nomenclature according to their documentation. An example of a property type of a Type in your MutuallyRecursivePet interface is "string" for (name) or "number" (for age). Let me know if my terminology is incorrect and what the appropriate term is if so.

ccho-mongodb (approved):
Looks pretty good. As there's a a fair amount of new/revised content, there are some new minor issues to address.
I'm leaving an approval, but If there's new content to review or discussions required, please reach out.

source/includes/limitations/4.3.rst
```diff
@@ -0,0 +1,113 @@
...
+:manual:`Dot Notation </core/document/#dot-notation>`
+in the MongoDB manual.
+
+.. _node-driver-limitations-mutual-recursion:
+
+Mutual Recursion
+~~~~~~~~~~~~~~~~
+
+You cannot specify a mutually recursive type as a type parameter in version
+{+version+} of the driver.
```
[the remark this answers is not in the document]
biniona-mongodb: No, I do not believe you can have a singular mutually recursive type (there must be at least 2 types). Although a mutually recursive type is formed by multiple types, AFAIK you refer to a mutually recursive type by one of the types ([see this discussion of trees in wikipedia](https://en.wikipedia.org/wiki/Recursive_data_type#Mutually_recursive_data_types), a `forest` and `tree` are discussed individually though they are defined by each other).

source/includes/limitations/4.3.rst
```diff
@@ -0,0 +1,113 @@
...
+.. _node-driver-limitations-recursive-pet:
+
+.. code-block:: typescript
+   :emphasize-lines: 2
+
+   interface RecursivePet {
+      pet?: RecursivePet;
+      name: string;
+      age: number;
+   }
```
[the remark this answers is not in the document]
biniona-mongodb: I believe mutual recursion is considered[ a form of recursion](https://en.wikipedia.org/wiki/Mutual_recursion#Data_types), and thus it can be said that all limitations relate to recursive types.

That said, I agree that moving this example into the recursive types section benefits the page, as readers do not need to understand the relationship between recursion and mutual recursion in order to avoid errors in their programs.

source/fundamentals/typescript.txt
```diff
@@ -79,81 +89,13 @@ You can find a code snippet that shows how to specify a type for the ``FindCurso
...
+.. _node-driver-limitations:
 
-   The syntax of the query that does not use dot notation is cumbersome and hard
-   to understand, and may not be worth the type safety obtained from
-   avoiding dot notation.
+Limitations For Driver Version {+version+}
+----------------------------------
 
-For more information on dot notation, see :manual:`the MongoDB Manual </core/document/#dot-notation>`.
+.. include:: includes/limitations/{+version+}.rst
```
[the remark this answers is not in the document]
biniona-mongodb: I've moved the entire section into the limitations file. Agree it makes more sense to have the entire section within the file (weird to have everything but the heading there), and a "limitations file" should not be a requirement for a successful build (hopefully in the future a limitations file is no longer needed).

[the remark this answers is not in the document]
biniona-mongodb: Unfortunately, the header is placed at an `h3` level if drawn from the included file. I've added the header back to the original TypeScript file. I agree this will require the addition of a new limitations file when the driver is updated to v4.4.

source/includes/limitations/4.3.rst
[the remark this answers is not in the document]
ccho-mongodb: I think there might be some confusion. The definition "recursive type is a type that references itself" does not cover mutual recursive types, which requires a second type. A mutually recursive type is one that needs to be recursive in terms of another type.

source/code-snippets/typescript/dot-notation.ts
```diff
@@ -15,11 +15,11 @@ const collection = database.collection<TestType>("<your collection>");
...
+interface User {
+  email: string;
 }
 
-const database = client.db("<your db>");
-const collection = db.collection<TestNumber>("...");
-collection.find({ someRandomKey: "Accepts any type!" });
+const database = client.db("<your database>");
+const collection = db.collection<User>("<your collection>");
+collection.find({ age: "Accepts any type!" });
```
dariakp, line 24: Just a note that this will probably need to change soon to something like `{ $unknownOperator: "Accepts any type!" }` because we actually have a user submitted fix to make sure that we validate unknown keys that aren't $-prefixed but allow all $-prefixed keys for forward compatibility with new server operators. Right now the corresponding note is correct, but once we release a fix, it will need to be updated.

(see https://github.com/mongodb/node-mongodb-native/pull/3115)

dariakp (changes requested):
Just have one note here, I am not sure if we need to do anything differently for this PR or wait to merge it. I also have a question regarding where the examples of how to use dot notation correctly are, since the page only seems to talk about limitations for recursive types, but no examples of non-recursive dot notation use. Is there a follow up ticket to add that example?

source/code-snippets/typescript/dot-notation.ts
biniona-mongodb: Hi Daria! There is a ticket to add the "[query nested fields without dot notation](https://jira.mongodb.org/browse/DOCSP-20479)" example back to a different section of the documentation (DevED-DBX agreed that TS page seemed only tangentially related to dot notation and was not the correct place for this content).

The page that should explain to readers how to use Dot Notation [is the Update Arrays in a Document Page](https://docs.mongodb.com/drivers/java/sync/current/fundamentals/crud/write-operations/embedded-arrays/#specifying-array-elements) (linking to Java), however it seems like the corresponding Node page does not include [explicitly discuss dot notation](https://docs.mongodb.com/drivers/node/current/fundamentals/crud/write-operations/embedded-arrays/). I've created a ticket[ to add this content ](https://jira.mongodb.org/browse/DOCSP-20598).

As this content has already been through a few internal reviews, I think it is best to close out this pull request and handle [the upcoming changes](https://github.com/mongodb/node-mongodb-native/pull/3115) as part of a new task (any substantial additions will require an additional copy review which will take a decent amount of time as this PR has grown fairly large). 

Sorry for the long response! Let me know if I have failed to address any of your concerns.

biniona-mongodb: Good point, I agree an example of working dot notation type checking would benefit readers. I've added a "[dot notation](https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-20071-Support-Dot-Notation-TS/fundamentals/typescript/#dot-notation)" subsection to the Features section Let me know if this addresses your concerns!

source/fundamentals/typescript.txt
```diff
@@ -90,9 +90,56 @@ class in the
...
+
+.. code-block:: typescript
+
+   interface ClassificationPet {
+     name: string;
+     age: number;
+     classification: { genus: "Canis" | "Felis"; color: string };
+   }
+
+The following code snippet correctly raises a type error by specifying
```
dariakp, line 117:
```suggestion
The following code snippet correctly raises a type error when specifying
```

dariakp (changes requested):
Just a small nit here, otherwise looks good, thank you! Though I'd like @baileympearson or @nbbeeken to take a look just to make sure we haven't missed anything, since new docs are pretty extensive.

nbbeeken (approved):
LGTM, good example of dot notation's type safety, I appreciate us sticking with the Pet theme 😁

dariakp (approved):
LGTM - thanks for implementing all the feedback! Please wait for @nbbeeken to take a second look before merging

PR #277, opened 2022-01-24
(DOCSP-18745) Node CSFLE fundementals

zach-carr:
## Pull Request Info

**For copy review:**
In the Example section, I link out to [this University GitHub example](https://github.com/mongodb-university/csfle-guides/tree/master/nodejs) rather than creating my own. In the Java driver CSFLE docs, Ross put together a [Java tour](https://github.com/mongodb/mongo-java-driver/tree/master/driver-sync/src/examples/tour) (here are the [examples in the Java driver docs](https://docs.mongodb.com/drivers/java/sync/current/fundamentals/csfle/#examples)). Do you think the University GitHub is enough, or should I find/create examples closer to Ross' tour here?

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-18745

### Snooty build log:
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=620ab93a382d7fe4ee04f393

https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-18745/fundamentals/csfle/

### Self-Review Checklist

### DIFF
--- a/source/fundamentals/csfle.txt
+++ b/source/fundamentals/csfle.txt
@@ -96,3 +96,197 @@ key using the Node.js driver, see MongoDB University's GitHub
    MongoDB recommends using local key management only for testing 
    purposes, and using a remote key management service
    for production.
+
+The examples on this page use a local key, but you can also use integrated support for cloud-based key management
+services from AWS, Azure, and GCP. Each example program execution creates a new master key in memory and drops the
+``test.coll`` collection.
+
+.. tip::
+
+   In the examples, the in-memory master key is lost when the application finishes running. If you'd like to retain
+   and decrypt documents from previous runs, you can save the local master key to a file for reuse and remove the logic
+   to drop the collection.
+
+.. warning::
+
+   MongoDB recommends using local key management only for testing purposes, and using a remote key management service
+   for production.
+
+.. _auto-encryption-decryption-java:
+
+Automatic Encryption and Decryption
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The following example shows how to use ``mongodb-client-encryption`` to 
+configure a CSFLE-enabled client with a data key and JSON schema map.
+
+The data in sensetive fields is automatically encrypted 
+before insertion, and decrypted when calling ``findOne()``
+on the client side. Querying this collection from a client that isn't 
+configured for CSFLE will return the document in encrypted form.
+
+.. note::
+
+   Auto encryption requires MongoDB **Enterprise** or **Atlas**.
+
+.. code-block:: node
+
+   // Import "mongodb" and "mongodb-client-encryption"
+   const mongodb = require("mongodb")
+   const { ClientEncryption } = require("mongodb-client-encryption")
+   
+   // Set your database URI
+   const uri = "mongodb+srv://<user>>:<password>@cluster0.gvzx6.mongodb.net/myFirstDatabase?retryWrites=true&w=majority";
+   
+   // Create a regular client without encryption/ decryption capabilities
+   async function getRegularClient() {
+     const client = new mongodb.MongoClient(uri, {
+       useNewUrlParser: true,
+       useUnifiedTopology: true,
+     })
+     return await client.connect()
+   }
+   
+   const regularClient = getRegularClient();
+   
+   // Create a data key
+   async function getDataKey(client) {
+     const encryption = new ClientEncryption(client, {
+       keyVaultNamespace: 'encryption.__keyVault',
+       kmsProviders: null
+     })
+   
+     dataKey = await encryption.createDataKey("local")
+     return dataKey.toString("base64")
+   }
+   
+   const dataKey = getDataKey(regularClient);
+   
+   // Create a JSON schema map with the data key
+   function createJsonSchemaMap(dataKey) {
+     return {
+       "medicalRecords.patients": {
+         bsonType: "object",
+         // specify the encryptMetadata key at the root level of the JSON Schema.
+         // As a result, all encrypted fields defined in the properties field of the
+         // schema will inherit this encryption key unless specifically overwritten.
+         encryptMetadata: {
+           keyId: [new Binary(Buffer.from(dataKey, "base64"), 4)]
+         },
+         properties: {
+           insurance: {
+             bsonType: "object",
+             properties: {
+               policyNumber: {
+                 encrypt: {
+                   bsonType: "int",
+                   algorithm: "AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic"
+                 }
+               }
+             }
+           },
+           medicalRecords: {
+             encrypt: {
+               bsonType: "array",
+               algorithm: "AEAD_AES_256_CBC_HMAC_SHA_512-Random"
+             }
+           },
+           bloodType: {
+             encrypt: {
+               bsonType: "string",
+               algorithm: "AEAD_AES_256_CBC_HMAC_SHA_512-Random"
+             }
+           },
+           ssn: {
+             encrypt: {
+               bsonType: "int",
+               algorithm: "AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic"
+             }
+           }
+         }
+       }
+     }
+   }
+   
+   const schemaMap = createJsonSchemaMap(dataKey);
+   
+   // Use the data key and schema to create a CSFLE-enabled client
+   async function getCsfleEnabledClient(schemaMap) {
+     const client = new mongodb.MongoClient(uri, {
+       useNewUrlParser: true,
+       useUnifiedTopology: true,
+       monitorCommands: true,
+       autoEncryption: {
+         // The key vault collection contains the data key that the client uses to encrypt and decrypt fields.
+         keyVaultNamespace: 'encryption.__keyVault',
+         // The client expects a key management system to store and provide the application's master encryption key.
+         // For now, we will use a local master key, so they use the local KMS provider.
+         kmsProviders: null,
+         // The JSON Schema that we have defined doesn't explicitly specify the collection to which it applies.
+         // To assign the schema, they map it to the medicalRecords.patients collection namespace
+         schemaMap
+       }
+     })
+     return await client.connect()
+   }
+   
+   const csfleClient = getCsfleEnabledClient(schemaMap);
+   
+   // Create sample data
+   let exampleDocument = {
+     name: "Jon Doe",
+     ssn: 241014209,
+     bloodType: "AB+",
+     medicalRecords: [
+       {
+         weight: 180,
+         bloodPressure: "120/80"
+       }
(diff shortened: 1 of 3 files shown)


source/fundamentals/csfle.txt
```diff
@@ -0,0 +1,98 @@
...
+The ``mongodb-client-encryption`` module is the official client 
+encryption module for the MongoDB Node.js driver. It contains bindings 
+to communicate with the native library that manages the encryption.
+
+Add it to your project using ``npm``:
+
+.. code-block:: sh
+   :copyable: true
+
+   npm install mongodb-client-encryption
```
terakilobyte, line 60: I know newer versions of npm automatically save, but let's be explicit should people be using older versions. Same throughout the other `npm` steps.

```suggestion
   npm install mongodb-client-encryption --save
```

terakilobyte (changes requested):
A few small items I pointed out. In review, I think it would be good to have the same examples as are present on the equivalent Java page, linking to the repo at the end.

source/fundamentals/csfle.txt
zach-carr: Good call, added to both `npm install` blocks

zach-carr: I'll work on equivalent examples to add

zach-carr:
Staging: https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-18745/

source/fundamentals/csfle.txt
```diff
@@ -0,0 +1,171 @@
...
+
+Example
+-------
+
+The following example shows how to configure a CSFLE-enabled client 
+with a local key and a JSON schema. Values in fields specified by your 
+JSON schema are automatically encrypted before insertion, and decrypted 
+when calling ``find()`` with a CSFLE-enabled client.
+
+Considerations
```
terakilobyte, line 95: This heading obfuscates the example itself in the on this page widget.

terakilobyte (changes requested):
Overall looks good, but I think there's a lot of value in having a contained example within the page, much like our "by example" section. I also think an important piece that needs to be included is explicit encryption since we don't cover that in the companion project. Thoughts?

source/fundamentals/csfle.txt
zach-carr: Replacing considerations with notes, similar to the Java page

zach-carr:
> I think there's a lot of value in having a contained example within the page, much like our "by example" section.

Understood, I agree that a contained example is much more helpful. I'll expand this so it's runnable. 

My thinking/concern is that a runnable example will be harder to read through and understand, and end up being the same content that we link out to in the Node CSFLE github example. I was going for as minimal an example as possible that highlights the fundamental CSFLE setup. I think with a little expansion there's a good middle ground [along these Usage Example lines](https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-18745/usage-examples/command/#example).

> I also think an important piece that needs to be included is explicit encryption since we don't cover that in the companion project.

Agreed, we do need to cover explicit encryption. After talking to Chris, I was thinking I would add that via [DOCSP-20572](https://jira.mongodb.org/browse/DOCSP-20572) since it's already ticketed out there. That may take me some time as well and I don't want to hold this part of the page up (I should have mentioned that plan, we definitely do need another example).

terakilobyte (approved):
lgtm for automatic encryption

source/fundamentals/csfle.txt
```diff
@@ -0,0 +1,224 @@
...
+
+.. code-block:: sh
+   :copyable: true
+
+   npm install mongodb-client-encryption --save
+
+``mongocryptd``
+~~~~~~~~~~~~~~~
+
+``mongocryptd`` is a binary run as a daemon / process that is used for
```
nbbeeken, line 62:
```suggestion
``mongocryptd`` is launched automatically by the package, and it is used for 
```

Just clearing a bit of circular language here, a daemon is a type of process the slash doesn't make sense to me here.

source/fundamentals/csfle.txt
```diff
@@ -0,0 +1,224 @@
...
+.. code-block:: sh
+   :copyable: true
+
+   npm install mongodb-client-encryption --save
+
+``mongocryptd``
+~~~~~~~~~~~~~~~
+
+``mongocryptd`` is a binary run as a daemon / process that is used for 
+automatic encryption. ``mongodb-crypt`` communicates with
```
nbbeeken, line 63:
```suggestion
automatic encryption. ``mongocryptd`` communicates with 
```
I think this is what was intended?

source/fundamentals/csfle.txt
```diff
@@ -0,0 +1,224 @@
...
+- Save `master-key.txt <https://github.com/mongodb-university/csfle-guides/raw/master/nodejs/master-key.txt>`__
+  to the same directory as this example code.
+- Start a ``mongod`` locally on the default port 27017.
+- Start a ``mongocryptd`` locally on the default port 27020.
+
+.. code-block:: javascript
+  
+   const { MongoClient, Binary } = require("mongodb");
+   const { ClientEncryption } = require("mongodb-client-encryption");
+   const fs = require("fs");
```
nbbeeken, line 104:
```suggestion
   const fs = require("fs/promises");
```
Blocking the event loop isn't best practice, since we're using async/await everywhere here we can update this.

source/fundamentals/csfle.txt
```diff
@@ -0,0 +1,224 @@
...
+   
+   async function getRegularClient() {
+     const client = new MongoClient("mongodb://localhost:27017", { 
+       useNewUrlParser: true,
+       useUnifiedTopology: true,
+     });
```
nbbeeken, line 110:
```suggestion
     const client = new MongoClient("mongodb://localhost:27017");
```

No longer need these options in v4.x of the driver

source/fundamentals/csfle.txt
```diff
@@ -0,0 +1,224 @@
...
+       autoEncryption: {
+         keyVaultNamespace: "encryption.__keyVault",
+         kmsProviders: {
+           local: {
+             key: fs.readFileSync("./master-key.txt"),
+           },
+         },
+         schemaMap,
+       },
+     });
```
nbbeeken, line 127:
```suggestion
     const client = new MongoClient("mongodb://localhost:27017", {
       autoEncryption: {
         keyVaultNamespace: "encryption.__keyVault",
         kmsProviders: {
           local: {
             key: await fs.readFile("./master-key.txt"),
           },
         },
         schemaMap,
       },
     });
```

source/fundamentals/csfle.txt
```diff
@@ -0,0 +1,224 @@
...
+       },
+     };
+   }
+   
+   async function makeDataKey(client) {
+     const encryption = new ClientEncryption(client, {
+       keyVaultNamespace: "encryption.__keyVault",
+       kmsProviders: {
+         local: {
+           key: fs.readFileSync("./master-key.txt"),
```
nbbeeken, line 155:
```suggestion
           key: await fs.readFile("./master-key.txt"),
```

source/fundamentals/csfle.txt
```diff
@@ -0,0 +1,224 @@
...
+         encryptMetadata: {
+           keyId: [new Binary(Buffer.from(dataKey, "base64"), 4)],
+         },
+         properties: {
+           ssn: {
+             encrypt: {
+               bsonType: "int",
+               algorithm: "AEAD_AES_256_CBC_HMAC_SHA_512-Deterministic",
+             },
+           },
```
nbbeeken, line 144: Total suggestion here:
We can leave this as is, it doesn't change the usefulness of the example at all, but to my knowledge its common practice to store something like a social security number as a string, it doesn't have as much useful meaning as an int. For example, it has a certain length that is required, and you often want to check the last four digits. Neither of those operations are particularly easy with an Int but are with a string.

source/fundamentals/csfle.txt
```diff
@@ -0,0 +1,224 @@
...
+         .db("users")
+         .collection("ssns");
+   
+       const exampleDocument = {
+         name: "Jon Doe",
+         ssn: 241014209,
+       };
+   
+       await csfleClientSsnsColl.updateOne(
+         { ssn: exampleDocument["ssn"] },
```
nbbeeken, line 195:
```suggestion
         { ssn: exampleDocument.ssn },
```

You might want to run this code through some kind of [formatter](https://prettier.io/playground/) for consistency, accessing an object with a string index is unusual when everything else uses just dot `.` access.

source/fundamentals/csfle.txt
```diff
@@ -0,0 +1,224 @@
...
+   
+     } finally {
+       if (regularClient) await regularClient.close();
+       if (csfleClient) await csfleClient.close();
+     }
+   }
+   
+   let regularClient = null;
+   let csfleClient = null;
+   run(regularClient, csfleClient).catch(console.dir);
```
nbbeeken, line 224:
```suggestion
run().catch(error => {
   console.dir(error);
   process.exit(1);
});
```
These are in the top level of the file right? the indentation looked off. Also the variables here are just null when passed in, I'm not sure I see the use of keeping them?

nbbeeken (changes requested):

source/fundamentals/csfle.txt
zach-carr: Sounds good, appreciate the clarification

source/fundamentals/csfle.txt
zach-carr: Good point, thanks

source/fundamentals/csfle.txt
zach-carr: They are, but the whole example is in a code block - I think the indentation is correct.

I think you're right about the null variables - I had thought I needed them in the global scope for some reason. Took your suggestion!

source/fundamentals/csfle.txt
zach-carr: Got it, that makes a lot of sense. Still, I'm planning to leave this as-is to match the University example as closely as possible.

nbbeeken (approved):
One small inconsistency in the formatting but otherwise I think this is good to go so I'll leave the ✅  here

source/fundamentals/csfle.txt
```diff
@@ -0,0 +1,220 @@
...
+       };
+   
+       await csfleClientSsnsColl.updateOne(
+         { ssn: exampleDocument.ssn },
+         { $set: exampleDocument },
+         { upsert: true }
+       );
+   
+       const csfleFindResult = await csfleClientSsnsColl.findOne({
+         ssn: exampleDocument["ssn"],
```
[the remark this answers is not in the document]
zach-carr: Thanks for reviewing!

PR #280, opened 2022-01-26
DOCSP-20056: UTF-8 validation options

ccho-mongodb:
### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-20056

### Snooty build log:
https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=queue&jobId=61f2c8ddbcf82347e58d0b02

[What's New link](https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-20056-utf8-option/whats-new/#what-s-new-in-4.3)
[ToC](https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-20056-utf8-option/fundamentals/)
[UTF-8 Validation Page](https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-20056-utf8-option/fundamentals/utf8-validation/)

### Self-Review Checklist

### DIFF
--- a/source/fundamentals/utf8-validation.txt
+++ b/source/fundamentals/utf8-validation.txt
@@ -16,30 +16,32 @@ Overview
 --------
 
 In this guide, you can learn how to enable or disable the {+driver-short+}'s
-**UTF-8** validation feature. **UTF-8** is a character encoding specification
+**UTF-8** validation feature. UTF-8 is a character encoding specification
 that ensures compatibility and consistent presentation across most operating
-systems, applications, and language character sets. By default, the driver
-checks documents for any characters that are not encoded in a valid UTF-8
-format when it transfers data between your application and MongoDB.
+systems, applications, and language character sets.
 
-If you enable validation, the driver substitutes invalid UTF-8 characters
-with valid UTF-8 ones when transferring data between your application and
-MongoDB. The validation adds processing overhead since it needs to verify the
-data and to make substitutions when necessary.
+If you *enable* validation, the driver throws an error when it attempts to
+convert data that contains invalid UTF-8 characters. The validation adds
+processing overhead since it needs to check the data.
 
-If you disable validation, your application can avoid the processing
-overhead caused by validation, but cannot guarantee consistent presentation of
-invalid UTF-8 data.
+If you *disable* validation, your application avoids the validation processing
+overhead, but cannot guarantee consistent presentation of invalid UTF-8 data.
 
-Read the sections below to learn how to set UTF-8 validation using the
-{+driver-short+}.
+The driver enables UTF-8 validation by default. It checks documents for any
+characters that are not encoded in a valid UTF-8 format when it transfers data
+between your application. If it encounters invalid UTF-8 characters, it
+throws an error.
 
 .. note::
 
-   The {+driver-short+} automatically substitutes invalid UTF-8 characters
-   in your application before sending them to MongoDB regardless of whether
-   you enable validation. You can only retrieve invalid UTF-8 characters
-   from MongoDB.
+   The current version of the {+driver-short+} automatically substitutes
+   invalid UTF-8 characters with alternate valid UTF-8 ones prior to
+   validation when you send data to MongoDB. Therefore, the validation
+   only throws an error when the setting is enabled and the driver
+   receives invalid UTF-8 document data from MongoDB.
+
+Read the sections below to learn how to set UTF-8 validation using the
+{+driver-short+}.
 
 .. _nodejs-specify-utf-8-validation:
 
@@ -49,7 +51,10 @@ Specify the UTF-8 Validation Setting
 You can specify whether the driver should perform UTF-8 validation by
 defining the ``enableUtf8Validation`` setting in the options parameter
 when you create a client, reference a database or collection, or call a
-CRUD operation. By default, the {+driver-short+} enables UTF-8 validation.
+CRUD operation. If you omit the setting, the driver enables UTF-8 validation.
+
+See the following for code examples that demonstrate how to disable UTF-8
+validation on the client, database, collection, or CRUD operation:
 
 .. code-block:: javascript
 
--- a/source/fundamentals/utf8-validation.txt
+++ b/source/fundamentals/utf8-validation.txt
@@ -1,0 +1,116 @@
+.. _nodejs-utf-8-validation:
+
+================
+UTF-8 Validation
+================
+
+.. default-domain:: mongodb
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+In this guide, you can learn how to enable or disable the {+driver-short+}'s
+UTF-8 validation feature. By default, your driver checks documents for any
+characters that are not encoded in a valid UTF-8 format when it transfers data
+between your application and MongoDB.
+
+**UTF-8** is a character encoding specification that is compatible with
+most operating systems, applications, and language character sets. See the
+Wikipedia entry on :wikipedia:`UTF-8 <UTF-8>` for more information on this
+standard.
+
+If you enable validation, the driver substitutes invalid UTF-8 characters
+with valid UTF-8 ones when transferring data between your application and
+MongoDB. The validation adds processing overhead since it needs to verify the
+data and to make substitutions when necessary.
+
+If you disable validation, your application can avoid the processing
+overhead caused by validation, but cannot guarantee consistent presentation of
+invalid UTF-8 data.
+
+.. note::
+
+   The {+driver-short+} automatically substitutes invalid UTF-8 characters
+   in your application before sending them to MongoDB regardless of whether
+   you enable validation. You can only retrieve invalid UTF-8 characters
+   from MongoDB.
+
+.. _nodejs-specify-utf-8-validation:
+
+Specify the UTF-8 Validation Setting
+------------------------------------
+
+You can specify whether the driver should perform UTF-8 validation by
+defining the ``enableUtf8Validation`` setting in the options parameter
+when you create a client, reference a database or collection, or call a
+CRUD operation. By default, the {+driver-short+} enables UTF-8 validation.
+
+.. code-block:: javascript
+
+   // disable UTF-8 validation on the client
+   new MongoClient('<connection uri>', { enableUtf8Validation: false });
+
+   // disable UTF-8 validation on the database
+   client.db('<database name>', { enableUtf8Validation: false });
+
+   // disable UTF-8 validation on the collection
+   db.collection('<collection name>', { enableUtf8Validation: false });
+
+   // disable UTF-8 validation on a specific operation call
+   await collection.findOne({ title: 'Cam Jansen'}, { enableUtf8Validation: false });
+
+If your application reads invalid UTF-8 from MongoDB while the
+``enableUtf8Validation`` option is enabled, it throws a ``BSONError`` that
+contains the following message:
+
+.. code-block::
+
+   Invalid UTF-8 string in BSON document
+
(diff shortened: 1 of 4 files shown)


source/fundamentals/utf8-validation.txt
```diff
@@ -0,0 +1,116 @@
...
+**UTF-8** validation feature. **UTF-8** is a character encoding specification
+that ensures compatibility and consistent presentation across most operating
+systems, applications, and language character sets. By default, the driver
+checks documents for any characters that are not encoded in a valid UTF-8
+format when it transfers data between your application and MongoDB.
+
+If you enable validation, the driver substitutes invalid UTF-8 characters
+with valid UTF-8 ones when transferring data between your application and
+MongoDB. The validation adds processing overhead since it needs to verify the
+data and to make substitutions when necessary.
```
zach-carr, line 28: [comment] Minor, but this reads like an extra "to" to me since you already say "needs to verify..." 
```suggestion
data and make substitutions when necessary.
```

source/fundamentals/utf8-validation.txt
```diff
@@ -0,0 +1,116 @@
...
+
+If you disable validation, your application can avoid the processing
```
zach-carr, line 30: [comment] It sounds like if you disable validation, your application will definitely avoid the processing overhead
```suggestion
If you disable validation, your application avoids the processing
```

source/fundamentals/utf8-validation.txt
```diff
@@ -0,0 +1,116 @@
...
+
+Read the sections below to learn how to set UTF-8 validation using the
+{+driver-short+}.
+
+.. note::
+
+   The {+driver-short+} automatically substitutes invalid UTF-8 characters
+   in your application before sending them to MongoDB regardless of whether
+   you enable validation. You can only retrieve invalid UTF-8 characters
+   from MongoDB.
```
zach-carr, line 42: [question] Are invalid UTF-8 characters that are stored in the database inadvertently substituted when read/written back, even with this setting set to false? 

1. with UTF validation set to false, retrieve invalid UTF-8 characters from MongoDB as part of a string field
2. make a change to that string that does not affect the invalid character
3. update the database with the new string, still containing the invalid character

Is this what you mean by "cannot guarantee consistent presentation of invalid UTF-8 data."? I don't think anything needs to change, but for me this note immediately raises the question "what if I want to keep my existing invalid UTF characters safely stored in MongoDB?"

source/fundamentals/utf8-validation.txt
```diff
@@ -0,0 +1,116 @@
...
+
+   // disable UTF-8 validation on the collection
+   db.collection('<collection name>', { enableUtf8Validation: false });
+
+   // disable UTF-8 validation on a specific operation call
+   await collection.findOne({ title: 'Cam Jansen'}, { enableUtf8Validation: false });
+
+If your application reads invalid UTF-8 from MongoDB while the
+``enableUtf8Validation`` option is enabled, it throws a ``BSONError`` that
+contains the following message:
```
zach-carr, line 70: [comment] I thought that it would just make a substitution based on this line: "If you enable validation, the driver substitutes invalid UTF-8 characters with valid UTF-8 ones when transferring data between your application and MongoDB." 

Suggest introducing this behavior in that second paragraph that starts "If you enable validation,..."

source/fundamentals/utf8-validation.txt
```diff
@@ -0,0 +1,116 @@
...
+   :depth: 2
+   :class: singlecol
+
+Overview
+--------
+
+In this guide, you can learn how to enable or disable the {+driver-short+}'s
+**UTF-8** validation feature. **UTF-8** is a character encoding specification
+that ensures compatibility and consistent presentation across most operating
+systems, applications, and language character sets. By default, the driver
```
zach-carr, line 21: [comment] It sounds like the default is true/enabled, but that's not 100% clear to me here because the next paragraph starts with "If you enable validation, ..."

Suggest something along the lines of "By default, UTF-8 validation is enabled and the driver checks documents..."

Later on the page does say "By default, the Node.js driver enables UTF-8 validation." but I think that could be clarified up here.

zach-carr (commented):
The page looks good but I'm not 100% sure I understand:
- What happens when you read invalid UTF-8 data from MongoDB with this validation feature enabled. Does it substitute or throw an error? 
- What exactly "cannot guarantee consistent presentation of invalid UTF-8 data" means. I think it means that invalid UTF-8 data may still be substituted depending on the CRUD operations you run. 

I left some questions and small suggestions

source/fundamentals/utf8-validation.txt
ccho-mongodb: I try to use simple present rather than present perfect ("Use the Simplest Tense" from the style guide), but am fine with the change if it's more clear.

I wasn't able to think of a great way to keep it in simple tense and try to emphasize that it definitely happens. The closest version I could think of is "If you disable validation, your application does not [incur] the processing overhead of validation" (or some other verb that is less complex than "incur").

source/fundamentals/utf8-validation.txt
ccho-mongodb: > Are invalid UTF-8 characters that are stored in the database inadvertently substituted when read/written back, even with this setting set to false?

They're converted in your application by the js-bson library as mentioned here: "The Node.js driver automatically substitutes invalid UTF-8 characters in your application before sending them to MongoDB regardless of whether you enable validation." You can't access any actual invalid UTF-8 from the driver, so if you try to insert what you fetched from MongoDB, what gets written is not the same value, but rather a replacement).

Was this admonition unclear, and if so, what would you suggest?

source/fundamentals/utf8-validation.txt
ccho-mongodb: Oops, I think there's a mistake in the paragraph that starts "if you enable validation" as mentioned in my [other comment](https://github.com/mongodb/docs-node/pull/280/files#r793102498). It doesn't substitute when retrieving data with invalid UTF-8, it fails. If the setting is enabled, the error is thrown which means it doesn't complete the operation. I can include that information about the error if it wasn't clear. Whenever you throw an error in an application, you interrupt whatever operation you were attempting.

I wonder if providing a matrix/table would be helpful or confusing. Something like this:


```
                         |  Read Operations | Write Operations
-----------------------------------------------------
Validation on:  |                               |
-----------------------------------------------------
Validation off:  |                              |
```

source/fundamentals/utf8-validation.txt
ccho-mongodb: I'll include both the statement you suggested and the description. There hasn't been enough context introduced yet by this point in the overview to assume they know what it means to enable UTF-8 validation, so I think having the current info can be helpful. Let me know if that makes sense.

source/fundamentals/utf8-validation.txt
zach-carr: I'm OK with "incur" myself. Maybe not the most ubiquitous word but I know I've used it before when it fits the situation like this.

source/fundamentals/utf8-validation.txt
zach-carr: OK that makes sense. "You can only retrieve invalid UTF-8 characters from MongoDB" is very clear. 

What's less clear to me is whether you can work with invalid UTF-8 characters in your application after reading them from MongoDB. 

Substituting invalid characters "before sending them to MongoDB" makes it sound like the substitution is part of a write operation and you could work with invalid characters in your application up until you write them back to MongoDB. If that's correct I think the note is clear and nothing needs to be changed - I just wanted to be sure I understood that you might read an invalid character and then write back a substitution.

---

One way to clarify that all of this only applies to read operations might be in the **Specify the UTF-8 Validation Setting** section - it mentions calling CRUD operations (CRUD is mentioned a few more times after), should this just be Read operations? All the examples show reads. That may or may not work as a clarification; I don't know if you can supply this to create/update/delete.

source/fundamentals/utf8-validation.txt
zach-carr: Got it. The table would definitely be helpful for me!

source/fundamentals/utf8-validation.txt
zach-carr: It does make sense not to introduce "enabling validation" before providing the context - maybe you could change the next paragraph intro to something like:

"If you use the default setting or explicitly enable validation, the driver ..." 

I don't love my own suggestion, but I do think calling out that "enabled" is the default earlier on would be helpful. 

This isn't a blocking comment, it is ultimately clear on the page and a user doesn't need to know what the default is right away.

source/fundamentals/utf8-validation.txt
ccho-mongodb: > What's less clear to me is whether you can work with invalid UTF-8 characters in your application after reading them from MongoDB if you do not enable validation.

From what I read in the code, I think that you can, but it depends in what capacity. You won't be able to write that data back to MongoDB as invalid UTF-8, but you have access to it in your application. I don't know whether it's in scope to talk about potential use cases for this.


> Substituting invalid characters "before sending them to MongoDB" makes it sound like the substitution is part of a write operation and you could work with invalid characters in your application up until you write them

I've reworded this, but it is true that you can work with invalid characters in your application up until you write them unless you're already working with BSON (not sure that's an option in the Node.js driver). When the driver calls the js-bson library to serialize the data to BSON, it performs the utf8 conversion, so any invalid UTF-8 is substituted by an alternative UTF-8 value.

> One way to clarify that the validation setting only applies to read operations might be in the Specify the UTF-8 Validation Setting section - it mentions calling CRUD operations (CRUD is mentioned a few more times after), should this just be Read operations?

I don't know whether it's true that validation only applies to read operations, but you can provide it to any CRUD operation. I think there might be a non-read operation case in which `update()` returns the updated document in a callback, and that could contain invalid UTF-8 in one of the fields that wasn't updated. Maybe the DBX engineers can verify this (I haven't dug deeper into storing invalid UTF-8 in MongoDB since it seemed out of scope for this page).

source/fundamentals/utf8-validation.txt
ccho-mongodb: I'm going to hold off on this for now because I think my latest commit is more correct and clear, which would make this redundant.

source/fundamentals/utf8-validation.txt
zach-carr: Got it, thanks for explaining :+1:

source/fundamentals/utf8-validation.txt
```diff
@@ -0,0 +1,121 @@
...
+If you *enable* validation, the driver throws an error when it attempts to
+convert data that contains invalid UTF-8 characters. The validation adds
+processing overhead since it needs to check the data.
+
+If you *disable* validation, your application avoids the validation processing
+overhead, but cannot guarantee consistent presentation of invalid UTF-8 data.
+
+The driver enables UTF-8 validation by default. It checks documents for any
+characters that are not encoded in a valid UTF-8 format when it transfers data
+between your application. If it encounters invalid UTF-8 characters, it
```
zach-carr, line 32: [issue/ suggestion]

"... transfers data between your application and MongoDB", or

"... transfers data from MongoDB to your application" (read only?)

zach-carr (approved):
One last straightforward comment, everything else looks good. The intro is very clear to me now (probably helped to sleep on this, too) 👍

source/fundamentals/utf8-validation.txt
ccho-mongodb: Great catch, I missed that while rewriting it. It's bi-directional (serialization and deserialization), so between works best.

dariakp (approved):
LGTM, great explanation!

PR #281, opened 2022-01-27
Specify driver version in quick start

ccho-mongodb:
## Pull Request Info

Ensure that the quick start version corresponds to the correct version of the driver.

### Issue JIRA link:
None

### Snooty build log:
**Paste your workerpool job link here**

https://docs-mongodborg-staging.corp.mongodb.com/NNNNNNN/node/docsworker-xlarge/NNNNNNNNNNN/

### Self-Review Checklist

- [ ] Does it render on staging correctly?
- [ ] Are the staging and workerpool job links in the PR description updated?

### DIFF
--- a/source/quick-start.txt
+++ b/source/quick-start.txt
@@ -37,20 +37,25 @@ to check your version of Node and npm, as well as installation
 instructions, see `downloading and installing Node.js and npm 
 <https://docs.npmjs.com/downloading-and-installing-node-js-and-npm>`_.
 
+Select the :guilabel:`Clone the Project` tab if you want to build your
+project from a premade dependency file or the :guilabel:`Create the Project`
+tab if you want to create the project dependency file yourself.
+
 .. tabs::
 
    .. tab:: Clone the Project
       :tabid: clone the project
 
-      First, install Git using `Git's Getting Started Guide
+      First, ensure you have ``git`` installed in your development
+      environment or otherwise install Git using `Git's Getting Started Guide
       <https://git-scm.com/book/en/v2/Getting-Started-Installing-Git>`__.
 
-      Then, in your shell, clone the `js-starter repository
+      Then, from your shell, clone the `js-starter repository
       <https://github.com/mongodb-university/js-starter>`__:
 
       .. code-block:: bash
 
-         git clone https://github.com/mongodb-university/js-starter.git
+         git clone --branch {+version+} https://github.com/mongodb-university/js-starter.git
 
       Next, navigate into the repository:
 
@@ -103,7 +108,7 @@ instructions, see `downloading and installing Node.js and npm
 
       .. code-block:: bash
 
-         npm install mongodb
+         npm install mongodb@v{+version+}
 
       This command performs the following actions:
 


source/quick-start.txt
```diff
@@ -37,20 +37,25 @@ to check your version of Node and npm, as well as installation
...
+tab if you want to create the project dependency file yourself.
+
 .. tabs::
 
    .. tab:: Clone the Project
       :tabid: clone the project
 
-      First, install Git using `Git's Getting Started Guide
+      First, ensure you have ``git`` installed in your development
+      environment or otherwise install Git using `Git's Getting Started Guide
```
biniona-mongodb, line 50: Comment: Having monospace and non-monospace "Git" in the same sentence was a bit odd for me. I'm not sure it increases clarity have the initial `git` in monospace, so might be worth changing. Non-blocking either way you go.

source/quick-start.txt
```diff
@@ -37,20 +37,25 @@ to check your version of Node and npm, as well as installation
 instructions, see `downloading and installing Node.js and npm 
 <https://docs.npmjs.com/downloading-and-installing-node-js-and-npm>`_.
 
+Select the :guilabel:`Clone the Project` tab if you want to build your
+project from a premade dependency file or the :guilabel:`Create the Project`
```
biniona-mongodb, line 41: Comment: I've never heard a `package.json` referred to as a dependency file (not wrong, but just doesn't seem quite right to me). I googled around a bit this site refers to it as a "[manifest](https://nodejs.dev/learn/the-package-json-guide)" which sounds like a better descriptor to me (captures more than listing dependencies). Non-blocking either way.

```suggestion
project from a premade manifest file or the :guilabel:`Create the Project`
```

source/quick-start.txt
```diff
@@ -37,20 +37,25 @@ to check your version of Node and npm, as well as installation
...
       <https://git-scm.com/book/en/v2/Getting-Started-Installing-Git>`__.
 
-      Then, in your shell, clone the `js-starter repository
+      Then, from your shell, clone the `js-starter repository
       <https://github.com/mongodb-university/js-starter>`__:
 
       .. code-block:: bash
 
-         git clone https://github.com/mongodb-university/js-starter.git
+         git clone --branch v{+version+} https://github.com/mongodb-university/js-starter.git
```
biniona-mongodb, line 58: Praise: Awesome job versioning the starter project, really like how version-sensitive the docs are becoming.

biniona-mongodb (approved):
LGTM + comments

source/quick-start.txt
ccho-mongodb: Manifest files are a more general category, more often used to describe metadata (a description of a package).

A "dependency file" is a term used for any file that contains dependencies and is used across languages. I haven't found any other term used to classify package.json. If you know of one that is not jargon, let me know.

source/quick-start.txt
ccho-mongodb: I agree that it should be consistent across the same form when used outside of a title, so the instance within "otherwise install Git using" should be formatted similarly.
However, I was referring to git, the package name (see text-formatting in the style guide). I think I can just use the pronoun "it" in the aforementioned instance and leave the first instance as a monospaced package name.
Does that make sense?

source/quick-start.txt
ccho-mongodb: Also, from the link you posted:

> The package.json file is kind of a manifest for your project.

This is not the same as calling it a manifest file.

PR #294, opened 2022-02-23
(DOCSP-20130) TS - _id and Inserts

biniona-mongodb:
## Pull Request Info

Document behavior how the type of the `_id` field effects the behavior of insert operations. This PR together with https://github.com/mongodb/docs-node/pull/295 aim to satisfy the A/C of DOCSP-20130.

### Issue JIRA link:
https://jira.mongodb.org/browse/DOCSP-20130

### Snooty build log:

https://workerpool-boxgs.mongodbstitch.com/pages/job.html?collName=dotcom_queue&jobId=621eb0549c4c7f6bfb9f8fda

https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-20130-TS-id/fundamentals/typescript/#insert-operations-and-the-_id-field

### Self-Review Checklist


source/fundamentals/typescript.txt
```diff
@@ -142,6 +142,127 @@ section.
...
+      } 
+
+   The following sequence of insert operations inserts one document
+   with an ``_id`` of type ``number`` and one document with an 
+   ``_id`` of type ``ObjectId`` into your collection:
+
+   .. code-block:: typescript
+
+      collection.insertOne({ _id: 1, name: "Scar" });
+      collection.insertOne({ name: "Ursula" });
```
terakilobyte, line 213: question: Won't this trigger an error since the interface makes `_id` required? If not, is the data in the table on lines 154-186 correct?

source/fundamentals/typescript.txt
```diff
@@ -142,6 +142,127 @@ section.
...
+
+      const client = new MongoClient(uri, {
+        pkFactory: { createPk: () => Math.floor(Math.random() * Number.MAX_SAFE_INTEGER) },
+      });
+
+If you must specify the ``_id`` field in the type you define to represent
+documents in your collection but you do not want to specify values for the
+``_id`` field in insert operations, use the ``OptionalId`` type. The
+``OptionalId`` type accepts a type parameter as an argument and returns
+that type with an optional ``_id`` field.
```
terakilobyte, line 228: issue: Can this be reworded from an if clause? The action seems to be inserting without specifying an `_id`. Maybe something like:

> To insert documents of a type that requires an _id field (note: this seems like we should just tell people to do this instead of presenting it as an option, may want to ask the node team if they'd consider this a best practice), use the ``OptionalId`` type to wrap your document type.

terakilobyte (changes requested):
One issue, one question. Overall it looks clean.

source/fundamentals/typescript.txt
biniona-mongodb: You are correct, I missed a question mark on line 202.

source/fundamentals/typescript.txt
biniona-mongodb: I'm not sure the suggested change captures the intended meaning of the existing text. From [Github issue](https://github.com/mongodb/node-mongodb-native/pull/3100#issuecomment-1010365060):

> On the other hand, on inserts, we have to guess the use case based on the schema, keeping in mind that the _id can be omitted entirely, provided explicitly, or provided indirectly via a pkFactory option. In order to accommodate users who provide the _id explicitly and definitely do want the type to be strictly enforced on inserts, we can’t make the _id always optional, which means we have to differentiate between optional and required schema _id in genera

I do not believe it is an accurate statement that all readers who require `_id` in their Schema also should use `OptionalId`. I believe the `OptionalId` use case is as follows:

> For your use case, if the collection type is something that is used independently in other parts of the code base as a document model, we recommend simply wrapping it in our OptionalId helper type when passing it at the collection creation point: db.collection<OptionalId<Test>>; alternatively, you could specify the collection interface with an optional or omitted _id property and use the WithId wrapper to access the type elsewhere, e.g., for testing.

Based off of this, I believe the "If statement" correctly captures the small subset of readers this applies to (you both muse say the `_id` is required (for accurately modeling documents) but you don't want to specify `_id` values on insert).

Let me know if my line of reasoning in this comment makes sense or if you still see issues in the text!

biniona-mongodb: Added "[as required](https://github.com/mongodb/docs-node/pull/294/commits/99efcd478a0bd42ce4609b383547eb34039cb5cc)" to the preceding text to try and clarify the use case.

terakilobyte: I think this needs to be asked, not inferred. The way I'm reading this, they must either specify a PKFactory, OptionalId, or provide their own _id on every insert. If I'm using the driver, I'd want to go with the "least" work option possible. This reads "Do this medium work, or do a lot of work". That seems wrong, hence why I'd like clarification from the Node team.

terakilobyte (changes requested):
With the back and forth on the "if clause", I'm now pretty sure we're missing the mark here. There's a lot of good information here, but it should be presented in a way that states what will happen, why it happens, and how to avoid it from least to most work.

source/fundamentals/typescript.txt
biniona-mongodb: Discussed over slack. Agreed on moving note lower, and making a series of follow up tickets to break the TS page into a drawer.

source/fundamentals/typescript.txt
```diff
@@ -142,6 +142,127 @@ section.
...
+         **primary key factory**.
+       | A primary key factory is a class that defines a system for creating
+         unique identifiers for documents and implements the ``PkFactory``
+         interface.
+       | The default primary key factory generates ``_id`` values of type
+         ``ObjectId``.
+
+If you must specify the ``_id`` field as required in the type you define to represent
+documents in your collection but you do not want to specify values for the
+``_id`` field in insert operations, use the ``OptionalId`` type. The
```
dariakp, line 190:
```suggestion
``_id`` field in insert operations, use the ``OptionalId`` helper type when creating the collection. The
```

source/fundamentals/typescript.txt
```diff
@@ -142,6 +142,127 @@ section.
...
+``OptionalId`` type to insert a document without specifying a value for the
+``_id`` field:
+
+.. code-block:: typescript
+
+   const database = client.db("<your database>");
+   const collection = db.collection<OptionalId<IdPet>>("<your collection>");
+   
+   collection.insertOne({
+     name: Spot,
```
dariakp, line 215:
```suggestion
     name: "Spot",
```

dariakp (approved):
Just a couple small things here for clarity (and correctness), otherwise LGTM

PR #321, opened 2022-04-25
DOCSP-21125 pkfactory and upsert operations

kyuan-mongodb:
# Pull Request Info

Added an admonition to inform users to not use pkfactory with upsert operations.

JIRA - <https://jira.mongodb.org/browse/DOCSP-21125>
Staging:
- <https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-21125/fundamentals/typescript/#insert-operations-and-the-_id-field>
- <https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-21125/fundamentals/crud/write-operations/upsert/#performing-an-upsert>
- <https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-21125/fundamentals/crud/write-operations/change-a-document/> 
- <https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-21125/usage-examples/updateOne/>
- <https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-21125/usage-examples/replaceOne/>

## Self-Review Checklist

### DIFF
--- a/source/fundamentals/crud/write-operations/change-a-document.txt
+++ b/source/fundamentals/crud/write-operations/change-a-document.txt
@@ -122,6 +122,8 @@ an :doc:`upsert </fundamentals/crud/write-operations/upsert>` which
 attempts to perform an update, but if no documents are matched, inserts
 a new document.
 
+.. include:: /includes/pkfactory-upsert.rst
+
 You cannot modify the ``_id`` field of a document nor change a field to
 a value that violates a unique index constraint. See the MongoDB server manual
 for more information on :manual:`unique indexes </core/index-unique/>`.
@@ -197,6 +199,8 @@ does not make any changes. Replace operations can be configured to perform
 an :doc:`upsert </fundamentals/crud/write-operations/upsert>` which
 attempts to perform the replacement, but if no documents are matched, it
 inserts a new document.
+
+.. include:: /includes/pkfactory-upsert.rst
 
 You cannot modify the ``_id`` field of a document nor change a field to
 a value that violates a unique index constraint. See the MongoDB server manual
--- a/source/fundamentals/crud/write-operations/upsert.txt
+++ b/source/fundamentals/crud/write-operations/upsert.txt
@@ -1,3 +1,5 @@
+.. _node-fundamentals-upsert:
+
 ======================================
 Insert or Update in a Single Operation
 ======================================
@@ -76,6 +78,8 @@ update the document, we can set ``upsert`` to ``true`` in our call to
    const options = { upsert: true };
    collection.updateOne(query, update, options);
 
+.. include:: /includes/pkfactory-upsert.rst
+   
 After you run the operation above, your collection should resemble the
 following, whether the "Deli Llama" document existed in your collection
 beforehand:
--- a/source/fundamentals/typescript.txt
+++ b/source/fundamentals/typescript.txt
@@ -190,6 +190,8 @@ documents in your collection but you do not want to specify values for the
 create your collection. The ``OptionalId`` type accepts a type parameter as an
 argument and returns that type with an optional ``_id`` field.
 
+.. include:: /includes/pkfactory-upsert.rst
+
 The following code snippet defines the ``IdPet`` interface, which
 includes a type for the ``_id`` field:
 
--- a/source/includes/pkfactory-upsert.rst
+++ b/source/includes/pkfactory-upsert.rst
@@ -1,0 +1,12 @@
+.. important:: PkFactory in Upsert Operations
+
+    When performing an :ref:`upsert operation
+    <node-fundamentals-upsert>`, the driver doesn't use any
+    ``PkFactory`` methods. Instead, the driver always creates
+    ``ObjectId`` values for the ``_id`` field of the
+    upserted documents.
+
+    If you want to use ``PkFactory`` methods, peform a :ref:`find
+    operation <node-fundamentals-retrieve-data>`, then an :ref:`update
+    <node-fundamentals-change-a-document>` or :ref:`insert
+    <node-fundamentals-insert-data>` operation.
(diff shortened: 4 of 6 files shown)


source/fundamentals/crud/write-operations/pkFactory.txt
```diff
@@ -0,0 +1,99 @@
+.. _node-pkfactory:
+
+===================
+Primary Key Factory
```
terakilobyte, line 4: issue: I like this title for the page in general, but the way it shows up in the TOC doesn't allow it to work. I think to make it fit with the rest of the TOC items, it should be changed to something more action-oriented like "Generate your own _id"

source/fundamentals/crud/write-operations/pkFactory.txt
```diff
@@ -0,0 +1,99 @@
...
+
+In this guide, you can learn about the primary key factory features
+and limitations in the MongoDB Node.js driver. 
+
+Behavior
+--------
+
+A primary key factory is a class that defines a system for creating
+unique identifiers for documents and implements  the ``pkFactory``
+interface.
```
terakilobyte, line 26: issue: This is great content but I think it's a bit out of order. If the page is about how to generate custom _id values (primary key), it might be better to go that route initially.

> In this guide, you can learn how to generate your own custom ``_id`` values using the driver's **pkFactory** option. <now define pkFactor>.

source/fundamentals/crud/write-operations/pkFactory.txt
```diff
@@ -0,0 +1,99 @@
...
+
+Assume you apply the following interface to your ``Collection``
+instance:
+
+.. code-block:: typescript
+
+   interface I {
+     _id?: number;
+     name: string;
+   }
```
terakilobyte, line 68: issue: This is typescript specific so we may want to call that and encapsulate within a tip or something. People not using typescript won't be able to use the `Interface` keyword even though they can use a `pkFactory`.

source/fundamentals/crud/write-operations/pkFactory.txt
```diff
@@ -0,0 +1,99 @@
...
+   collection.insertOne({ name: "Ursula" });
+
+To keep the ``_id`` consistent as whole numbers, apply a primary key
+factory to your ``MongoClient`` instance like this:
+
+.. code-block:: typescript
+
+   const client = new MongoClient(uri, {
+     pkFactory: { createPk: () => Math.floor(Math.random() * Number.MAX_SAFE_INTEGER) },
+   });
```
terakilobyte, line 86: issue: There are a few minor issues here.

1. I expected to see this in a top level heading for "Specify your own `pkFactory`. This is a (mostly) great example and it shouldn't be hidden since it illustrates how to specify a pkFactory, and why you would.
2. Rather than "whole number", let's use `Integer`.
3. But! Instead of using an integer here (which is a very bad _id scheme), how about [UUID](https://www.npmjs.com/package/uuid) instead, which is a nice alternative.

terakilobyte (changes requested):
A few minor issues I found, but this is great content.

source/fundamentals/crud/write-operations/pkFactory.txt
```diff
@@ -0,0 +1,124 @@
...
+
+Assume you have a collection, ``characters.villans``, that contains
+documents with a villans name. The following example inserts
+documents with an ``_id`` of type ``ObjectID`` into your collection:
+
+.. code-block:: javascript
+
+   const uuid = require('bson');
+   const { MongoClient } = require("mongodb");
+
```
terakilobyte, line 79: issue: `villain` is misspelled.
issue: Will it really insert documents with type `ObjectID`? Isn't that the default behavior? We should show users non-default behavior.
issue: Does the example need to be so long? It seems we can be matter-of-fact about it in the previous section and not show a full example.

terakilobyte (changes requested):

source/fundamentals/crud/write-operations/pkFactory.txt
kyuan-mongodb: Apologies, I didn't realize the `_id` would be saved as an `ObjectID` regardless of the type we specify in the `pkFactory` option. And yes, I agree that the example should be removed as it's only implementing the code snippet right above it.

source/fundamentals/crud/write-operations/pkFactory.txt
```diff
@@ -0,0 +1,77 @@
...
+.. code-block:: javascript
+
+   const { UUID } = require('bson');
+   ...
+   const client = new MongoClient(uri, { 
+     useUnifiedTopology: true, 
+     pkFactory: { createPk: () =>  new UUID().toBinary() }
+   });
+
+.. note:: TypeScript Data Consistency
```
baileympearson, line 56: From looking at the deleted section of `typescript.txt`, I think the actual point we'd want to discuss here is the following:

If a user is using a `pkFactory` and the user attempts to insert a document that already has an `_id` field **and** that `_id` field is of a different type than the type returned by the `pkFactory`, then they will have inconsistent data.  

This behavior is a result of how the driver applies `pkFactory` The driver only applies `pkFactory` if the document provided by the user does not have an `_id` property.  So something like the following generates two documents with different `_id` types:

```typescript 
const client = new MongoClient(uri, {
   // pkFactory returns a number
    pkFactory: { createPk: () => Math.floor(Math.random() * Number.MAX_SAFE_INTEGER) }, 
  });

....
// inserts document with a string id
collection.insertOne({ _id: "hello", name: "Scar" });  

 // pkFactory will be used to generate a number id
collection.insertOne({ name: "Ursula" });         
```

This behavior exists in both Javascript and Typescript

baileympearson (changes requested):
Looks good overall!!  Some typos and two important notes.  LMK if you have any questions!

baileympearson (approved):

PR #425, opened 2022-08-22
add ipv6 info to faq

terakilobyte:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-21804>
Staging - <https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-21804/faq/>

## Self-Review Checklist

### DIFF
--- a/source/faq.txt
+++ b/source/faq.txt
@@ -12,20 +12,28 @@ Frequently Asked Questions
    :depth: 2
    :class: singlecol
 
-How to fix MongoServerSelectionError: connect ECONNREFUSED ::1:27017 error?
----------------------------------------------------------------------------
+How to fix a "MongoServerSelectionError: connect ECONNREFUSED ::1:27017" error?
+-------------------------------------------------------------------------------
 
-Beginning with version 17, Node defaults to using ``IPV6``. You have a few
-options to resolve this:
+Beginning with version 17, Node.js defaults to using ``IPV6``. You have a few
+options to resolve this error:
 
-- You can start your ``mongod`` with :manual:`the --ipv6 flag </reference/program/mongod/#std-option-mongod.--ipv6>`, or
-  you can set the ``net.ipv6`` option to ``true`` :manual:`in your configuration file </reference/configuration-options/#mongodb-setting-net.ipv6>`.
+- Start your ``mongod`` in ``IPV6`` mode with :manual:`the --ipv6 flag </reference/program/mongod/#std-option-mongod.--ipv6>`, or
+  setting the ``net.ipv6`` option to ``true`` :manual:`in your configuration file </reference/configuration-options/#mongodb-setting-net.ipv6>`.
 
   .. code-block:: sh
+     :caption: Starting mongod with the --ipv6 flag
 
      mongod --ipv6
 
-- Specify ``family: 4`` as an `option to your MongoClient <{+api+}/interfaces/MongoClientOptions.html#family>`__.
+  .. code-block:: yaml
+     :caption: Setting the net.ipv6 option in the configuration file
+
+     net:
+       ipv6: true
+
+- Explicitly use ``IPV4`` by specifying ``family: 4`` as an
+  `option to your MongoClient <{+api+}/interfaces/MongoClientOptions.html#family>`__.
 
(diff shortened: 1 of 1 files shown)


source/faq.txt
```diff
@@ -12,6 +12,27 @@ Frequently Asked Questions
    :depth: 2
    :class: singlecol
 
+How to fix MongoServerSelectionError: connect ECONNREFUSED ::1:27017 error?
+---------------------------------------------------------------------------
+
+Beginning with version 17, Node defaults to using ``IPV6``. You have a few
```
ccho-mongodb, line 18: Issue: The term "Node" doesn't seem to be a common way to refer to Node.js (e.g. always referred to as Node.js on https://nodejs.dev.

Suggestion: replace "Node" with "Node.js" or "Node.js driver" (if it's the driver and not the language).

source/faq.txt
```diff
@@ -12,6 +12,27 @@ Frequently Asked Questions
    :depth: 2
    :class: singlecol
 
+How to fix MongoServerSelectionError: connect ECONNREFUSED ::1:27017 error?
```
ccho-mongodb, line 15: Issue: the noun requires an article ("a" or "the"). Also, could be easier to identify the error text.

Suggestion: separate the error text using quotation marks.
```suggestion
How to fix a "MongoServerSelectionError: connect ECONNREFUSED ::1:27017" error?
```

source/faq.txt
```diff
@@ -12,6 +12,27 @@ Frequently Asked Questions
...
+---------------------------------------------------------------------------
+
+Beginning with version 17, Node defaults to using ``IPV6``. You have a few
+options to resolve this:
```
ccho-mongodb, line 19: Identify "this" since the prior sentence refers to a default setting, not the issue.
 
```suggestion
options to resolve this error:
```

source/faq.txt
```diff
@@ -12,6 +12,27 @@ Frequently Asked Questions
...
+
+- You can start your ``mongod`` with :manual:`the --ipv6 flag </reference/program/mongod/#std-option-mongod.--ipv6>`, or
```
ccho-mongodb, line 21: Suggestion: Parallel phrasing for the listed options:

```suggestion
- Start your ``mongod`` with :manual:`the --ipv6 flag </reference/program/mongod/#std-option-mongod.--ipv6>`, or
```

source/faq.txt
```diff
@@ -12,6 +12,27 @@ Frequently Asked Questions
...
+  you can set the ``net.ipv6`` option to ``true`` :manual:`in your configuration file </reference/configuration-options/#mongodb-setting-net.ipv6>`.
```
ccho-mongodb, line 22: Suggestion: since there are two solutions listed here, it could be helpful to identify which one is shown below. Alternatively, separate each bullet point into a single option.

source/faq.txt
```diff
@@ -12,6 +12,27 @@ Frequently Asked Questions
...
+
+  .. code-block:: sh
+
+     mongod --ipv6
+
+- Specify ``family: 4`` as a `option to your MongoClient <{+api+}/interfaces/MongoClientOptions.html#family>`__.
```
ccho-mongodb: Issue: "a option" -> "an option"

Question: Is "family: 4" specifying using ipv4 and not 6? It could be helpful to identify that as the solution and then how to accomplish it.
Suggestion: "Use ipv4 instead of ipv6 and specify this as an option to your MongoClient"

source/faq.txt
ccho-mongodb: Issue:
Seems like there are a couple solutions if I understand the second one correctly: one is to use ipv6 and the other is to use ipv4. I think it could be helpful to explain those as the two alternative paths they can take to determine whether the listed solution applies to them.

Suggestion:
Bullet 1) "If your server is running on IPv4, specify the { family: 4 } option when instantiating your MongoClient"
Bullet 2) "If you can broadcast your server on IPv6, restart your server with the ..."

ccho-mongodb (approved):
A few minor issues and suggestions.

PR #456, opened 2022-10-25
AWS lambda redirect

rustagir:
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-26110-symlink/fundamentals/connection/
**And check redirect from left hand tree under Connection tab**


biniona-mongodb (approved):
LGTM, could you delete this page https://github.com/rustagir/docs-node/blob/DOCSP-26110-symlink/source/fundamentals/connection/lambda.txt as it is no longer used?

PR #469, opened 2023-01-11
DOCSP-26238: create keys from variables

rustagir:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-26238
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-26238-key-from-variable/fundamentals/typescript/#referencing-keys-that-incorporate-variables

## Self-Review Checklist

### DIFF
--- a/source/fundamentals/typescript.txt
+++ b/source/fundamentals/typescript.txt
@@ -126,7 +126,7 @@ Creating Keys from Variables
 
 If you need to query a collection or perform another operation using a
 key created from specified variables, you must use an ``as const``
-assertion when specifying the key value. This mechanism will allow your
+assertion when specifying the key. This mechanism will allow your
 code to compile successfully as long as the input types are correct.
 
 The following code snippet defines the ``ClassificationPet`` interface
--- a/source/fundamentals/typescript.txt
+++ b/source/fundamentals/typescript.txt
@@ -84,7 +84,6 @@ You can find a code snippet that shows how to specify a type for the ``FindCurso
 class in the
 :ref:`Find Multiple Documents Usage Example <node-driver-find-usage-example-code-snippet>`.
 
-
 Type Safety and Dot Notation
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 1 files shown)


source/fundamentals/typescript.txt
```diff
@@ -122,6 +121,45 @@ The type error raised by the preceding code snippet is as follows:
    ...
    Type '"Sylvilagus"' is not assignable to type 'Condition<"Canis" | "Felis">'.
 
+Creating Keys from Variables
+````````````````````````````
+
+If you need to query a collection or perform another operation using a
```
mongoKart, line 127: Issue: [Style Guide discourages "using" on its own for ambiguity reasons](https://www.mongodb.com/docs/meta/style-guide/terminology/alphabetical-terms/#u)

```suggestion
If you need to query a collection or perform another operation with a
```

source/fundamentals/typescript.txt
```diff
@@ -122,6 +121,45 @@ The type error raised by the preceding code snippet is as follows:
...
+key created from specified variables, you must use an ``as const``
```
mongoKart, line 128: Issue: Pretty sure I get this, but the wording is a little strange. When I think of a key being "created", I think of the operation that sets a value into a key that never existed (like your code sample below). Does that same thing happen when just performing a query, like you mention in the previous line? In other words, is the `as const` assertion required _any time_ you're referring to a key via a variable, or just when creating the key?

I think you can also drop the word "specified" before "variables; it's likely assumed that if the user is using a variable, they have control over it.

Suggestion: Specify the cases in which this is relevant

source/fundamentals/typescript.txt
```diff
@@ -122,6 +121,45 @@ The type error raised by the preceding code snippet is as follows:
...
+     name: string;
+     mealtimes: Mealtime[];
+   }
+   
+   interface Mealtime{
+     time: string;
+     amount: number;
+   }
+
+The following code snippet performs a find and update operation on a
```
mongoKart, line 149: [Compound modifier](https://www.mongodb.com/docs/meta/style-guide/style/punctuation/#hyphens-in-compound-modifiers)

```suggestion
The following code snippet performs a find-and-update operation on a
```

mongoKart (changes requested):
Nice. One question about the scope of use

source/fundamentals/typescript.txt
rustagir: By creating a key I just mean selecting one. The key used in the example may or may not exist -- the document is not given for us to confirm. I mean creating a key reference -- putting together a key (basically a path to a nested (it technically could be top level too) field name). `as const` is requires any time you refer to a key that incorporates a variable as a field/nested field/index/part of a field name/whatever else. I updated the subsection header and example introduction to hopefully make this more clear.

mongoKart (approved):
Perfect. Awesome work

PR #470, opened 2023-01-12
DOCSP-25706: electionid prioritization

rustagir:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-25706
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-25706-electionId-prioritzation/whats-new/#what-s-new-in-4.11

## Self-Review Checklist

### DIFF
--- a/source/whats-new.txt
+++ b/source/whats-new.txt
@@ -82,12 +82,11 @@ To learn more, see the `v4.12.0 Release Highlights <https://github.com/mongodb/n
 What's New in 4.11
 ------------------
 
-When using the 4.11 {+driver-short+} with MongoDB Server version 6.0 or
-greater, note that the driver prioritizes ``electionId`` before
-``setVersion`` during Server Discovery and Monitoring events. In
-previous versions, the prioritization order was reversed.
+When connecting to MongoDB Server version 6.0 or later, the driver prioritizes
+``electionId`` settings before ``setVersion`` settings during Server Discovery and
+Monitoring events. In previous versions, the prioritization order was reversed.
 
-The release also includes added support for **mutually
+This release also includes added support for **mutually
 recursive** collection schema types. The driver also provides type safety for
 dot-notation queries up to a depth of eight in this release. At a depth greater
 than or equal to eight, Typescript successfully compiles your code but does not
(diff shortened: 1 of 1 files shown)


source/whats-new.txt
```diff
@@ -82,7 +82,12 @@ To learn more, see the `v4.12.0 Release Highlights <https://github.com/mongodb/n
 What's New in 4.11
 ------------------
 
-The 4.11 {+driver-short+} release includes added support for **mutually
+When using the 4.11 {+driver-short+} with MongoDB Server version 6.0 or
```
ccho-mongodb, line 85: Suggestion:
- Since this is in the 4.11 header, I think it is redundant to mention the driver version (in this case). It's also redundant in the other versions as well, but that's outside the scope of this PR.
- The phrase "note that" doesn't seem necessary when introducing information.
- I try to avoid the "with" preposition whenever possible since you can usually interpret it multiple ways and can therefore be ambiguous. Instead, this could describe the action -- "When connecting to MongoDB Server v6.0 or later, the driver"

source/whats-new.txt
```diff
@@ -82,7 +82,12 @@ To learn more, see the `v4.12.0 Release Highlights <https://github.com/mongodb/n
...
+greater, note that the driver prioritizes ``electionId`` before
+``setVersion`` during Server Discovery and Monitoring events. In
```
ccho-mongodb, line 87: Suggestion:
While some readers could infer what "prioritizes X over Y" refers to, I think it's easier to understand when identifying that it's the setting value. Maybe something like this would identify the setting values more precisely:
"the driver prioritizes any value in the electionId setting before it considers values in the setVersion setting."

source/whats-new.txt
```diff
@@ -82,7 +82,12 @@ To learn more, see the `v4.12.0 Release Highlights <https://github.com/mongodb/n
...
+previous versions, the prioritization order was reversed.
+
+The release also includes added support for **mutually
```
ccho-mongodb, line 90: Suggestion: Use "this" to specify the context. E.g. "This release includes ..."

PR #586, opened 2023-02-23
DOCSP-27149 guided quick start

ccho-mongodb:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-27149>
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-27149-guided-quick-start/quick-start/

## Self-Review Checklist


source/quick-start/create-a-connection-string.txt
```diff
@@ -0,0 +1,65 @@
+.. _node-quick-start-connection-string:
+
+==========================
+Create a Connection String
+==========================
+
+You can connect to your MongoDB deployment by providing a
+**connection URI**, also called a *connection string*, which
+instructs the driver how to connect to a MongoDB deployment
```
jordan-smith721, line 9:
```suggestion
instructs the driver on how to connect to a MongoDB deployment
```

source/quick-start/create-a-connection-string.txt
```diff
@@ -0,0 +1,65 @@
...
+and how to behave while connected.
+
+The connection string includes the hostname or IP address and 
+port of your deployment, the authentication mechanism, user credentials 
+when applicable, and connection options.
+
+To connect to an instance or deployment not hosted on Atlas, see
+:ref:`Other Ways to Connect to MongoDB <node-other-ways-to-connect>`.
```
jordan-smith721, line 17: Suggestion: 
Clicking on this link takes the user out of the tutorial completely, which also collapses it in the sidebar. I think this could be confusing to some and they may struggle to find their way back, or to find where they left off when they clicked it. Could we make this a link that opens in a new tab instead to mitigate those potential issues?

jordan-smith721 (approved):
Nice work on making this! LGTM with just a couple non-blocking suggestions

source/index.txt
```diff
@@ -32,12 +28,11 @@ You can add the driver to your application to work with MongoDB
 in JavaScript. Download it using `npm <https://www.npmjs.com/>`__
 or set up a runnable project by following our Quick Start guide.
 
-
 Quick Start
 -----------
 
 Learn how to establish a connection to MongoDB Atlas and begin
-working with data in the :doc:`Quick Start </quick-start>` section.
+working with data in the step-by-step :doc:`Quick Start </quick-start>`.
```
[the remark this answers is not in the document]
ccho-mongodb: Thanks for pointing that out! Since this PR isn't initially going to be backported, I think it might be better to leave alone and address somewhere else.

source/quick-start/create-a-connection-string.txt
ccho-mongodb: Great suggestion! I think it might be ok this way for a couple reasons:

1) The quick start is designed to achieve the goal of connecting via Atlas and requires the sample data sets to be loaded. If they are connecting to a deployment not hosted on Atlas, the rest of the directions won't really work without a fair amount of sleuthing.
2) I'm not sure there's a way to specify a new tab with a `:ref:` or `:doc:` directive; the external links will be styled differently, and could be confusing since they're generally used for external links only. Not sure if there's any written guidance, but I'm guessing they're treated as two different entities in the UX design based on the appended icon that follows external links.

PR #640, opened 2023-04-04
DOCSP-18653: remove username from x509 example

rustagir:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-18653
[**Staging**](https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-18653-x509-no-user/fundamentals/authentication/mechanisms/#x.509)

## Self-Review Checklist

### DIFF
--- a/source/fundamentals/authentication/mechanisms.txt
+++ b/source/fundamentals/authentication/mechanisms.txt
@@ -252,12 +252,6 @@ Pass the location of your client certificate file as the value of
 .. literalinclude:: /code-snippets/authentication/x509.js
    :language: javascript
 
-.. note::
-
-   You do not need to specify a username when authenticating with this
-   mechanism. However, if you provide a username, you must encode it so
-   that the server parses it correctly.
-
 TLS/SSL Options
 ~~~~~~~~~~~~~~~
 
--- a/source/fundamentals/authentication/mechanisms.txt
+++ b/source/fundamentals/authentication/mechanisms.txt
@@ -233,27 +233,30 @@ The driver checks for your credentials in the following sources in order:
 
 The ``X.509`` authentication mechanism uses
 :wikipedia:`TLS <Transport_Layer_Security>` with X.509 certificates to
-authenticate your user, identified by the distinguished name (DN) of your
-client certificate. For more information on determining the *subject* name from
-the X.509 certificate, see the MongoDB manual :manual:`X.509 tutorial
-</tutorial/configure-x509-client-authentication/#add-x-509-certificate-subject-as-a-user>`.
+authenticate by retrieving the distinguished name (DN) from the
+client certificate.
 
 You can specify this authentication mechanism by setting the following
-parameters of your
-:manual:`connection string </reference/connection-string/>`:
+parameters of your :manual:`connection string </reference/connection-string/>`:
 
-- Set the ``authMechanism`` parameter to the value ``MONGODB-X509``
-- Set the ``tls`` parameter to the value ``true``
+- Set the ``authMechanism`` parameter to ``MONGODB-X509``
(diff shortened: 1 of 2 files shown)


source/fundamentals/authentication/mechanisms.txt
```diff
@@ -233,28 +233,31 @@ The driver checks for your credentials in the following sources in order:
...
+   Always **URI encode** the certificate file path using the
+   ``encodeURIComponent`` method to ensure it is parsed correctly.
 
 .. literalinclude:: /code-snippets/authentication/x509.js
    :language: javascript
 
+.. note::
+
+   You do not need to specify a username when authenticating with this
+   mechanism. However, if you provide a username, you must encode it so
```
ccho-mongodb, line 258:
```suggestion
   mechanism. However, if you provide a username, you must URI encode it so
```

source/fundamentals/authentication/mechanisms.txt
```diff
@@ -233,28 +233,31 @@ The driver checks for your credentials in the following sources in order:
...
 .. important::
-   Always **URI encode** the username and certificate file path using the
-   ``encodeURIComponent`` method to ensure they are correctly parsed.
+   Always **URI encode** the certificate file path using the
+   ``encodeURIComponent`` method to ensure it is parsed correctly.
 
 .. literalinclude:: /code-snippets/authentication/x509.js
    :language: javascript
 
+.. note::
```
ccho-mongodb, line 255: Suggestion:

Based on the comments on the ticket, it's not needed for all the versions of the Node.js driver docs we maintain. I would omit this since it might cause users to ask why they might include it.

source/fundamentals/authentication/mechanisms.txt
```diff
@@ -233,28 +233,31 @@ The driver checks for your credentials in the following sources in order:
 
 The ``X.509`` authentication mechanism uses
 :wikipedia:`TLS <Transport_Layer_Security>` with X.509 certificates to
-authenticate your user, identified by the distinguished name (DN) of your
-client certificate. For more information on determining the *subject* name from
-the X.509 certificate, see the MongoDB manual :manual:`X.509 tutorial
-</tutorial/configure-x509-client-authentication/#add-x-509-certificate-subject-as-a-user>`.
+authenticate by retrieving the distinguished name (DN) from the
+client certificate.
```
ccho-mongodb, line 237: Question:
I thought the removed sentence provides information that helps the user determine the certificate is the one that corresponds to their MongoDB user? Is there enough info in this section on how to set up x.509 authentication without it?

ccho-mongodb (approved):
LGTM! Left a suggestion and question.

source/fundamentals/authentication/mechanisms.txt
rustagir: Good call. I wasn't sure because the ticket says we might want to tell users that it is optional but requires URI encoding... But if there's really no use case where users need to supply the user, best to not reference it at all

source/fundamentals/authentication/mechanisms.txt
rustagir: N/A because removing note

source/fundamentals/authentication/mechanisms.txt
rustagir: It was my understanding that since username is no longer required, users do not need to know how to retrieve the subject from their certificate.

ccho-mongodb: Ok. It wasn't clear to me whether the link was trying to confirm the certificate was for the user or to extract the subject/username.

PR #643, opened 2023-04-05
DOCSP-26065: socketTimeoutMS behavior

rustagir:
# Pull Request Info

JIRA - [DOCSP-26065](https://jira.mongodb.org/browse/DOCSP-26065)
Staging - [FAQ](https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-26065-sockettimeout-behavior/faq/#what-happens-to-my-operations-if-my-client-disconnects-)

## Self-Review Checklist

### DIFF
--- a/source/faq.txt
+++ b/source/faq.txt
@@ -113,37 +113,31 @@ is on the other side of the planet and has a latency of 10000
 milliseconds, setting the ``connectTimeoutMS`` to anything lower will
 prevent the driver from ever connecting to that member.
 
-Should I Use "socketTimeoutMS" as a Way of Preventing Long-Running Operations From Slowing Down the Server?
--------------------------------------------------------------------------------------------------------------
-
-No, you should **not** use ``socketTimeoutMS`` to end operations that
-may run for too long and slow down the application. Attempting to do so
-may not achieve the intended result. Closing a socket is costly because
-the connection pool may create a new socket to replace the one on which
-the timeout occurred.
+What Happens to My Operations if My Client Disconnects?
+-------------------------------------------------------
 
 Starting in {+mdb-server+} version 4.2, the server terminates
 :manual:`certain operations </release-notes/4.2/#client-disconnection>` such as
 aggregations and find operations if the client disconnects before the
-operation completes.
+it successfully executes an operation.
 
-However, closing the socket does not terminate all operations, including
-all write operations. These operations continue to run on the MongoDB
-server, which could cause data inconsistencies if the application
-retries the operation on failure.
+Some operations, including all write operations, continue to run on the
+{+mdb-server+} even if the client disconnects. This behavior can cause data
+inconsistencies if your application retries the operation after failure.
 
-There are important use cases for ``socketTimeoutMS``, which
-are detailed in the following list:
+How Can I Confirm That the Driver Closed Unusable Sockets?
+----------------------------------------------------------
 
-- A MongoDB process errors out.
-- A misconfigured firewall causes a socket connection without sending a
-  ``FIN`` packet.
+If you experience unexpected network behavior or receive errors when
+executing a MongoDB process, you may not receive confirmation that the
+driver closed the corresponding socket correctly.
 
-In these cases, there is no way to detect that the connection has died.
-Setting the ``socketTimeoutMS`` option is essential to ensure that the sockets
-are closed correctly. A good rule of thumb is to set ``socketTimeoutMS``
-to two to three times the length of the slowest operation that you
-execute with the driver.
+To make sure that the driver closes sockets correctly in these cases,
+set the ``socketTimeoutMS`` option on your ``MongoClient``. Once any
+process times out, the driver will close the socket. We recommend that
+you select a value for ``socketTimeoutMS`` that is two to three times
+as long as the length of the slowest operation that your application
+executes.
 
 How Can I Prevent Sockets From Timing out Before They Become Active?
 --------------------------------------------------------------------
--- a/source/faq.txt
+++ b/source/faq.txt
@@ -113,31 +113,33 @@ is on the other side of the planet and has a latency of 10000
 milliseconds, setting the ``connectTimeoutMS`` to anything lower will
 prevent the driver from ever connecting to that member.
 
-What Happens to My Operations if My Client Disconnects?
--------------------------------------------------------
+What Happens to Running Operations If the Client Disconnects?
+-------------------------------------------------------------
 
 Starting in {+mdb-server+} version 4.2, the server terminates
-:manual:`certain operations </release-notes/4.2/#client-disconnection>` such as
-aggregations and find operations if the client disconnects before it
-successfully executes an operation.
+certain operations such as aggregations and find operations if the
+client disconnects the operations successfully completes. To see a full
+list of operations affected by this behavior, see the :manual:`Server version
+4.2 release notes </release-notes/4.2/#client-disconnection>` in the
+Server manual.
 
-Some operations, including all write operations, continue to run on the
+Other operations, such as write operations, continue to run on the
 {+mdb-server+} even if the client disconnects. This behavior can cause data
-inconsistencies if your application retries the operation after failure.
+inconsistencies if your application retries the operation after the
+client disconnects.
 
 How Can I Confirm That the Driver Closed Unusable Sockets?
 ----------------------------------------------------------
 
-If you experience unexpected network behavior or receive errors when
-executing a MongoDB process, you may not receive confirmation that the
-driver closed the corresponding socket correctly.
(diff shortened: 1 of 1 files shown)


source/faq.txt
```diff
@@ -113,33 +113,31 @@ is on the other side of the planet and has a latency of 10000
...
+Starting in {+mdb-server+} version 4.2, the server terminates
+:manual:`certain operations </release-notes/4.2/#client-disconnection>` such as
+aggregations and find operations if the client disconnects before it
+successfully executes an operation.
 
-Closing the socket causes a reconnect of the driver's connection pool,
-introducing latency to any other queued up operations. Chronically slow
-operations will, therefore, cause a large number of reconnect requests,
-negatively impacting throughput and performance.
+Some operations, including all write operations, continue to run on the
```
ccho-mongodb, line 124: Suggestion:

I think a transition signal word would help the reader immediately understand that this paragraph introduces the exception cases.

```suggestion
Other operations, such as write operations, continue to run on the
```

ccho-mongodb, line 124: Suggestion:

I think a transition signal word would help the reader immediately understand that this paragraph introduces the exception cases.
I think the adjective "some" doesn't seem to add any information and could be omitted.

```suggestion
Other operations, such as write operations, continue to run on the
```

source/faq.txt
```diff
@@ -113,33 +113,31 @@ is on the other side of the planet and has a latency of 10000
...
+{+mdb-server+} even if the client disconnects. This behavior can cause data
+inconsistencies if your application retries the operation after failure.
```
ccho-mongodb, line 126: Suggestion:

I think, but am not sure, that this refers to the disconnect as the failure and not some other type of write failure. If so, it would be helpful to describe it as a disconnect since that may not be similar to a general failure.

```suggestion
inconsistencies if your application retries the operation after a disconnection.
```

source/faq.txt
```diff
@@ -113,33 +113,31 @@ is on the other side of the planet and has a latency of 10000
...
-Also, closing the socket does not terminate the operation; it will
-continue to run on the MongoDB server, which could cause data
-inconsistencies if the application retries the operation on failure.
+How Can I Confirm That the Driver Closed Unusable Sockets?
+----------------------------------------------------------
 
-However, there are important use cases for ``socketTimeoutMS`` -
-consider the following cases:
+If you experience unexpected network behavior or receive errors when
+executing a MongoDB process, you may not receive confirmation that the
```
ccho-mongodb, line 132:
```suggestion
If you experience unexpected network behavior or if a MongoDB process
exits with an error, you may not receive confirmation that the
```

source/faq.txt
```diff
@@ -113,33 +113,31 @@ is on the other side of the planet and has a latency of 10000
...
+driver closed the corresponding socket correctly.
```
ccho-mongodb, line 133: Suggestion:
The adverb "correctly" modifies the closure, so it's usually easier to associate when placed next to word it's related to.

```suggestion
driver correctly closed the corresponding socket.
```

source/faq.txt
```diff
@@ -113,33 +113,31 @@ is on the other side of the planet and has a latency of 10000
...
-- A MongoDB process erroring out
-- A misconfigured firewall causing a socket connection without sending a ``FIN`` packet
-
-In those cases, there is no way to detect that the connection has died.
-Setting the ``socketTimeoutMS`` is essential to ensure that the sockets
-are closed correctly. A good rule of thumb is to set ``socketTimeoutMS``
-to two to three times the length of the slowest operation that runs
-through the driver.
+To make sure that the driver closes the socket correctly in these cases,
+set the ``socketTimeoutMS`` option on your ``MongoClient`` instance. Once any
```
ccho-mongodb, line 136: Suggestion:

Just in case it can be on the connection string, maybe it would be more flexible to call socketTimoutMS a connection option.

```suggestion
set the ``socketTimeoutMS`` connection option. When a
```

source/faq.txt
```diff
@@ -113,33 +113,31 @@ is on the other side of the planet and has a latency of 10000
...
+process times out, the driver will successfully close the socket. We
```
ccho-mongodb, line 137: Suggestion:

I think it could be helpful to describe the process as a "MongoDB" one as done earlier since it's a general term.

I think it's not a definite outcome that the driver will _successfully_ close it as something else could go wrong.

```suggestion
MongoDB process times out, the driver will close the socket. We
```

source/faq.txt
```diff
@@ -113,33 +113,31 @@ is on the other side of the planet and has a latency of 10000
...
+recommend that you select a value for ``socketTimeoutMS`` that is two to
+three times as long as the length of the slowest operation that your
```
ccho-mongodb, line 139:
```suggestion
three times as long as the expected duration of the slowest operation that your
```

source/faq.txt
```diff
@@ -156,18 +154,19 @@ avoid closing.
 
 One message every 3000 milliseconds is not enough to keep the sockets
 active, so several of the sockets will time out after 5000 milliseconds.
-Reduce the ``poolSize`` in the connection settings to fix the problem.
+Reduce the ``maxPoolSize`` option in the connection settings to lower
+the number of connection the driver can maintain in the connection
+pool.
```
ccho-mongodb, line 159: Issue:

I like the intent of providing more specificity in what the user should try to accomplish by changing the setting to fix the problem. However, there are a couple minor issues:

- I think "reduce X option" doesn't seem like the right subject-verb agreement. Since it corresponds to the "lower the number of [connections]", the verb only needs to be used once.
- Also, I think "number of connection" should be "number of connections".

Suggestion:

```suggestion
To avoid excessive socket timeouts, lower the number of connections that
the driver can maintain in the connection pool, specified by the ``maxPoolSize`` 
option.
```

source/faq.txt
```diff
@@ -156,18 +154,19 @@ avoid closing.
...
 
-To specify the optional ``poolSize`` setting for your ``MongoClient``, declare
+To specify the optional ``maxPoolSize`` setting for your ``MongoClient``, declare
 it in the ``options`` object of the constructor as follows:
 
 .. code-block:: javascript
 
    const client = new MongoClient(uri, {
-     poolSize: <integer value>,
+     maxPoolSize: <integer value>,
```
ccho-mongodb, line 167: Question: does this change in the setting name apply to all versions of the driver that we document? If it applies to a different set of versions than the other changes, perhaps a different PR would work better. Alternatively, maybe setting a source constant for this with a value that differs based on version might work too.

source/faq.txt
```diff
@@ -256,10 +255,9 @@ a separate connection pool for the slow operation, isolating it from
 other, faster operations.
 
 .. note::
-    If the number of operations is greater than the set ``poolSize`` and
+    If the number of operations is greater than the set ``maxPoolSize`` and
```
ccho-mongodb, line 258:
```suggestion
    If the number of operations is greater than the ``maxPoolSize`` and
```

ccho-mongodb (changes requested):
This version looks much better, thanks for making the changes! Mostly minor suggestions, would like to take a quick look after you've made the updates.

source/faq.txt
rustagir: Yes, since Node v3.5 this is the correct name for this option

ccho-mongodb (approved):
LGTM! I think there's a run-on sentence that could be broken up, but no need for another review.

source/faq.txt
```diff
@@ -113,33 +113,33 @@ is on the other side of the planet and has a latency of 10000
...
-Setting the ``socketTimeoutMS`` is essential to ensure that the sockets
-are closed correctly. A good rule of thumb is to set ``socketTimeoutMS``
-to two to three times the length of the slowest operation that runs
-through the driver.
+What Happens to Running Operations If the Client Disconnects?
+-------------------------------------------------------------
+
+Starting in {+mdb-server+} version 4.2, the server terminates
+certain operations such as aggregations and find operations if the
+client disconnects the operations successfully completes. To see a full
```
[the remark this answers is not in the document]
rustagir: eliminated a clause to shorten

source/faq.txt
```diff
@@ -113,33 +113,33 @@ is on the other side of the planet and has a latency of 10000
...
+list of operations affected by this behavior, see the :manual:`Server version
+4.2 release notes </release-notes/4.2/#client-disconnection>` in the
+Server manual.
+
+Other operations, such as write operations, continue to run on the
+{+mdb-server+} even if the client disconnects. This behavior can cause data
```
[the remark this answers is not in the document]
rustagir: Not changing because it seems like the mechanism is not that the server "allows" certain operations to continue running, but rather that it marks certain operations for termination.

PR #682, opened 2023-05-12
DOCSP-29850: v5.5 release version updates

rustagir:
# Pull Request Info

JIRA - [DOCSP-29850](https://jira.mongodb.org/browse/DOCSP-29850)
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-29850-v5.5-release/
[compat](https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-29850-v5.5-release/compatibility/#mongodb-compatibility)
[whats new](https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-29850-v5.5-release/whats-new/#std-label-version-5.5)

## Self-Review Checklist


source/whats-new.txt
```diff
@@ -36,6 +37,18 @@ Learn what's new in:
...
 
+.. _version-5.5:
+
+What's New in 5.5
+-----------------
+
+New features of the 5.5 {+driver-short+} release include:
+
+- The driver now accurately detects Function-as-a-Service (FaaS)
+  environments in AWS.
```
ccho-mongodb, line 48: Issue:
This seems like it fixes a bug, but that bug hasn't clearly been described. From reading this, I'm not sure what it means to more accurately detect a FaaS environment in AWS.

Suggestion:
Provide more detail on what the behavior was previously and what it more accurately detects.

source/whats-new.txt
```diff
@@ -36,6 +37,18 @@ Learn what's new in:
...
+- You must upgrade ``mongodb-client-encryption`` to version 2.8.0 or
+  later if you want to create a QE collection while using this driver version.
```
ccho-mongodb, line 50: Issue:

1) As recommended by [the style guide](https://www.mongodb.com/docs/meta/style-guide/style/abbreviations/), explain the terms before using abbreviations.

2 I think) QE collection is not the official name for encrypted collection; instead, use "encrypted collection" as referred to on this page.

Suggestion:

1) Use "Queryable Encryption".

2) Use "encrypted collection" as it is referred to on [this page](https://www.mongodb.com/docs/manual/core/queryable-encryption/fundamentals/manage-collections/).

Both suggestions incorporated:
"You must use ... v2.8.0 or later to encrypt collections by using the Queryable Encryption feature."

ccho-mongodb (changes requested):
Thanks for getting to these updates quickly! I think that the What's New items need to be more detailed so that users are confident they understand the changes in the new version.

source/whats-new.txt
rustagir: I guess the "in this this driver version" is implicit

source/whats-new.txt
```diff
@@ -36,6 +37,20 @@ Learn what's new in:
...
+.. _version-5.5:
+
+What's New in 5.5
+-----------------
+
+New features of the 5.5 {+driver-short+} release include:
+
+- The driver now accurately detects Function-as-a-Service (FaaS)
+  environments in AWS by considering AWS environment variables only if
+  they begin with ``AWS_Lambda_``.
```
ccho-mongodb, line 49: Good to know, thanks for clarifying this!

source/whats-new.txt
ccho-mongodb: > I guess the "in this this driver version" is implicit

I agree, and thanks for catching the mistake in my suggestion.

I think there is also nuance here in which omitting the driver version is important. From what I know so far, the previous version of the driver won't be able to create encrypted collections when using older versions of mongodb-client-encryption either as that functionality will be removed from the server too.

ccho-mongodb (approved):
LGTM

PR #690, opened 2023-05-17
Docsp 29488 - Add Connection Troubleshooting Guide

DBirtolo-mdb:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-29488>
Staging - <https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-29488/connection-troubleshooting/>

## Self-Review Checklist

### DIFF
--- a/source/connection-troubleshooting.txt
+++ b/source/connection-troubleshooting.txt
@@ -1,0 +1,319 @@
+.. _node-connection-troubleshooting:
+
+==========================
+Connection Troubleshooting
+==========================
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+This page offers potential solutions to issues you might encounter when
+using the {+driver-long+} to connect to a MongoDB deployment.
+
+.. note::
+
+   This page addresses only connection issues. If you encounter any other issues
+   with MongoDB or the driver, visit the following resources:
+
+   - The :ref:`Frequently Asked Questions (FAQ) <node-faq>` for the
+     {+driver-short+}
+   - The :ref:`Issues & Help <node-issues-help>` page, which has
+     information about reporting bugs, contributing to the driver, and 
+     finding additional resources
+   - The `MongoDB Community Forums <https://community.mongodb.com>`__ for
+     questions, discussions, or general technical support
+
+Connection Error
+~~~~~~~~~~~~~~~~
+
+The following error message is a general message indicating that the driver
+cannot connect to a server on the specified hostname or port:
+
+.. code-block:: none
+   :copyable: false
+
+   Error: couldn't connect to server 127.0.0.1:27017
+
+The following sections describe methods that may help resolve the issue.
+
+.. _node-troubleshooting-connection-string-port:
+
+Check Connection String
+-----------------------
+
+Verify that the hostname and port number in the connection string are both
+accurate. In the sample error message, the hostname is ``127.0.0.1`` and the
+port is ``27017``. The default port value for a MongoDB instance is
+``27017``, but you can configure MongoDB to communicate on another port.
+
+.. _node-troubleshooting-connection-firewall:
+
+Configure Firewall
+------------------
+
+Assuming that your MongoDB deployment uses the default port, verify that port
+``27017`` is open in your firewall. If your deployment uses a different port,
+verify the correct port is open in your firewall.
+
+.. warning::
+
+   Do not open a port in your firewall unless you are sure that it's the port
+   used by your MongoDB instance.
+
+ECONNREFUSED Error
+~~~~~~~~~~~~~~~~~~
+
+If the connection is refused when the driver attempts to connect to the MongoDB
+instance, it generates this error message:
+
+.. code-block:: none
+   :copyable: false
+
+   MongoServerSelectionError: connect ECONNREFUSED ::1:27017
+
+The following sections describe methods that may help resolve the issue.
+
+Start mongod in IPV6 Mode
+-------------------------
+
+Beginning with version 17, Node.js defaults to using ``IPV6``. You can start
+your ``mongod`` in ``IPV6`` mode with
+:manual:`the --ipv6 flag </reference/program/mongod/#std-option-mongod.--ipv6>`,
+or setting the ``net.ipv6`` option to ``true``
+:manual:`in your configuration file </reference/configuration-options/#mongodb-setting-net.ipv6>`.
+
+  .. code-block:: sh
+     :caption: Starting mongod with the --ipv6 flag
+
+     mongod --ipv6
+
+  .. code-block:: yaml
+     :caption: Setting the net.ipv6 option in the configuration file
+
+     net:
+       ipv6: true
+
+Use IPV4 with Your Mongo Client
+-------------------------------
+
+Beginning with version 17, Node.js defaults to using ``IPV6``. You can
+explicitly use ``IPV4`` with your client by specifying ``family: 4`` as an
+`option to your MongoClient <{+api+}/interfaces/MongoClientOptions.html#family>`__.
+
+  .. code-block:: js
+
+      const client = new MongoClient(uri, {
+        family: 4,
+      });
+
+ECONNRESET Error
+~~~~~~~~~~~~~~~~
+
+If the connection is reset when the driver calls ``client.connect()``, it
+generates this error message:
+
+.. code-block:: none
+   :copyable: false
+
+   MongoServerSelectionError: connect ECONNRESET ::1:27017
+
+The following section describes a method that may help resolve the issue.
+
+Increase Number of File Descriptors
+-----------------------------------
+
+In most operating systems, each connection is associated with a `file
+descriptor
+<https://www.computerhope.com/jargon/f/file-descriptor.htm>`_. There is
+typically a limit set by the operating system on the number of file
+descriptors used by a single process. An ``ECONNRESET`` error can occur
+if the connection pool size surpasses the limit of ``file descriptors``.
+
+Consider the following operation:
+
+.. literalinclude:: /code-snippets/faq/econnresetWithClientConnect-example.js
+  :language: javascript
+  :linenos:
+
+If this operation causes an ``ECONNRESET`` error, you may have run into
+the ``file descriptor`` limit for your Node.js process. In that case you
+must increase the number of ``file descriptors`` for the Node.js process. On
+MacOS and Linux you do this with the `ulimit
+<https://ss64.com/bash/ulimit.html>`_ shell command.
+
+.. code-block:: sh
+
+   ulimit -n 6000
+
+This sets the maximum number of ``file descriptors`` for the process to
+6000, allowing Node.js to connect with a pool size of 5000 sockets.
+
+Authentication Error
+~~~~~~~~~~~~~~~~~~~~
+
+The {+driver-short+} can fail to connect to a MongoDB instance if
+the authorization is not configured correctly. In these cases, the driver raises
+an error message similar to one of the following messages:
+
+.. code-block:: none
+   :copyable: false
+
+   Command failed with error 18 (AuthenticationFailed): 'Authentication
+   failed.' on server localhost:27017.
+
+.. code-block:: none
+   :copyable: false
+
+   connection() error occurred during connection handshake: auth error:
+   sasl conversation error: unable to authenticate using mechanism
+   "SCRAM-SHA-256": (AuthenticationFailed) Authentication failed.
+
+The following sections describe methods that may help resolve the issue.
+
+.. _node-troubleshooting-connection-string-auth:
+
+Check Connection String
+-----------------------
+
+An invalid connection string is the most common cause of authentication
+issues when attempting to connect to MongoDB.
+
+.. tip::
+
+   For more information about connection strings,
+   see :ref:`Connection URI <node-connection-uri>` in the Connection Guide.
+
+If your connection string contains a username and password, ensure that they
+are in the correct format.
+
+.. note::
+
+   If the username or password includes any of the following characters, they
+   must be `percent encoded <https://tools.ietf.org/html/rfc3986#section-2.1>`__:
+
+   .. code-block:: none
+
+      : / ? # [ ] @
+
+When connecting to a replica set, you should include all of the hosts
+in the replica set in your connection string. Separate each of the hosts
+in the connection string with a comma. This enables the driver to establish a
+connection if one of the hosts is unreachable.
+
+.. _node-troubleshooting-connection-admin:
+
+Verify User Is in Authentication Database
+-----------------------------------------
+
+To successfully authenticate a connection by using a username and password,
+the username must be defined in the authentication database. The default
+authentication database is the ``admin`` database. To use a different database
+for authentication, specify the ``authSource`` in the connection string. The
+following example instructs the driver to use ``users`` as the authentication
+database:
+
+.. code-block:: javascript
+   :copyable: true
+
+   const { MongoClient } = require("mongodb");
+   const uri = "mongodb://<username>:<password>@<hostname>:<port>/?authSource=users";
+   const client = new MongoClient(uri);
+
+Error Sending Message
+~~~~~~~~~~~~~~~~~~~~~
+
+When the driver fails to send a command after you make a request,
+it often displays the following general error message:
+
+.. code-block:: none
+   :copyable: false
+
+   com.mongodb.MongoSocketWriteException: Exception sending message
+
+The following sections describe methods that may help resolve the issue.
+
+Check Connection String
+-----------------------
+
+Verify that the connection string in your app is accurate. For more information
+about verifying your connection string, see
+:ref:`Connection Error <node-troubleshooting-connection-string-port>`
+and :ref:`Authentication Error <node-troubleshooting-connection-string-auth>`.
+
+Verify User Is in Authentication Database
+-----------------------------------------
+
+Verify the user is in the correct authentication database. For more
+information about the authentication database, see
+:ref:`Authentication Error <node-troubleshooting-connection-admin>`.
+
+Configure Firewall
+------------------
+
+The firewall needs to have an open port for communicating with the MongoDB
+instance. For more information about configuring the firewall, see
+:ref:`Connection Error <node-troubleshooting-connection-firewall>`.
+
+.. _node-troubleshooting-connection-number-connections:
+
+Check the Number of Connections
+-------------------------------
+
+This solution could apply to the generic message, as well as the following
+specific error message:
+
+.. code-block:: none
+   :copyable: false
+
+   connection refused because too many open connections
+
+Each ``MongoClient`` instance supports a maximum number of concurrent open
+connections in its connection pool. The configuration parameter ``maxPoolSize``
+defines this value and is set to ``100`` by default. If there are already a
+number of open connections equal to ``maxPoolSize``, the server waits until
+a connection becomes available. If this wait time exceeds the ``maxIdleTimeMS``
+value, the driver responds with an error.
+
+For more information about how connection pooling works, see
+:ref:`How Does Connection Pooling Work in the Node Driver? <node-faq-connection-pool>`
+in the FAQ.
+
+Timeout Error
+~~~~~~~~~~~~~
+
+Sometimes when you send a request through the driver to the server, the request
+times out. When this happens, you might receive an error message
+similar to the following message:
+
+.. code-block:: none
+   :copyable: false
+
+   timed out while checking out a connection from connection pool: context canceled
+
+If you receive this error, try the following methods to resolve the
+issue.
+
+Set Timeout Option
+------------------
+
+You can specify the ``timeoutMS`` option in your connection string to define
+how long the driver will wait to connect.
+
+The following example sets the single timeout value to 5 seconds using the
+connection string option:
+
+.. code-block:: javascript
+
+   const { MongoClient } = require("mongodb");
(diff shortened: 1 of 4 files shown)


source/connection-troubleshooting.txt
```diff
@@ -0,0 +1,325 @@
...
+-----------------------------------
+
+In most operating systems, each connection is associated with a `file
+descriptor
+<https://www.computerhope.com/jargon/f/file-descriptor.htm>`_. There is
+typically a limit set by the operating system on the number of file
+descriptors used by a single process. An ``ECONNRESET`` error can occur
+if the connection pool size surpasses the limit of ``file descriptors``.
+
+Consider the following operation:
```
ccho-mongodb, line 135: Suggestion:

I think it could be helpful to be more descriptive of the code example, especially since there isn't any additional text after it that directs the reader's attention to what they should be looking at in it.

```suggestion
Consider the following connection code that sets the maximum size of the connection pool:
```

source/connection-troubleshooting.txt
```diff
@@ -0,0 +1,325 @@
...
+must increase the number of ``file descriptors`` for the Node.js process. On
+MacOS and Linux you do this with the `ulimit
+<https://ss64.com/bash/ulimit.html>`_ shell command.
+
+.. code-block:: sh
+
+   ulimit -n 6000
+
+This sets the maximum number of ``file descriptors`` for the process to
+6000, allowing Node.js to connect with a pool size of 5000 sockets.
```
ccho-mongodb, line 152: Suggestion:
I think file descriptors are used for more than just connections, so maybe there is no guarantee. I think it is sufficient to draw the connection between file descriptors and connections, perhaps earlier in the description.

E.g.
"The driver needs to use file descriptors to reference each connection in its connection pool."

source/connection-troubleshooting.txt
```diff
@@ -0,0 +1,325 @@
...
+
+Authentication Error
+~~~~~~~~~~~~~~~~~~~~
+
+The {+driver-short+} can fail to connect to a MongoDB instance if
+the authorization is not configured correctly. In these cases, the driver raises
+an error message similar to one of the following messages:
```
ccho-mongodb, line 159: Suggestion:

I think it would be worthwhile to identify to the user that the example and troubleshooting advice is specifically related to ``SCRAM-SHA-256`` authentication and not any other type (e.g. ``LDAP``, ``TLS``, ``MONGODB-AWS``, etc. This applies to all the sub-sections.

E.g.
"fail to connect to a MongoDB deployment if your username and password are not configured for SCRAM-SHA-256 authentication correctly."

source/connection-troubleshooting.txt
```diff
@@ -0,0 +1,325 @@
...
+
+.. note::
+
+   If the username or password includes any of the following characters, they
+   must be `percent encoded <https://tools.ietf.org/html/rfc3986#section-2.1>`__:
+
+   .. code-block:: none
+
+      : / ? # [ ] @
+
```
ccho-mongodb, line 200: Suggestion:
I think it would be helpful in this case to show the code that they can use to percent encode their username/password.

source/connection-troubleshooting.txt
```diff
@@ -0,0 +1,325 @@
...
+When connecting to a replica set, you should include all of the hosts
+in the replica set in your connection string. Separate each of the hosts
+in the connection string with a comma. This enables the driver to establish a
+connection if one of the hosts is unreachable.
```
ccho-mongodb, line 204: Suggestion:
This seems unrelated to authentication; would it be ok to remove?

source/connection-troubleshooting.txt
```diff
@@ -0,0 +1,325 @@
...
+
+.. _node-troubleshooting-connection-admin:
+
+Verify User Is in Authentication Database
+-----------------------------------------
+
+To successfully authenticate a connection by using a username and password,
```
ccho-mongodb, line 211: Suggestion:
I think it would be helpful to link to info on how to check for a user's permissions in the authentication database or perhaps a sample query. Otherwise, the user doesn't have a good path forward to checking whether their user is in the default admin database.
Also could be good to mention that connecting to a locally-hosted MongoDB deployment from the same machine doesn't require any auth so that, for those situations, the user could eliminate this as a potential solution.

source/connection-troubleshooting.txt
```diff
@@ -0,0 +1,325 @@
...
+
+   const { MongoClient } = require("mongodb");
+   const uri = "mongodb://<username>:<password>@<hostname>:<port>/?authSource=users";
+   const client = new MongoClient(uri);
+
+Error Sending Message
+~~~~~~~~~~~~~~~~~~~~~
+
+When the driver fails to send a command after you make a request,
+it often displays the following general error message:
```
ccho-mongodb, line 229: Suggestion:
When describing a hypothetical situation, describing something as occurring "often" may not carry the same meaning as "may or may not", the latter of which I think you intended here. Also, I don't think the page established what a "general " classification of an error message is, so I would avoid using that adjective.

```suggestion
it may display the following error message:
```

source/connection-troubleshooting.txt
```diff
@@ -0,0 +1,325 @@
...
+   :copyable: false
+
+   com.mongodb.MongoSocketWriteException: Exception sending message
+
+The following sections describe methods that may help resolve the issue.
+
+Check Connection String
+-----------------------
+
+Verify that the connection string in your app is accurate. For more information
```
ccho-mongodb, line 241: Suggestion:

I think this is the first mention of an app. I think this should either be consistently used or omitted.

```suggestion
Verify that the connection string is accurate. For more information
```

source/connection-troubleshooting.txt
```diff
@@ -0,0 +1,325 @@
...
+it often displays the following general error message:
+
+.. code-block:: none
+   :copyable: false
+
+   com.mongodb.MongoSocketWriteException: Exception sending message
+
+The following sections describe methods that may help resolve the issue.
+
+Check Connection String
```
ccho-mongodb, line 238: Question:
I'm not certain how this is related to failing to send a command. If they did not get a connection-related error previously, what circumstances could make this error happen that are related to the connection string? Perhaps there are some options on the connection string that might cause this? If so, could those be included here?

source/connection-troubleshooting.txt
```diff
@@ -0,0 +1,325 @@
...
+-----------------------
+
+Verify that the connection string in your app is accurate. For more information
+about verifying your connection string, see
+:ref:`Connection Error <node-troubleshooting-connection-string-port>`
+and :ref:`Authentication Error <node-troubleshooting-connection-string-auth>`.
+
+Verify User Is in Authentication Database
+-----------------------------------------
```
ccho-mongodb, line 247: Suggestion: 
Perhaps "Verify User Permissions in the Authentication Database" would be more specific. The user can be denied permissions in the auth database, so it wouldn't be limited to checking for existence there.

source/connection-troubleshooting.txt
```diff
@@ -0,0 +1,325 @@
...
+Check the Number of Connections
+-------------------------------
+
+This solution could apply to the generic message, as well as the following
+specific error message:
+
+.. code-block:: none
+   :copyable: false
+
+   connection refused because too many open connections
```
ccho-mongodb, line 271: Suggestion:

Since the rest are split out by error message and potential solutions, I think it would make sense to split this one into its own heading to ensure there's a uniform structure and therefore expectation of where to find information.

source/connection-troubleshooting.txt
```diff
@@ -0,0 +1,325 @@
...
+   timed out while checking out a connection from connection pool: context canceled
+
+If you receive this error, try the following methods to resolve the
+issue.
+
+Set connectTimeoutMS
+--------------------
+
+To prevent the driver from hanging during connection or to prevent the
+driver from spending too long trying to reach unreachable replica sets,
```
ccho-mongodb, line 303: Issue:
I think "replica set node" or "node" is the appropriate term. The "replica set" is the term that describes the topology of multiple nodes.

Suggestion:

```suggestion
driver from spending too much time attempting to reach unreachable replica set nodes,
```

source/connection-troubleshooting.txt
```diff
@@ -0,0 +1,325 @@
...
+you can set the ``connectTimeoutMS`` option of your
+:ref:`connection options <node-connection-options>`.
+You should ensure the ``connectTimeoutMS`` setting is not lower than
+the longest network latency you have to a member of the set. If one of the
```
ccho-mongodb, line 307:
```suggestion
the highest network latency you have to a member of the set. If one of the
```

source/connection-troubleshooting.txt
```diff
@@ -0,0 +1,325 @@
...
+secondary members is on the other side of the planet and has a latency of 10000
+milliseconds, setting the ``connectTimeoutMS`` to anything lower will
+prevent the driver from ever connecting to that member.
```
ccho-mongodb, line 310: Suggestion:

Continue to use concrete numbers to make the comparison. Avoid future tense when possible.
 
```suggestion
milliseconds, setting the ``connectTimeoutMS`` to 9000 prevents
prevent the driver from connecting to that member.
```

source/connection-troubleshooting.txt
```diff
@@ -0,0 +1,325 @@
...
+specific error message:
+
+.. code-block:: none
+   :copyable: false
+
+   connection refused because too many open connections
+
+Each ``MongoClient`` instance supports a maximum number of concurrent open
+connections in its connection pool. The configuration parameter ``maxPoolSize``
+defines this value and is set to ``100`` by default. If there are already a
```
ccho-mongodb, line 275: Suggestion:

Avoid passive voice. I think it could be good to provide the reader with information on what actions they can take rather than mention the existence of something.

```suggestion
connections in its connection pool. You can configure the ``maxPoolSize``
parameter which has a default value of ``100``. If there are already a
```

source/connection-troubleshooting.txt
```diff
@@ -0,0 +1,325 @@
...
+
+For more information about how connection pooling works, see
+:ref:`How Does Connection Pooling Work in the Node Driver? <node-faq-connection-pool>`
+in the FAQ.
+
+Timeout Error
+~~~~~~~~~~~~~
+
+Sometimes when you send a request through the driver to the server, the request
+times out. When this happens, you might receive an error message
```
ccho-mongodb, line 288: Suggestion: 
Put the blame on the network rather than leave the cause unspecified ("sometimes [it happens]").

E.g.
"When the network is not able to deliver a request from the driver to the server quickly enough, it can time out."

ccho-mongodb (changes requested):
Lots of great info here. 

I think making the solutions more actionable, linking out to server docs for administrator-type actions, and keeping the structure and terminology uniform will definitely help users find and solve their related issues.

source/connection-troubleshooting.txt
```diff
@@ -0,0 +1,325 @@
...
+   - The :ref:`Issues & Help <node-issues-help>` page, which has
+     information about reporting bugs, contributing to the driver, and 
+     finding additional resources
+   - The `MongoDB Community Forums <https://community.mongodb.com>`__ for
+     questions, discussions, or general technical support
+
+Connection Error
+~~~~~~~~~~~~~~~~
+
+The following error message is a general message indicating that the driver
```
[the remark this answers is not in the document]
DBirtolo-mdb: Unfortunately not. While that is one of the reasons this error can be generated, this is also a catch all that gets generated if there is not a more specific message to filter up. Since this can be generated for a number of reasons as a general catch-all, I can't go into detail for all the reasons it can be generated.

source/connection-troubleshooting.txt
```diff
@@ -0,0 +1,325 @@
...
+   MongoServerSelectionError: connect ECONNRESET ::1:27017
+
+The following section describes a method that may help resolve the issue.
+
+Increase Number of File Descriptors
+-----------------------------------
+
+In most operating systems, each connection is associated with a `file
+descriptor
+<https://www.computerhope.com/jargon/f/file-descriptor.htm>`_. There is
```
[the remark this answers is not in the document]
DBirtolo-mdb: I thought the link was approved since this was a copy from the old FAQ (https://www.mongodb.com/docs/drivers/node/current/faq/#what-can-i-do-if-i-m-getting--econnreset--when-calling--client.connect----) and moved over (as opposed to completely new authored content). I will remove the link when I reword the section.

source/connection-troubleshooting.txt
```diff
@@ -0,0 +1,325 @@
...
+
+.. literalinclude:: /code-snippets/faq/econnresetWithClientConnect-example.js
+  :language: javascript
+  :linenos:
+
+If this operation causes an ``ECONNRESET`` error, you may have run into
+the ``file descriptor`` limit for your Node.js process. In that case you
+must increase the number of ``file descriptors`` for the Node.js process. On
+MacOS and Linux you do this with the `ulimit
+<https://ss64.com/bash/ulimit.html>`_ shell command.
```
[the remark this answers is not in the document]
DBirtolo-mdb: See above about assuming the link was acceptable since it was already published and the content was moved.

source/connection-troubleshooting.txt
DBirtolo-mdb: I was trying to avoid duplicating too much and avoiding too many internal links. The issue is that this solution could work either for the generic error message or for this specific error message. It felt  like it was too much repetition for very similar types of errors.

source/connection-troubleshooting.txt
DBirtolo-mdb: This specific issue is something that was documented on multiple forum posts that the programmer created the user in another database other than the default, such as `userPermissions` or such. When they moved the user over to that database or specified that database to use as the authentication database, it worked. So I believe this terminology is correct. The user itself needs to exist in the database.

Verifying user permissions is also an issue and can be a different item (although interestingly, that didn't come up in any searching - maybe because it is standard enough that folks know to check it).

source/connection-troubleshooting.txt
DBirtolo-mdb: I believe this can happen if the user does not have the appropriate permissions, so the thought is to make sure the user in the connection string is connect. I will specify in the content.

source/connection-troubleshooting.txt
DBirtolo-mdb: Due to another comment, I added a section about verifying the users permissions that links to the RBAC docs. This is specifically about making sure the user is in the right database, because that is a common issue based on forum hits.

source/connection-troubleshooting.txt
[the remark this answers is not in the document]
DBirtolo-mdb: I decided not to include the external command because I don't think that we should be recommending that. I still want to mention it, but don't want to give exact instructions for something that can have significant ripple effects.

source/connection-troubleshooting.txt
ccho-mongodb: Revisiting the doc, I don't remember whether which error this specific one was tied to (there are two duplicate headers), but if this one is tied to "com.mongodb.MongoSocketWriteException: Exception sending message", I would guess that this has to do with the network and not with authentication nor authorization. I would double check with DBX engineers or get an external review to verify this potential cause.

Regarding authentication and authorization:

Unless I'm mistaken, if the user does not have any roles, they cannot perform any actions. When the driver attempts to run a command with that user for a command they do not have permissions in their role for, the server will return an error.

When you create a user in the authentication database, you must define their authorization roles to have permissions to do anything. E.g. https://www.mongodb.com/docs/manual/tutorial/create-users/#create-additional-users-for-your-deployment

In a local deployment, you don't need a user and automatically get admin access as that unnamed user.

source/connection-troubleshooting.txt
```diff
@@ -0,0 +1,343 @@
...
+are sending. MongoDB uses Role-Based Access Control (RBAC) to control access
+to a MongoDB system. For more information about how to configure RBAC in MongoDB,
+see :manual:`Default MongoDB Port </core/authorization/>`.
+
+Verify the User Is in the Authentication Database
+-------------------------------------------------
+
+Verify the user is in the correct authentication database. For more
+information about the authentication database, see
+:ref:`Authentication Error <node-troubleshooting-connection-admin>`.
```
ccho-mongodb, line 257: Suggestion:
I think the content in this section is subsumed by the info in the prior header "Check User Permissions" and therefore should be omitted. If the reader checks for the user permissions, they must find the user first.

source/connection-troubleshooting.txt
```diff
@@ -0,0 +1,343 @@
...
+Set connectTimeoutMS
+--------------------
+
+To prevent the driver from hanging during connection or to prevent the
+driver from spending too much time attempting to reach unreachable replica set
+nodes, you can set the ``connectTimeoutMS`` option of your
+:ref:`connection options <node-connection-options>`.
+You should ensure the ``connectTimeoutMS`` setting is not lower than
+the highest network latency you have to a member of the set. If one of the
+secondary members is on the other side of the planet and has a latency of 10000
```
ccho-mongodb, line 326: Suggestion:

I think it's possible to leave out the potential cause for latency since that isn't important to the explanation but might be perceived that way. "If [both] X and Y, then..."

```suggestion
secondary members has a latency of 10000
```

source/connection-troubleshooting.txt
```diff
@@ -0,0 +1,343 @@
...
+if the number of connections exceeds this limit.
+
+Consider the following example that creates a client and attempts to connect to
+a local deployment:
+
+.. literalinclude:: /code-snippets/faq/econnresetWithClientConnect-example.js
+  :language: javascript
+  :linenos:
+
+This operation can create a maximum number of 5000 connections as indicated by
```
ccho-mongodb, line 135: Issue: 
1) I think claiming that the "operation can create... connections" is incorrect. 

2) The example only really shows how to specify maxPoolSize on the connection URI. 


Suggestion:

1) I don't think `connect()` is responsible for creating 5000 connections. I think the `connect()` operation has been a no-op in the Node.js driver for a few years which is why our connection examples need to perform a command to determine whether the connection was successful. 

The driver manages the connection pool. The operation only shows how you specify the setting in the connection string. I would present a scenario without using the code example, and appropriately attribute connection creation to the driver.

2) I think the explanation included in this paragraph is important, but the code snippet does not match the intent of the description. I think either of the following might be better:
- Omitting it 
- Relocating it to after explaining maxPoolSize setting explanation, and using it just to show how to set it on the connection URI.

source/connection-troubleshooting.txt
ccho-mongodb: Question:
Did you get a chance to look at this comment?

source/connection-troubleshooting.txt
```diff
@@ -0,0 +1,343 @@
...
+   For more information about connection strings,
+   see :ref:`Connection URI <node-connection-uri>` in the Connection Guide.
+
+If your connection string contains a username and password, ensure that they
+are in the correct format.
+
+.. note::
+
+   If the username or password includes any of the following characters, they
+   must be `percent encoded <https://tools.ietf.org/html/rfc3986#section-2.1>`__:
```
ccho-mongodb, line 191: Suggestion:

This seems like a distinct item to check rather than supplemental information that might go in an admonition. I think it would make sense to move the content outside of an admonition.

ccho-mongodb (changes requested):

source/connection-troubleshooting.txt
DBirtolo-mdb: I pulled it into the main text of the topic, because it feels like it fits strongly with verifying that the URI string is correct.

source/connection-troubleshooting.txt
```diff
@@ -74,27 +75,27 @@ instance, it generates this error message:
...
+Ensure MongoDB and Your Client Use the Same Protocol
+----------------------------------------------------
 
 In Node.js v17 and later, the DNS resolver uses ``IPv6`` by default when both
-the client and host support both. You can configure your MongoDB deployment
-to use ``IPv6`` mode when starting with ``mongod`` or ``mongos``. For more
-information about how to specify ``IPv6`` mode, see
+the client and host support both. If MongoDB uses one mode (such as IPv4) and your
+client uses a different mode (such as IPv6), this will result in the error message
+above.
```
ccho-mongodb, line 89: Suggestion:
The [style guide advises against using parentheses](https://www.mongodb.com/docs/meta/style-guide/style/punctuation/#parentheses) other than for a select few reasons.


```suggestion
the client and host support both. For example, if MongoDB uses IPv4 and your client uses a different mode, such as IPv6, the driver returns the prior error message.
```

source/connection-troubleshooting.txt
```diff
@@ -183,22 +174,27 @@ issues when attempting to connect to MongoDB using ``SCRAM-SHA-256``.
...
 
-   .. code-block:: none
+The following example shows how to percent encode "#MyP@assword?":
 
-      : / ? # [ ] @
+.. code-block:: javascript
+
+    const password = '#MyP@assword?';
+    const encodedPassword = encodeURIComponent(password);
+    console.log(encodedPassword);
```
ccho-mongodb, line 191: Issue:
The indentation looks off by one space.

Suggestion:

Fix the indentation. Applies to all instances.

The example also seems more verbose than necessary.

```javascript
  const encodedPassword = encodeURIComponent('#MyP@assword?');
  // The value of encodedPassword is "%23MyP%40assword%3F"
```

If you want to keep the split between code and output, this could be the code:

```javascript
  console.log(encodeURIComponent('#MyP@assword?'));
```

ccho-mongodb (approved):
LGTM, a couple suggestions. Please make sure the indentation is fixed and there are no build warnings or errors before merging.

PR #702, opened 2023-06-05
DOCSP-30358: Run a Command Fundamentals page

rustagir:
# Pull Request Info

JIRA - [DOCSP-30358](https://jira.mongodb.org/browse/DOCSP-30358) and [DOCSP-30119](https://jira.mongodb.org/browse/DOCSP-30119)
[Staging](https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-30358-runcommand-fundies/fundamentals/run-command/)

## Self-Review Checklist


source/fundamentals/run-command.txt
```diff
@@ -0,0 +1,165 @@
...
+
+To run a database command, you must specify the command and any relevant
+parameters in a command document, then pass the command document to a
+command execution method. The {+driver-short+} provides the following methods
+to run database commands:
+
+- ``command()``, which returns the command response as a
+  ``Document`` type. You can use this method with any database command.
+- ``runCursorCommand()``, which returns the command response as an iterable
+  ``RunCommandCursor`` type. You can use this method if your database command
```
mongoKart, line 32: S: "only if"?

```suggestion
  ``RunCommandCursor`` type. You can use this method only if your database command
```

source/fundamentals/run-command.txt
```diff
@@ -0,0 +1,165 @@
...
+   const result = await myDB.command({ hello: 1 });
+
+For a full list of database commands and corresponding parameters, see
+the :ref:`Additional Information section <addl-info-runcommand>`.
+
+.. note:: Read Preference
+
+   ``command()`` and ``runCursorCommand()`` do not obey the read
+   preference you may have set on your ``Db`` object elsewhere in
+   your code. You can set a read preference for command execution by
```
mongoKart, line 50: Q: What do they do instead?

source/fundamentals/run-command.txt
```diff
@@ -0,0 +1,165 @@
...
+   passing the relevant options object to either method:
```
mongoKart, line 51: Q: What makes the objects "relevant"? Does each method accept a different object format?

source/fundamentals/run-command.txt
```diff
@@ -0,0 +1,165 @@
...
+      const result = await myDB.command(commandDoc, commandOptions);
+   
+   For more information on read preference options, see :manual:`Read
+   Preference </core/read-preference/>` in the Server manual.
+
+Response
+--------
+
+Each method returns a ``Document`` object or a cursor that contains
+the response from the database after the command has been executed. Each
```
mongoKart, line 65: S: I'm guessing this means `command()` returns a `Document` and `runCursorCommand` returns a cursor. If so, can we clarify? By saying "Each method...", it sounds like each of the two methods will return one type or the other depending on some unspecified circumstance.

source/fundamentals/run-command.txt
```diff
@@ -0,0 +1,165 @@
...
+       - ``clusterTime``, which is the timestamp of the highest known cluster time for the member.
+       - ``signature``, which is a document that contains the hash of the cluster time and the ID
+         of the key used to sign the cluster time.
+
+Example
+-------
+
+The following code shows how you can use the ``runCursorCommand()`` method to
+run the ``checkMetadataConsistency`` command on the ``testDB`` database
+and iterates through the results:
```
mongoKart, line 110: I: "iterates", with an s, suggests it's the second thing "The following code" is doing, after "shows how you can use...".  To me, it's more natural that the code is showing how you can do two things: 1) use the command and 2) iterate through the results, in which case it should be "iterate", with no s

```suggestion
and iterate through the results:
```

source/fundamentals/run-command.txt
```diff
@@ -0,0 +1,165 @@
...
+       maxKeyObj: ...,
+       ...
+     }
+   }
+
+.. note::
+
+   If you store the command response in a cursor, you see only the
+   command result documents when you access the contents of the cursor. You won't
+   see the ``ok``, ``operationTime``, and ``$clusterTime`` fields.
```
mongoKart, line 149: Q: is there a way to see this info when using a cursor?

mongoKart (changes requested):
nicely done. mostly questions and suggestions

source/fundamentals/run-command.txt
rustagir: Yes, RunCommandOptions and RunCursorCommandOptions. I can clarify the pairs

source/fundamentals/run-command.txt
rustagir: I don't think so. I can try printing the cursor a few different ways and see what happens, though.

rustagir: after some testing it seems like you cant

source/fundamentals/run-command.txt
```diff
@@ -18,6 +18,13 @@ In this guide, you can learn how to run a database command with the
 administrative and diagnostic tasks, such as fetching server statistics,
 initializing a replica set, or running an aggregation pipeline.
 
+.. important:: Use Driver Methods if Possible
```
mongoKart, line 21:
```suggestion
.. important:: Prefer Driver Methods to Database Commands
```

mongoKart (approved):
couple suggestions but lgtm otherwise

PR #728, opened 2023-07-25
DOCSP-31596: improve slow operation faq

rustagir:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-31596>
Staging - https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-31596-improve-faq/faq/#how-can-i-prevent-a-slow-operation-from-delaying-other-operations-

## Self-Review Checklist

### DIFF
--- a/source/faq.txt
+++ b/source/faq.txt
@@ -190,7 +190,7 @@ will close the socket. We recommend that you select a value
 for ``socketTimeoutMS`` that is two to three times as long as the
 expected duration of the slowest operation that your application executes.
 
-How Can I Prevent Sockets From Timing out Before They Become Active?
+How Can I Prevent Sockets From Timing Out Before They Become Active?
 --------------------------------------------------------------------
 
 Having a large connection pool does not always reduce reconnection
@@ -267,20 +267,34 @@ are some things to check:
 How Can I Prevent a Slow Operation From Delaying Other Operations?
 ------------------------------------------------------------------
 
+You can prevent an operation from slowing down your application by
+specifying the size of the connection pool, which is the cache of
+connections that the driver maintains at any time. By increasing the
+maximum connection pool size, you can reduce application latency.
+
 To control the maximum size of a connection pool, you can set the
 ``maxPoolSize`` option in the :ref:`connection options
 <node-connection-options>`. The default value of ``maxPoolSize`` is
 ``100``. If the number of in-use connections to a server reaches
 ``maxPoolSize``, the next request to that server will wait 
-until a connection becomes available. To prevent long-running operations
-from slowing down your application, you can increase ``maxPoolSize``.
+until a connection becomes available. The following code sets
+``maxPoolSize`` to ``150`` when creating a new ``MongoClient``:
 
-The driver does not limit the number of requests that can wait for
-sockets to become available. Requests wait for the amount of time
-specified in the ``waitQueueTimeoutMS`` option, which
-defaults to ``0`` (no limit). You should set this option if it is
-more important to stop long-running operations than it is to complete
-every operation.
+.. code-block:: js
+   
+   const client = new MongoClient(uri, { maxPoolSize: 150 });
+
+You can also limit application latency by specifying the amount of time
+(in milliseconds) that requests wait for socket availability. The driver
+does not limit this wait time by default, but to specify a value, you
+can set the ``waitQueueTimeoutMS`` option. You should set this option if
+it is more important to stop long-running operations than it is to
+complete every operation. The following code sets ``waitQueueTimeoutMS``
+to ``100`` when creating a new ``MongoClient``:
+
+.. code-block:: js
+   
+   const client = new MongoClient(uri, { waitQueueTimeoutMS: 100 });
 
 .. tip::
    


source/faq.txt
```diff
@@ -267,25 +267,40 @@ are some things to check:
...
+maximum connection pool size, you can reduce application latency.
+
 To control the maximum size of a connection pool, you can set the
 ``maxPoolSize`` option in the :ref:`connection options
 <node-connection-options>`. The default value of ``maxPoolSize`` is
 ``100``. If the number of in-use connections to a server reaches
 ``maxPoolSize``, the next request to that server will wait 
-until a connection becomes available. To prevent long-running operations
-from slowing down your application, you can increase ``maxPoolSize``.
+until a connection becomes available. The following code sets
```
ccho-mongodb, line 280: Question:
From reading this FAQ entry (especially from the waitQueueTimeout section), I thought that the issue was that the client couldn't open enough sockets/connections to send commands, but in the wording of this sentence, it sounds like the server has to wait because there aren't enough connections/sockets to send the response of a command back to the client. Which one is correct or is it both?

Whatever the answer is, could this relationship be clarified in the introductory paragraph?

ccho-mongodb (changes requested):
From what I gathered, the major edit was adding an example which I think is helpful!
One of my comments on this PR includes a suggested structure that I think could be helpful to follow in order to reduce the potential assumptions a reader might need to make to understand the problem and solutions. Please feel free to reach out if you have any questions that might be more efficient to answer synchronously!

source/faq.txt
```diff
@@ -261,31 +261,53 @@ socket connections.
...
+modify your connection settings to increase the size of the connection
+pool, a group of connections to the server that the driver maintains at
+any time.
+
+To specify the maximum size of a
+connection pool, you can set the ``maxPoolSize`` option in the
+:ref:`connection options <node-connection-options>`. The default value
+of ``maxPoolSize`` is ``100``. If the number of in-use connections to a
+server reaches ``maxPoolSize``, the next operation sent to the server
+will wait until a connection becomes available. The following code sets
```
ccho-mongodb, line 300: Issue:

Regarding https://github.com/mongodb/docs-node/pull/728#discussion_r1275260135

I think the "operation waits" is not an accurate characterization of an operation. I think an operation should either be the driver sending the command that corresponds to the operation or the server processing/returning a response to an operation.

Suggestion:
If it happens to be the driver waiting to send it to the server, maybe this would describe it:
"the next operation the driver attempts to run on the server needs to wait for an available connection."

If it is the server waiting on the driver because the command has already been sent and executed:
"the operation sent to the server pauses until a free connection to the driver becomes available."

source/faq.txt
```diff
@@ -267,25 +267,40 @@ are some things to check:
...
-defaults to ``0`` (no limit). You should set this option if it is
-more important to stop long-running operations than it is to complete
-every operation.
+You can also limit application latency by specifying the amount of time
+(in milliseconds) that requests wait for socket availability. The driver
+does not limit this wait time by default, but to specify a value, you
+can set the ``waitQueueTimeoutMS`` option. You should set this option if
+it is more important to stop long-running operations than it is to
+complete every operation. The following code sets ``waitQueueTimeoutMS``
+to ``100`` when creating a new ``MongoClient``:
```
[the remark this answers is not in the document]
ccho-mongodb: Question: was this section intentionally removed? If so, what were the reasons for removing it?

source/faq.txt
```diff
@@ -261,46 +261,53 @@ socket connections.
...
+- Enable the database profiler on your deployment. To learn more, see
+  :manual:`Database Profiler </tutorial/manage-the-database-profiler/>`
+  in the Server manual.
+- Run the ``db.currentOp()`` :manual:`shell command </reference/method/>`. To learn more, see the 
+  :manual:`db.currentOp() </reference/method/db.currentOp/>`
+  documentation in the Server manual.
+- Enable connection pool monitoring. To learn more, see
+  :ref:`node-connection-pool-monitoring`.
+
+Once you determine how and why your operations are delayed, if possible,
```
ccho-mongodb, line 285: Suggestion:
Omit "how and why" to fit better in the suggested flow of this section.

source/faq.txt
```diff
@@ -261,46 +261,53 @@ socket connections.
...
-maximum connection pool size, you can reduce application latency.
-
-To control the maximum size of a connection pool, you can set the
-``maxPoolSize`` option in the :ref:`connection options
-<node-connection-options>`. The default value of ``maxPoolSize`` is
-``100``. If the number of in-use connections to a server reaches
-``maxPoolSize``, the next request to that server will wait 
-until a connection becomes available. The following code sets
+Before you make changes to your connection configuration or application,
+you should determine which operations could be causing delays. You can
```
ccho-mongodb, line 272: Suggestion:

I think this is a pretty good approach, but seems slightly out of order because it mentions the solution ("make changes to your connection configuration or application") before explaining how to determine whether the problem is applicable to the reader.

I took a look back at the original FAQ entry referenced in [DOCSP-29272](https://jira.mongodb.org/browse/DOCSP-29272) because it was a little unclear to me whether the "other operations" in the title was referring to MongoDB operations.
 
I think, unless you add it to the title of the FAQ entry, the first paragraph needs to clarify this.

I think that the delay this is referring to can only exist when you're sharing a single MongoClient (one connection pool) and making concurrent asynchronous calls.

Here's one way I can think of to present the info:

E.g.
"
When you run multiple MongoDB operations concurrently, a slow operation can cause delays to other operations. The slow operation keeps connections to MongoDB busy which can cause other operations to wait for a free connection.

If you suspect that slow MongoDB operations are delaying other ones, you can check each operation's performance using the following methods:

- Enable the database profiler...
- ...

"

source/faq.txt
rustagir: From the way it is described in the spec and the manual documentation, it seems like this is a Server-side limit. That is, the driver sends the request/executes the operation, but it waits on the server side until a connection is freed

source/faq.txt
```diff
@@ -267,25 +267,40 @@ are some things to check:
...
-sockets to become available. Requests wait for the amount of time
-specified in the ``waitQueueTimeoutMS`` option, which
-defaults to ``0`` (no limit). You should set this option if it is
-more important to stop long-running operations than it is to complete
-every operation.
+You can also limit application latency by specifying the amount of time
+(in milliseconds) that requests wait for socket availability. The driver
+does not limit this wait time by default, but to specify a value, you
+can set the ``waitQueueTimeoutMS`` option. You should set this option if
+it is more important to stop long-running operations than it is to
```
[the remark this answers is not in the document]
rustagir: @ccho-mongodb this is the response I left for why I removed the section on `waitQueueTimeoutMS` from the answer. In my research I did not find that this setting stops long running operations, but just clears out the queue when operations have been waiting for too long (as determined by `waitQueueTimeoutMS`)

source/faq.txt
```diff
@@ -261,31 +261,58 @@ socket connections.
...
-
-The driver does not limit the number of requests that can wait for
-sockets to become available. Requests wait for the amount of time
-specified in the ``waitQueueTimeoutMS`` option, which
-defaults to ``0`` (no limit). You should set this option if it is
-more important to stop long-running operations than it is to complete
-every operation.
+When you use the same ``MongoClient`` instance to run multiple MongoDB
+operations concurrently, a slow operation can cause delays to other
+operations. The slow operation keeps a connection to MongoDB occupied,
```
ccho-mongodb, line 273: Suggestion:
I think this should describe the general case since it's not referring to a specific one in the previous sentence.

```suggestion
operations. Slow operations keep a connection to MongoDB occupied,
```

ccho-mongodb (approved):
Great work! Thanks for all the changes; I think it's much easier to read and understand compared to the version at the start. Left one suggestion.

PR #729, opened 2023-08-03
docsp-31720 - remove deprecated helpers

mongoKart:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-31720
Staging:
- [What's New](https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/docsp-31720-remove-methods/whats-new/#what-s-new-in-6.0)
- [Upgrade](https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/docsp-31720-remove-methods/upgrade/#version-6.x-breaking-changes)

## Self-Review Checklist

### DIFF
--- a/source/whats-new.txt
+++ b/source/whats-new.txt
@@ -12,6 +12,8 @@ What's New
 
 Learn what's new in:
 
+* :ref:`Version 6.0 <version-6.0>`
+* :ref:`Version 5.7 <version-5.7>`
 * :ref:`Version 5.7 <version-5.7>`
 * :ref:`Version 5.6 <version-5.6>`
 * :ref:`Version 5.5 <version-5.5>`
@@ -38,6 +40,19 @@ Learn what's new in:
 * :ref:`Version 4.0 <version-4.0>`
 * :ref:`Version 3.7 <version-3.7>`
 * :ref:`Version 3.6 <version-3.6>`
+
+.. _version-6.0:
+
+What's New in 6.0
+-----------------
+
+The {+driver-short+} v6.0 release includes the following features:
+
+- This release removes support for the ``addUser()`` helper command. Use the
+  :manual:`createUser</reference/command/createUser>` command instead.
+- This release removes support for the ``collStats`` helper command. Use the
+  :manual:`$collStats </reference/operator/aggregation/collStats>` aggregation operator
+  instead. 
 
 .. _version-5.7:
 
--- a/source/whats-new.txt
+++ b/source/whats-new.txt
@@ -48,9 +48,9 @@ What's New in 6.0
 
 The {+driver-short+} v6.0 release includes the following features:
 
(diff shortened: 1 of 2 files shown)


source/whats-new.txt
```diff
@@ -39,6 +41,19 @@ Learn what's new in:
...
 
+.. _version-6.0:
+
+What's New in 6.0
+-----------------
+
+The {+driver-short+} v6.0 release includes the following features:
+
+- This release removes support for the ``addUser()`` helper command. Use the
+  :manual:`createUser</reference/command/createUser>` command instead.
```
ccho-mongodb, line 52: Suggestion:

I think it could be helpful to provide information that the command the user should use is not a Node.js one. 

```suggestion
  :manual:`createUser </reference/command/createUser>` MongoDB Shell command instead.
```

source/whats-new.txt
```diff
@@ -39,6 +41,19 @@ Learn what's new in:
...
 * :ref:`Version 3.6 <version-3.6>`
 
+.. _version-6.0:
+
+What's New in 6.0
+-----------------
+
+The {+driver-short+} v6.0 release includes the following features:
+
+- This release removes support for the ``addUser()`` helper command. Use the
```
ccho-mongodb, line 51: Suggestion:

Even though "helper command" might be a valid way to describe it, I think it might be easier to understand and not require additional explanation if called a "method". 
Also, I wonder whether "this release" might be repetitive since it's already introduced by the text preceding the list:

```suggestion
- Removes support for the ``addUser()`` method. Use the
```

This might apply to Breaking Changes update too.

source/whats-new.txt
```diff
@@ -39,6 +41,19 @@ Learn what's new in:
...
+  :manual:`createUser</reference/command/createUser>` command instead.
+- This release removes support for the ``collStats`` helper command. Use the
```
ccho-mongodb, line 53: Question:

I'm not sure what a "helper command" is. Would "operation" be a better description? I see that it's called "operation" in the [related PR](https://github.com/mongodb/node-mongodb-native/pull/3638).

I think it would probably be good to differentiate from addUser() since I think that's a method you can call from a [database](https://mongodb.github.io/node-mongodb-native/5.4/classes/Db.html#addUser) or [Admin](https://mongodb.github.io/node-mongodb-native/5.4/classes/Admin.html#addUser) instance.

I see that it's also mentioned in the admonition for 5.4, so if you change it, it would need to be updated there too.

ccho-mongodb (approved):
LGTM, a couple suggestions/question.

source/whats-new.txt
mongoKart: will change admonition in a separate ticket to avoid backporting issues

PR #737, opened 2023-08-10
DOCSP-31487: removed options for command() v6

rustagir:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-31487>
Staging
- [what's new](https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-31487-command-options-changes/whats-new/#what-s-new-in-6.0)
- [upgrade guide](https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-31487-command-options-changes/upgrade/#version-6.x-breaking-changes)
- [run a command fundamentals](https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-31487-command-options-changes/fundamentals/run-command/#command-options)
- [run a command usage ex](https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-31487-command-options-changes/usage-examples/command/)

## Self-Review Checklist


source/fundamentals/run-command.txt
```diff
@@ -55,17 +55,46 @@ the current member's role in the replica set, on a database:
...
 
+.. _node-command-options:
+
+Command Options
+---------------
+
+You can set options when executing a database command by
+passing an options object to either command method. The ``command()`` method
+takes a ``RunCommandOptions`` object, and the ``runCursorCommand()`` method
+takes a ``RunCursorCommandOptions`` object.
```
ccho-mongodb, line 66: Suggestion:

1) I think it would be more clear to specify the command methods rather than expecting the reader to associate it with content from the middle of the previous section.

2) I think "set option ... by passing an options object" is redundant.

3) I think the association between "options" object and the two types of options objects could be more clear.


```suggestion
You can specify optional command behavior for either the ``command()`` or the ``runCursorCommand()`` method.

The ``command()`` method accepts a RunCommandOptions object. To learn more about the RunCommandOptions type, see the API documentation.

The ``runCursor() method`` accepts a ``runCursorCommand()`` object. To learn more about the RunCursorCommandOptions type, see the API documentation.
```

source/fundamentals/run-command.txt
```diff
@@ -55,17 +55,46 @@ the current member's role in the replica set, on a database:
...
+following options directly to the ``command()`` method:
+
+- ``comment``
+- ``enableUtf8Validation``
+- ``raw``
+- ``readPreference``
+- ``session``
+
+You can set command-specific options, including a write or read
+concern, in the command document that you pass to ``command()``. The
```
ccho-mongodb, line 78: Suggestion:

1) I think the reader would benefit from a clear differentiation between the command options described in the first paragraph and "command-specific options".

I also think that it might not be clear which options can be specified in the command document, and whether it depends on the command.

To contrast this with the "only the following options" from the prior paragraph:
"You can specify additional commands on certain commands..." 


2) I think the mention of write or read concerns as options that you can specify in the command document, followed by an example of maxTimeMS is confusing. To resolve this, provide the reader with a reference that they can check to determine which options they can provide which commands.

source/fundamentals/run-command.txt
```diff
@@ -55,17 +55,46 @@ the current member's role in the replica set, on a database:
...
+Command Options
+---------------
+
+You can set options when executing a database command by
+passing an options object to either command method. The ``command()`` method
+takes a ``RunCommandOptions`` object, and the ``runCursorCommand()`` method
+takes a ``RunCursorCommandOptions`` object.
+
+Starting in version 6.0 of the {+driver-short+}, you can pass only the
+following options directly to the ``command()`` method:
```
ccho-mongodb, line 69: Question: Does this apply to the runCursorCommand() method or only command() method? In either case, I think it could be good to call out the limitations for both methods.

source/fundamentals/run-command.txt
```diff
@@ -55,17 +55,46 @@ the current member's role in the replica set, on a database:
...
+     find: "values",
+     filter: { x: "abc" },
+     maxTimeMS: 2000,
+   };
+   const result = await myDB.command(commandDoc);
+
 .. note:: Read Preference
 
-   ``command()`` and ``runCursorCommand()`` do not obey the read
+   The ``command()`` and ``runCursorCommand()`` methods do not obey the read
```
ccho-mongodb, line 94: Suggestion:

I think it would be good to avoid humanizing inanimate objects when possible (Use effective verbs from the style guide).

"The ... methods ignore" or "The ... methods throw an error if..."

Also, in the next sentence, I think "earlier" is more correct than "elsewhere".

ccho-mongodb (changes requested):
Looks pretty good! I identified couple issues and suggestions. Would like to re-review once addressed.

source/upgrade.txt
```diff
@@ -70,6 +70,9 @@ Version 6.x Breaking Changes
   ``tlsCertificateKeyFile`` connection options when you call the
   ``MongoClient.connect()`` method, not when you create the
   ``MongoClient`` instance.
+- The ``Db.command()`` method does not take any command-specific
+  options. To learn more about the supported options, see the
```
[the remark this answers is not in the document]
rustagir: I assume that since the options are being **removed**, whereas they were previously just deprecated, the driver will throw an error.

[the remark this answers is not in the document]
ccho-mongodb: I think "take" is an issue related to [Don't Humanize Inanimate Objects](https://www.mongodb.com/docs/meta/style-guide/writing/use-effective-verbs/#don-t-humanize-inanimate-objects) (the style guide examples aren't the clearest) because it implies performing an action that is usually associated to an intention.

I think "accept" would be better. I think "command-specific options" also could be substituted for a more definitive description.

Suggestion:

"The Db.command() method only accepts [describe the options it accepts] options. To learn more about these options, see ..."

source/fundamentals/run-command.txt
rustagir: the ticket only says that it is for the `command()` method and from the feature PR, it seems  that `runCursorCommand()` is unchanged.

source/upgrade.txt
```diff
@@ -70,6 +70,10 @@ Version 6.x Breaking Changes
   ``tlsCertificateKeyFile`` connection options when you call the
   ``MongoClient.connect()`` method, not when you create the
   ``MongoClient`` instance.
+- The ``Db.command()`` method only accepts options that are not related
```
ccho-mongodb, line 73: Issue:

Perhaps my fault for providing this in the suggestion, but I think "only" is a misplaced modifier in this case.
From the [style guide on "only"](https://www.mongodb.com/docs/meta/style-guide/terminology/alphabetical-terms/#o):
> Place only preceding the word or group that it modifies. If only modifies a final word or phrase, it may precede or follow that word or phrase.

Suggestion:

I think In this case, "only" modifies options rather than "accepts".

```suggestion
- The ``Db.command()`` method accepts only options that are not related
```

source/fundamentals/run-command.txt
```diff
@@ -55,17 +55,55 @@ the current member's role in the replica set, on a database:
...
+- ``comment``
+- ``enableUtf8Validation``
+- ``raw``
+- ``readPreference``
+- ``session``
+
+You can set additional options, including a write or read
+concern, in the command document that you pass to
+the ``command()`` method. See the Server manual :manual:`entry
+</reference/command/>` for a command to learn about its associated
```
ccho-mongodb, line 85: Issues:

1) I think this could be confusing to a user since the page this links to shows the overview of runCommand() and adminCommand() rather than command() and "its associated options". I also wasn't able to locate the aforementioned write or read concern options.

2) The text of the link should be descriptive. See [Link and Cross-Reference Examples](https://www.mongodb.com/docs/meta/style-guide/style/links-and-cross-references/#link-and-cross-reference-examples).

Suggestion:

From looking at docs/manual/reference/command, it looks like the options are listed only on the target pages of the links of each command. It could be helpful to inform the reader that is how it is structured to reduce the chance of confusion. Also, I think it would be good to continue using the format "\<Page title\>" in the Server Manual.

E.g.
"Learn more about the available options for each command by following the link associated with the command in the :manual:\`User Commands </reference/command/#user-commands>\` section of the Server Manual"

source/fundamentals/run-command.txt
```diff
@@ -55,17 +55,55 @@ the current member's role in the replica set, on a database:
...
+       writeConcern: { w: "majority", wtimeout: 5000 }
+   };
+   const result = await myDB.command(commandDoc);
+
 .. note:: Read Preference
 
-   ``command()`` and ``runCursorCommand()`` do not obey the read
-   preference you may have set on your ``Db`` object elsewhere in
+   The ``command()`` and ``runCursorCommand()`` methods do not execute with
+   the read preference you may have set on your ``Db`` object earlier in
```
ccho-mongodb, line 104: Suggestion:

I think it would be easier to understand and more concise by using [Positive Statements](https://www.mongodb.com/docs/meta/style-guide/writing/use-positive-statements/):

```suggestion
   The ``command()`` and ``runCursorCommand()`` methods ignore the
   read preference setting you may have set on your ``Db`` object.
```

I don't think it's necessary to mention "earlier in your code" since "you may have set" refers to the recent past.

source/usage-examples/command.txt
```diff
@@ -4,20 +4,20 @@
...
-    the {+driver-short+} whenever possible.
-
-You can specify additional options in the ``options`` object passed in
-the second parameter of the ``command()`` method. For more information
-on the options you can pass, see the
-`db.command() API documentation <{+api+}/classes/Db.html#command>`__.
+You can run database operations by using the
+`Db.command() <{+api+}/classes/Db.html#command>`__ method. Call the
+``command()`` method with your command document on an instance of a
+database.
```
ccho-mongodb, line 10: Suggestion:

I think the second sentence attempts to fit too much information. I also think this could be more clear if broken down into communicating the following info in order:

1) The command() method allows you to run any MongoDB database operation. This is an instance method of the Db class.

2) To specify a command, create a command document that follows the command syntax (https://www.mongodb.com/docs/manual/reference/command/#command-syntax).

3) To run the command on the database, pass the command document to the command() instance method.

Let me know if you need any help crafting this description from this breakdown.

ccho-mongodb (changes requested):
I think this is nearly there, but have a few suggestions and would like to review a couple changes.

source/fundamentals/run-command.txt
rustagir: since the page describes commands outside of the User Commands category (such as `hello`), I will change the link and link text slightly

source/usage-examples/command.txt
rustagir: Since this is a usage example, I dont think we should describe the command syntax. I think it is sufficient to mention that users need to create a command document and then link out to the manual

source/fundamentals/run-command.txt
```diff
@@ -55,17 +55,58 @@ the current member's role in the replica set, on a database:
...
+
+You can set additional options, including a write or read
+concern, in the command document that you pass to
+the ``command()`` method. To learn more about the available options for
+a command, select it from the list of commands on the :manual:`Database
+Commands </reference/command/>` page of the Server manual.
+
+See the Server manual :manual:`entry
+</reference/command/>` for a command to learn about its associated
+options.
```
ccho-mongodb, line 90: Question:
Did you mean to include this? If so, it repeats same the in the previous paragraph (style guide issue and confusing user experience) and should be removed.

source/usage-examples/command.txt
ccho-mongodb: I think we're in agreement. I meant was that this should describe the method syntax (which it currently does, but could be more clear), not the command syntax. The important part is to mention that it is a document that contains a command because that's the required parameter.

source/fundamentals/run-command.txt
ccho-mongodb: Resolving and adding a comment to the revised version.

ccho-mongodb (changes requested):
Looks good, only a couple items left to address.

ccho-mongodb (approved):
Sorry, I just noticed an issue about the example. Could you address that before merging? Otherwise, I provided a couple suggestions and think this shouldn't require another review, but happy to take another look if you want.

PR #739, opened 2023-08-11
DOCSP-31924 mongodb-client-encryption changes for 6.0

jordan-smith721:
# Pull Request Info

JIRA - https://jira.mongodb.org/browse/DOCSP-31924
Staging -
[What's New ](https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-31924-move-clientEncryption/whats-new/#std-label-version-6.0)
[Upgrade](https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-31924-move-clientEncryption/upgrade/)

## Self-Review Checklist

### DIFF
--- a/source/upgrade.txt
+++ b/source/upgrade.txt
@@ -71,7 +71,7 @@ Version 6.x Breaking Changes
   ``MongoClient.connect()`` method, not when you create the
   ``MongoClient`` instance.
 - If you add ``mongodb-client-encryption`` as a dependency,
-  the major version number must match that of the {+driver-short}. For example,
+  the major version number must match that of the {+driver-short+}. For example,
   {+driver-short+} v6.x.x requires ``mongodb-client-encryption`` v6.x.x.
 - Automatic Encryption methods are now in the {+driver-short+}. You must
   import these methods from the driver instead of from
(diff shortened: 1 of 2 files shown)


source/upgrade.txt
```diff
@@ -70,6 +70,12 @@ Version 6.x Breaking Changes
   ``tlsCertificateKeyFile`` connection options when you call the
   ``MongoClient.connect()`` method, not when you create the
   ``MongoClient`` instance.
+- The major version number of ``mongodb-client-encryption`` must now be the
+  same as the major version number of the {+driver-short+}. For example,
```
ccho-mongodb, line 74: Suggestion:

I think this could be more succinct and could emphasize the condition under which they need to match.

```suggestion
- If you add ``mongodb-client-encryption`` as a dependency,
   the major version number must match that of the {+driver-short}.
   For example,
```

source/upgrade.txt
```diff
@@ -70,6 +70,12 @@ Version 6.x Breaking Changes
...
+  {+driver-short+} v6.x.x requires ``mongodb-client-encryption`` v6.x.x.
+- Automatic Encryption methods are now in the {+driver-short+}. You must now
+  import these methods from the driver rather than from
```
ccho-mongodb, line 77: Suggestion:

I think the first sentence establishes that there's a change, so I would omit the second "now". 

 think "rather than" implies something is preferred less which conflicts with "must".

```suggestion
- Automatic Encryption methods are now in the {+driver-short+}. You must
  import these methods from the driver instead of from
```

ccho-mongodb (approved):
LGTM, left a couple suggestions. The same suggestions apply to instances in both files.

PR #741, opened 2023-08-16
Various v6 breaking changes

jordan-smith721:
# Pull Request Info

JIRA - 
https://jira.mongodb.org/browse/DOCSP-31649
https://jira.mongodb.org/browse/DOCSP-31740
https://jira.mongodb.org/browse/DOCSP-32050
https://jira.mongodb.org/browse/DOCSP-31739
https://jira.mongodb.org/browse/DOCSP-31645

Staging - 
[Upgrade Driver Versions](https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-31739-remove-strings-objectid/upgrade/#version-6.x-breaking-changes)
[What's New](https://docs-mongodbcom-staging.corp.mongodb.com/node/docsworker-xlarge/DOCSP-31739-remove-strings-objectid/whats-new/#std-label-version-6.0)

## Self-Review Checklist

### DIFF
--- a/source/upgrade.txt
+++ b/source/upgrade.txt
@@ -77,12 +77,12 @@ Version 6.x Breaking Changes
   import these methods from the driver instead of from
   ``mongodb-client-encryption``.
 - Removed the ``ObjectId`` constructor that accepted a 12-character string.
-- The ``abortTransaction()`` and ``commitTransaction()`` methods return
+- Modified ``abortTransaction()`` and ``commitTransaction()`` methods to return
   ``null`` instead of the raw command results.
 - Removed connection option helpers that accepted values other than ``true``
   or  ``false`` as booleans. You must provide either ``true`` or ``false`` values in
   the connection string or to the MongoClient constructor.
-- Removed the Binary BSON type constructor that accepted a string.
+- Removed the ``Binary`` BSON type constructor that accepted a string.
 - The ``Binary.write()`` method no longer accepts a string to write to the binary
   BSON object.
 - The ClientEncryption API returns promises instead of callbacks.
--- a/source/upgrade.txt
+++ b/source/upgrade.txt
@@ -76,18 +76,16 @@ Version 6.x Breaking Changes
 - Automatic Encryption methods are now in the {+driver-short+}. You must
   import these methods from the driver instead of from
   ``mongodb-client-encryption``.
-- Removed support for strings with a length of 12 from the ``ObjectId``
-  constructor.
-- The ``abortTransaction`` and ``commitTransaction`` methods return
+- Removed the ``ObjectId`` constructor that accepted a 12-character string.
+- The ``abortTransaction()`` and ``commitTransaction()`` methods return
   ``null`` instead of the raw command results.
-- Removed connection option helpers that recognized values other than ``true``
-  or  ``false`` as booleans. You must provide either ``true`` or ``false`` values for
-  options that expect boolean values in the connection string or MongoClient
-  constructor.
+- Removed connection option helpers that accepted values other than ``true``
+  or  ``false`` as booleans. You must provide either ``true`` or ``false`` values in
+  the connection string or to the MongoClient constructor.
 - Removed the Binary BSON type constructor that accepted a string.
-- The ``Binary.write`` method no longer accepts a string to write to the binary
+- The ``Binary.write()`` method no longer accepts a string to write to the binary
   BSON object.
-- The ``ClientEncryption`` API returns promises instead of callbacks.
+- The ClientEncryption API returns promises instead of callbacks.
 
 .. _node-breaking-changes-v5.x:
 
--- a/source/upgrade.txt
+++ b/source/upgrade.txt
@@ -76,16 +76,18 @@ Version 6.x Breaking Changes
 - Automatic Encryption methods are now in the {+driver-short+}. You must
   import these methods from the driver instead of from
   ``mongodb-client-encryption``.
-- Removes support for strings with a length of 12 from the ``ObjectId``
+- Removed support for strings with a length of 12 from the ``ObjectId``
   constructor.
-- The ``abortTransaction`` and ``commitTransaction`` methods now return
+- The ``abortTransaction`` and ``commitTransaction`` methods return
   ``null`` instead of the raw command results.
-- Connection strings and the ``MongoClient`` constructor now accept only
-  the boolean values ``true`` and ``false``. Other values such as ``1``, or ``"yes"``
-  are no longer valid.
-- The ``Binary`` BSON type no longer accepts a string as a constructor argument.
-- The ``Binary.write`` method  no longer accepts a string as an argument.
-- ``ClientEncryption`` APIs now return promises instead of callbacks.
+- Removed connection option helpers that recognized values other than ``true``
+  or  ``false`` as booleans. You must provide either ``true`` or ``false`` values for
+  options that expect boolean values in the connection string or MongoClient
+  constructor.
+- Removed the Binary BSON type constructor that accepted a string.
+- The ``Binary.write`` method no longer accepts a string to write to the binary
+  BSON object.
+- The ``ClientEncryption`` API returns promises instead of callbacks.
 
 .. _node-breaking-changes-v5.x:
 
--- a/source/upgrade.txt
+++ b/source/upgrade.txt
@@ -76,6 +76,16 @@ Version 6.x Breaking Changes
 - Automatic Encryption methods are now in the {+driver-short+}. You must
   import these methods from the driver instead of from
   ``mongodb-client-encryption``.
+- Removes support for strings with a length of 12 from the ``ObjectId``
+  constructor.
+- The ``abortTransaction`` and ``commitTransaction`` methods now return
+  ``null`` instead of the raw command results.
+- Connection strings and the ``MongoClient`` constructor now accept only
+  the boolean values ``true`` and ``false``. Other values such as ``1``, or ``"yes"``
+  are no longer valid.
+- The ``Binary`` BSON type no longer accepts a string as a constructor argument.
+- The ``Binary.write`` method  no longer accepts a string as an argument.
+- ``ClientEncryption`` APIs now return promises instead of callbacks.
 
 .. _node-breaking-changes-v5.x:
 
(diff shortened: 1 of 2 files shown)


source/upgrade.txt
```diff
@@ -76,6 +76,16 @@ Version 6.x Breaking Changes
...
   import these methods from the driver instead of from
   ``mongodb-client-encryption``.
+- Removes support for strings with a length of 12 from the ``ObjectId``
+  constructor.
+- The ``abortTransaction`` and ``commitTransaction`` methods now return
+  ``null`` instead of the raw command results.
+- Connection strings and the ``MongoClient`` constructor now accept only
+  the boolean values ``true`` and ``false``. Other values such as ``1``, or ``"yes"``
+  are no longer valid.
+- The ``Binary`` BSON type no longer accepts a string as a constructor argument.
```
ccho-mongodb, line 86: Suggestion:

I think the related items should be closer together. For this item, the constructors for the Binary BSON type has changed. The change is that there's no longer a constructor for it that accepts strings.

To keep Binary BSON type and constructor together, you could reorder the sentence to something like this:
"Removed the Binary BSON type constructor that accepted a string."

source/upgrade.txt
```diff
@@ -76,6 +76,16 @@ Version 6.x Breaking Changes
...
+- The ``Binary.write`` method  no longer accepts a string as an argument.
```
ccho-mongodb, line 87: Question:
What did the string argument do? I think a user might want to know this as well to understand the change.

source/upgrade.txt
```diff
@@ -76,6 +76,16 @@ Version 6.x Breaking Changes
...
+- ``ClientEncryption`` APIs now return promises instead of callbacks.
```
ccho-mongodb, line 88: Question: 
Are there multiple API's? I think it's just a single one. If there's some distinction between the private and public API, I think that should be surfaced to the user. It's also likely that a private API didn't cause a breaking change to the reader's code since they're not supposed to be using that anyway.

ccho-mongodb (changes requested):
Good start! I left some comments/questions that might apply to both sections.

While I see a precedent of duplicating breaking changes in What's New, I think it's redundant, especially when the list of items is long. I also think there could be issues that stem from using "<major version>.x" for organizing breaking changes, since some of them could potentially be in minor version updates.
 
These structural suggestions are probably not in scope for this PR, but here's what I think we should ideally have prior to the v6 release:

- The "What's New" page mentions that there are breaking changes in that version, using "warning" since those can potentially disrupt operations if they upgrade, and links to the "Upgrade" page.
- The Upgrade Page should show specific versions in which the breaking change was introduced -- in this case, v6.0.0.

source/upgrade.txt
jordan-smith721: The ticket said "APIs" but looking into it more I think that wording was from the overall ticket referring to the different driver's implementations of the one API. Changed it to just be the ClientEncryption API

source/whats-new.txt
```diff
@@ -58,6 +58,24 @@ What's New in 6.0
...
+     or  ``false`` as booleans. You must provide either ``true`` or ``false`` values for
+     options that expect boolean values in the connection string or MongoClient
+     constructor.
+
+   - Removed the Binary BSON type constructor that accepted a string.
+
+   - The ``Binary.write`` method no longer accepts a string to write
+     to the binary BSON object.
+
+   - ``ClientEncryption`` APIs return promises instead of callbacks.
```
ccho-mongodb, line 77: Suggestion:

I think this should be changed to be consistent: https://github.com/mongodb/docs-node/pull/741#discussion_r1296520856

source/upgrade.txt
```diff
@@ -76,6 +76,18 @@ Version 6.x Breaking Changes
...
+- The ``abortTransaction`` and ``commitTransaction`` methods return
+  ``null`` instead of the raw command results.
+- Removed connection option helpers that recognized values other than ``true``
+  or  ``false`` as booleans. You must provide either ``true`` or ``false`` values for
+  options that expect boolean values in the connection string or MongoClient
+  constructor.
+- Removed the Binary BSON type constructor that accepted a string.
+- The ``Binary.write`` method no longer accepts a string to write to the binary
+  BSON object.
+- The ``ClientEncryption`` API returns promises instead of callbacks.
```
ccho-mongodb, line 90: Suggestion:

I think we use monospace formatting for [a number of types of items according to the style guide](https://www.mongodb.com/docs/meta/style-guide/style/text-formatting/), but have not seen it used for names of APIs. One example of an API we refer to often is "Stable API" which does not use monospace.

source/whats-new.txt
```diff
@@ -58,6 +58,24 @@ What's New in 6.0
      import these methods from the driver instead of from
      ``mongodb-client-encryption``.
 
+   - Removed support for strings with a length of 12 from the ``ObjectId``
+     constructor.
+
+   - The ``abortTransaction`` and ``commitTransaction`` methods return
```
ccho-mongodb, line 64: Suggestion:
Some entries use the "()" suffix to identify methods names. I think it helps make certain less ambiguous and would make sense to make these consistent.

source/whats-new.txt
```diff
@@ -58,6 +58,24 @@ What's New in 6.0
...
+     ``null`` instead of the raw command results.
+
+   - Removed connection option helpers that recognized values other than ``true``
```
ccho-mongodb, line 67: Suggestion:

While this style guide guidance is sometimes borderline, it mentions [don't humanize inanimate objects](https://www.mongodb.com/docs/meta/style-guide/writing/use-effective-verbs/#don-t-humanize-inanimate-objects). I think it is not common to describe methods as performing the "recognize" action:

```suggestion
   - Removed connection option helpers that accepted values other than ``true``
```

Other alternatives could be "parsed" or "casted"

source/whats-new.txt
```diff
@@ -58,6 +58,24 @@ What's New in 6.0
...
+     or  ``false`` as booleans. You must provide either ``true`` or ``false`` values for
+     options that expect boolean values in the connection string or MongoClient
+     constructor.
```
ccho-mongodb, line 70: Suggestion:

I think this might be a more direct way to explain the new requirement:

```suggestion
     or  ``false`` as booleans. You must provide either ``true`` or ``false`` values in
     the connection string or to the MongoClient constructor.
```

ccho-mongodb (approved):
Looks mostly good -- just a few suggestions. Make sure to update all the locations as I didn't want to duplicate comments.

Did you get a chance to read [this part of the previous review](https://github.com/mongodb/docs-node/pull/741#pullrequestreview-1581534558)? Could you communicate how you want to handle this when you get the chance?

source/whats-new.txt
jordan-smith721: Oops, looks like I pushed without saving that change. Fixed

source/upgrade.txt
jordan-smith721: Changed it to be non-monospaced.

source/whats-new.txt
```diff
@@ -58,6 +58,24 @@ What's New in 6.0
...
+
+   - Removed the Binary BSON type constructor that accepted a string.
+
+   - The ``Binary.write`` method no longer accepts a string to write
+     to the binary BSON object.
```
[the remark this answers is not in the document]
jordan-smith721: I'm not sure if the suggested change is correct since they didn't remove the method that accepted the string. They just removed string as one of the accepted parameters to that method

jordan-smith721:
> While I see a precedent of duplicating breaking changes in What's New, I think it's redundant, especially when the list of items is long. I also think there could be issues that stem from using ".x" for organizing breaking changes, since some of them could potentially be in minor version updates.
> 
> These structural suggestions are probably not in scope for this PR, but here's what I think we should ideally have prior to the v6 release:
> 
> * The "What's New" page mentions that there are breaking changes in that version, using "warning" since those can potentially disrupt operations if they upgrade, and links to the "Upgrade" page.
> * The Upgrade Page should show specific versions in which the breaking change was introduced -- in this case, v6.0.0.

I like the idea of just mentioning that there is a breaking change in "What's New" rather than duplicating the information. I'll make a ticket to make that change and sync with Rea on it since she's the other steward for Node. I can link the ticket here when it's created.
I'll create a ticket to update the Upgrade page as well with specific versions

source/upgrade.txt
```diff
@@ -76,6 +76,16 @@ Version 6.x Breaking Changes
 - Automatic Encryption methods are now in the {+driver-short+}. You must
   import these methods from the driver instead of from
   ``mongodb-client-encryption``.
+- Removed the ``ObjectId`` constructor that accepted a 12-character string.
+- The ``abortTransaction()`` and ``commitTransaction()`` methods return
+  ``null`` instead of the raw command results.
+- Removed connection option helpers that accepted values other than ``true``
+  or  ``false`` as booleans. You must provide either ``true`` or ``false`` values in
+  the connection string or to the MongoClient constructor.
+- Removed the Binary BSON type constructor that accepted a string.
```
ccho-mongodb, line 85: Suggestion:
I think the monospace formatting for "Binary" should still apply since it is a type/class name and helps the reader understand which type (in case there are multiple binary types).

source/upgrade.txt
```diff
@@ -76,6 +76,16 @@ Version 6.x Breaking Changes
...
+- The ``Binary.write()`` method no longer accepts a string to write to the binary
```
ccho-mongodb, line 86: Question:
Is this similar to the "Binary BSON type"? If so, I think it could be good to keep consistent formatting. If it is not related, I think further detail on how it is different could be helpful.

source/upgrade.txt
```diff
@@ -76,6 +76,16 @@ Version 6.x Breaking Changes
 - Automatic Encryption methods are now in the {+driver-short+}. You must
   import these methods from the driver instead of from
   ``mongodb-client-encryption``.
+- Removed the ``ObjectId`` constructor that accepted a 12-character string.
+- The ``abortTransaction()`` and ``commitTransaction()`` methods return
```
ccho-mongodb, line 80: Suggestion:

I think it could be helpful to match the format of the other list items by starting it with a verb.

```suggestion
   - Modified ``abortTransaction()`` and ``commitTransaction()`` methods to return
```

ccho-mongodb (approved):
LGTM, just a couple suggestions and a question.

jordan-smith721:
@ccho-mongodb here are the tickets to address your suggestions outside of this PR:
https://jira.mongodb.org/browse/DOCSP-32548
https://jira.mongodb.org/browse/DOCSP-32561

PR #751, opened 2023-08-25
DOCSP-31789: transaction return type

rustagir:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-31789>
Staging:
- [transactions fundamentals page](https://preview-mongodbrustagir.gatsbyjs.io/node/DOCSP-31789-transaction-return-type/fundamentals/transactions/)
- [core api example](https://preview-mongodbrustagir.gatsbyjs.io/node/DOCSP-31789-transaction-return-type/usage-examples/transaction-core/)
- [conv txn api example](https://preview-mongodbrustagir.gatsbyjs.io/node/DOCSP-31789-transaction-return-type/usage-examples/transaction-conv/)
- [whats new](https://preview-mongodbrustagir.gatsbyjs.io/node/DOCSP-31789-transaction-return-type/whats-new/#what-s-new-in-6.0)
- [upgrade guide](https://preview-mongodbrustagir.gatsbyjs.io/node/DOCSP-31789-transaction-return-type/upgrade/#version-6.x-breaking-changes)

## Self-Review Checklist


source/usage-examples/convenient-transactions.txt
```diff
@@ -0,0 +1,162 @@
...
+   :manual:`UnknownTransactionCommitResult description
+   </core/transactions-in-applications/#unknowntransactioncommitresult>` on
+   the Driver API page in the Server manual.
+
+The {+driver-short+} provides the following helper methods to perform a
+transaction by using the Convenient Transaction API:
+
+- ``withSession()``: Runs a given operation within an implicitly created
+  session. You do not have to handle the creation or ending of the
+  session.
```
nbbeeken, line 35:
```suggestion
- ``withSession()``: Runs a given function passing it a started session. Regardless of the function's outcome, the session will automatically be ended and related resources cleaned up.
```

An "implicit" session is specific terminology that refers to sessions created by the driver for every operation if a user didn't provide one. In this case, the session is still "explicit" as the user still has to pass it to each operation.

source/usage-examples/convenient-transactions.txt
```diff
@@ -0,0 +1,162 @@
...
+- The code returns a message acknowledging that the transaction
+  committed successfully with a copy of the purchase record.
+
+4. Prints the return type of ``withSession()``, which is either the
+   error message or the acknowledgment that the transaction completed.
+
+.. note::
+   
+   You must pass the session instance to each operation that
+   you want to run in that session.
```
nbbeeken, line 110: Probably good adds here:
- You must also `await` every operation 
  - we've seen this missed in JIRA tickets in the past
- You must NOT use parallelism (`Promise.all`)
  - Sessions cannot be used in parallel, but also are unable to report an error if they are used this way. It will likely lead to a server error, but inconsistently depending on timing.

source/code-snippets/transactions/txn-convenient.js
```diff
@@ -0,0 +1,81 @@
...
+// start-order-fail
+const order2 = [
+  { item: "volleyball", qty: 1 }
+];
+// end-order-fail
+
+// start-transaction
+const txnResult = await client.withSession(async (session) =>
+  session
+    .withTransaction(async () => {
```
nbbeeken, line 34:
```suggestion
    .withTransaction(async (session) => {
```

Minor, I think there's actually no functional difference but technically the API is that you should use the session passed to the transaction function, and shadowing variables like this is a perfectly fine way to make sure you don't grab something of the same name from the outer scope.

source/code-snippets/transactions/txn-convenient.js
```diff
@@ -0,0 +1,83 @@
...
+      const invColl = client.db("testdb").collection("inventory");
+      const recColl = client.db("testdb").collection("records");
+
+      let total = 0;
+      for (let i = 0; i < order.length; i++) {
```
nbbeeken, line 39:
```suggestion
      for (const item of orders) {
```

`order` -> `orders`, but also I'm not seeing it defined, should it be `order1` and `order2`?

rustagir: This code is not necessarily meant to be runnable, but i can add the `orders` variable that the user can set to `order1` or `order2`

source/fundamentals/transactions.txt
```diff
@@ -163,86 +173,88 @@ order details. Suppose you organize the data updates as follows:
...
+perform the multi-document transaction in a session. This example
+shows how to perform the following actions:
 
 1. Start a session
-#. Start a transaction, specifying transaction options
-#. Perform data operations in the same session
-#. Commit a transaction, or cancel it if the driver encounters an error
-#. End a session
+#. Start a transaction with transaction options
+#. Perform data operations in the session
```
ccho-mongodb, line 255:
```suggestion
#. Perform read and write operations in the session
```

source/fundamentals/transactions.txt
```diff
@@ -163,86 +173,88 @@ order details. Suppose you organize the data updates as follows:
...
+#. Commit the transaction, or abort it if the driver encounters an error
```
ccho-mongodb, line 256: Issue:

[Style guide ](https://www.mongodb.com/docs/meta/style-guide/terminology/alphabetical-terms/) on "abort":

```suggestion
#. Commit the transaction, or end it if the driver encounters an error
```

source/usage-examples.txt
```diff
@@ -120,3 +120,4 @@ Available Usage Examples
 - :doc:`Run a Command </usage-examples/command>`
 - :doc:`Watch for Changes </usage-examples/changeStream>`
 - :doc:`Perform Bulk Operations </usage-examples/bulkWrite>`
+- :doc:`Perform a Convenient Transaction </usage-examples/convenient-transactions>`
```
ccho-mongodb, line 123: Question:
I think the name of the API is "Convenient Transaction API", but does that mean that using it to complete a transaction is a "convenient transaction"? If not, maybe this could be "Use the Convenient Transaction API" instead.

ccho-mongodb (changes requested):
Looks pretty good so far. I will take a look at the Convenient Transaction API content in a followup review.

source/usage-examples/convenient-transactions.txt
```diff
@@ -0,0 +1,174 @@
+.. _node-usage-convenient-txn:
+
+================================
+Perform a Convenient Transaction
+================================
+
+The {+driver-short+} provides two APIs to perform transactions, the Core
+API and the **Convenient Transaction API**. By using the Convenient Transaction
```
ccho-mongodb, line 8: Similar comment here: https://github.com/mongodb/docs-node/pull/751/files#r1307545529

source/fundamentals/transactions.txt
ccho-mongodb: Regardless of the nomenclature of `abortTransaction`, I think in all instances that describe the action it takes, we should use "ends" per the style guide.

source/fundamentals/transactions.txt
```diff
@@ -63,40 +63,45 @@ MongoDB in the following sections of this guide:
 Transaction APIs
 ----------------
 
-Use the **Core API** to implement a transaction with the driver. To use the Core
-API, you declare the start and commit points of the transaction.
+The driver provides multiple APIs for performing transactions, including
+the **Core API** and Convenient Transaction API. To begin using the Core
+API, declare the start and commit points of the transaction.
+
+To see an example that uses the Convenient Transaction API, see the
```
ccho-mongodb, line 70: Issue:

I think the information that explains what's different about the Convenient Transaction and Core API is missing. As a user, I would want to know when to use one over the other.

Suggestion:

I think this page. given it is the only Fundamentals page about Transactions in this docs set, should describe the differences between the Core API and Convenient Transaction in detail. The Usage Example should probably only contain the minimum information needed to show an example of the syntax and output.

source/usage-examples.txt
ccho-mongodb: I wasn't able to find it in the [spec](https://github.com/mongodb/docs-node/pull/751/files#r1307624029), so I think it would be better to describe as mentioned.

source/usage-examples/convenient-transactions.txt
```diff
@@ -0,0 +1,174 @@
...
+- ``withSession()``: Runs a given operation within a session. You do not
+  have to handle the creation or termination of the session.
+- ``withTransaction()``: Starts a transaction and runs a provided
+  function. This method ensures that the ``commitTransaction()`` method
+  is called when all of the operations in the function complete.
+
+.. important:: Return Types and Values
+   
+   Starting in version 6.0 of the {+driver-short+}, these methods return
+   the value that the provided function returns.
```
ccho-mongodb, line 42: Issue:
I think shortly after you prepared this PR, we had a team discussion and decision to exclude this versioned information. 

Question:
Is the Convenient Transaction API new to some other version other than v6? Was there a previous behavior to compare it with? Or are these methods not specific to the Convenient Transaction API in which case they probably shouldn't be described as such in the previous list.

ccho-mongodb (changes requested):

source/code-snippets/transactions/txn-convenient.js
```diff
@@ -0,0 +1,83 @@
...
+          return "Item not found or insufficient quantity.";
+        }
+
+        // If inventory is sufficient, update inventory
+        // to account for the purchase.
+        await invColl.updateOne(
+          { item: item.item },
+          { $inc: { qty: -item.qty } },
+          { session }
+        );
```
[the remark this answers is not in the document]
rustagir: Yes, the operations are separated because the transaction should abort if no matching documents are found (aka insufficient quantity). I guess we still could combine this in one compound operation, and just abort the transaction based on the return type of `findOneAndUpdate()`, but I think separating them makes things more clear for the user to understand

source/code-snippets/transactions/txn-convenient.js
```diff
@@ -0,0 +1,83 @@
...
+  session
+    .withTransaction(async (session) => {
+      const invColl = client.db("testdb").collection("inventory");
+      const recColl = client.db("testdb").collection("records");
+
+      let total = 0;
+      for (const item of order) {
+        // Abort the transaction if the item
+        // does not exist or has insufficient inventory
+        const checkInventory = await invColl.findOne(
```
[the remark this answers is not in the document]
rustagir: will change in both code files!

source/code-snippets/transactions/txn-core.js
```diff
@@ -18,8 +18,8 @@ async function setup(client) {
     await customerColl.insertOne({ _id: 98765, orders: [] });
 
     await inventoryColl.insertMany([
-      { name: 'sunblock', sku: 5432, qty: 85 },
-      { name: 'beach towel', sku: 7865, qty: 41 },
+      { item: 'sunblock', sku: 5432, qty: 85 },
+      { item: 'beach towel', sku: 7865, qty: 41 },
```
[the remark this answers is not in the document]
rustagir: I'll just change the name of the field to item_id or something similar, not all users might be familiar with SKU anyways.

source/fundamentals/crud/compound-operations.txt
```diff
@@ -51,8 +51,8 @@ configurable :ref:`sort <node-fundamentals-sort>` and
    of these methods.
    
    Starting in version 6.0, this setting defaults to ``false``, which
-   means that each method returns the matched document, or, if no
-   document is matched, it returns ``null``. If you set
+   means that each method returns the matched document. If no
```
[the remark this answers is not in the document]
rustagir: Yes, but this was just a quick fix in this PR -- I think that further work is out of the scope of this documentation. Creating a ticket - [DOCSP-32806](https://jira.mongodb.org/browse/DOCSP-32806)

source/fundamentals/transactions.txt
```diff
@@ -63,40 +63,45 @@ MongoDB in the following sections of this guide:
 Transaction APIs
 ----------------
 
-Use the **Core API** to implement a transaction with the driver. To use the Core
-API, you declare the start and commit points of the transaction.
+The driver provides multiple APIs for performing transactions, including
+the **Core API** and Convenient Transaction API. To begin using the Core
```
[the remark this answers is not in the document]
rustagir: fixed this sentence as I've moved content from the UE to this page in a new section for the Conv. Txn API

source/fundamentals/transactions.txt
```diff
@@ -63,40 +63,45 @@ MongoDB in the following sections of this guide:
...
+API, declare the start and commit points of the transaction.
+
+To see an example that uses the Convenient Transaction API, see the
+:ref:`Perform a Convenient Transaction <node-usage-convenient-txn>` usage example.
```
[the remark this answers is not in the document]
rustagir: revised

source/fundamentals/transactions.txt
```diff
@@ -63,40 +63,45 @@ MongoDB in the following sections of this guide:
...
 
-Transaction Settings
---------------------
+Transaction Options
+-------------------
 
-When you instantiate a transaction, you can specify the following options to
-set the default behavior for that transaction:
+You can specify the following options to
+specify the behavior for a transaction when creating it:
```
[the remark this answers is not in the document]
rustagir: These options can be passed to either `startTransaction()` or `withTransaction()`. I will make that more clear.

source/fundamentals/transactions.txt
```diff
@@ -163,86 +173,88 @@ order details. Suppose you organize the data updates as follows:
      - Record the purchase information
 
    * - ``customers``
-     - update
-     - Append the order id to associate it with the customer
+     - update or upsert
+     - Append the order ``_id`` to the document for a customer
```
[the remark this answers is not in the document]
rustagir: From my understanding, if a customer makes a purchase, the _id from the order document is also appended to the customer document in this collection, so that there is something like an "order history" for each customer, in addition to a running, chronological record of all orders

source/fundamentals/transactions.txt
```diff
@@ -163,86 +173,88 @@ order details. Suppose you organize the data updates as follows:
...
 
 Core API Implementation
 ~~~~~~~~~~~~~~~~~~~~~~~
 
-The code example in this section demonstrates how you can use the Core API to
-run the multi-document payment transaction in a session. This function
-shows how you can perform the following:
+The code example in this section demonstrates how to use the Core API to
+perform the multi-document transaction in a session. This example
+shows how to perform the following actions:
```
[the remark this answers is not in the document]
rustagir: I think that heading is confusing -- I will add some descriptive text here, but also change that heading.

source/usage-examples.txt
rustagir: Convenient Transaction API is the terminology that Neal provided, but I agree that its still a regular transaction. I think the Usage example can just be called Perform a transaction.

source/fundamentals/transactions.txt
rustagir: made significant changes to move information around

source/usage-examples/convenient-transactions.txt
rustagir: I can remove the versioning from this sentence - the previous behavior is described in the what;s new/upgrade pages.
These methods are specific to the this API.

source/code-snippets/transactions/txn-convenient.js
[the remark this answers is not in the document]
ccho-mongodb: I think splitting the operations makes it the example unnecessarily long and might cause confusion as to why the find is separated from update. I think it would make more sense to use the compound operation. I think the insert into the inventory collection sufficiently demonstrates a second operation in the transaction since it can't be combined in a single compound operation.

source/code-snippets/transactions/txn-convenient.js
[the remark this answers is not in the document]
ccho-mongodb: Suggestion:

I think "inStock" is shorter than "isInStock" and can convey the same meaning.

source/fundamentals/transactions.txt
```diff
@@ -160,89 +240,95 @@ order details. Suppose you organize the data updates as follows:
...
+If the payment fails, you can perform a transaction that includes all of
+the data changes to ensure you avoid making any partial updates that
+might cause data consistency issues for future operations.
 
 Sample Data
 ~~~~~~~~~~~
 
-The code examples require the following sample data in the ``testdb``
-database to run the multi-document payment transaction:
+The code examples use the following sample data in the ``testdb``
```
ccho-mongodb, line 267: Suggestion:

I couldn't find more than one example (Core API Implementation).
Could it make sense to entitle the parent header "Core API Example" and then move the admonition about the Convenient Transaction API usage example directly after that title?

source/fundamentals/transactions.txt
```diff
@@ -160,89 +240,95 @@ order details. Suppose you organize the data updates as follows:
...
 
 1. Start a session
-#. Start a transaction, specifying transaction options
-#. Perform data operations in the same session
-#. Commit a transaction, or cancel it if the driver encounters an error
-#. End a session
+#. Start a transaction with transaction options
+#. Perform read and write operations in the session to record the purchase
+#. Commit the transaction, or end it if the driver encounters an error
+#. End the session
```
ccho-mongodb, line 331: Suggestion:
I think it would be good to keep consistent with indefinite articles throughout the list. The prior items refer to an indefinite session and transaction, so the rest should be similar, e.g.:

"Perform read and write operations in a session
Commit a transaction or end it...
End a session"

source/fundamentals/transactions.txt
```diff
@@ -273,28 +359,34 @@ you want to run on that session.
...
+.. tip::
+   
+   To view an example that uses the Convenient Transaction API, see the
+   :ref:`node-usage-convenient-txn` usage example.
 
 .. _nodejs-transaction-example-payment-result:
 
-Payment Transaction Result
-~~~~~~~~~~~~~~~~~~~~~~~~~~
+Transaction Result
```
ccho-mongodb, line 380: Suggestion:

I think "Transaction" should be plural since the section describes two results -- one in which it was successful and another in which an error interrupts the transaction.

```suggestion
Transaction Results
```

source/upgrade.txt
```diff
@@ -102,6 +102,12 @@ Version 6.x Breaking Changes
   perform any operations in the session.
 - The ``includeResultMetadata`` option for compound operation methods is
   ``false`` by default.
+- The ``withSession()`` method returns the value that the provided
+  function returns. In previous driver versions, this function returns
+  ``void``. The ``withTransaction()`` method returns the value that the
```
ccho-mongodb, line 107: Suggestion:
I could be wrong, but I don't remember "void" being a valid Javascript type. I think it returns "undefined" and that it uses the void operator in the method signature:
- [withSession() API doc from v5.0](https://mongodb.github.io/node-mongodb-native/5.0/classes/MongoClient.html#withSession)
- ["A void function returns undefined"](https://stackoverflow.com/questions/44216949/does-a-async-function-which-returns-promisevoid-have-an-implicit-return-at-the)

source/usage-examples/transaction.txt
```diff
@@ -0,0 +1,141 @@
...
+
+.. seealso::
+   
+   To learn about the Convenient Transaction API in more detail, see the
+   :ref:`Convenient Transaction API <node-txn-convenient-api>` section of the
+   Transactions guide.
+
+   To learn about the Core API and view an example, see the
+   :ref:`Core API <node-txn-core-api>` section of the
+   Transactions guide.
```
ccho-mongodb, line 19: Suggestion:
I think since these link to sections of the same page, it can be easier to treat them together:

"To learn more about the Transactions APIs and see an transaction example that uses the Core API, see the Transactions documentation."

ccho-mongodb (changes requested):
I think this is moving in the right direction. However, I stopped my review when I realized the usage example contained similar, but different text than the fundamentals example. Here's the issue and a proposed solution. If anything doesn't make sense or if you have questions, we can connect synchronously next week.

Issue:

This usage example is the same example presented on the Fundamentals page, with slightly less info about the result. I think this arrangement could confuse users as to which section they should expect to find runnable code examples and which sections contain explanatory content.

Suggestion:

I think there's a relatively low-lift solution to the issue of content organization. 

- Keep this Usage Example, but use the wording from the reviewed fundamentals page when possible. I'm not sure it makes sense to maintain and review two similarly worded example scenarios.
- Add another Usage Example for the Core API implementation (the duplicated scenario and data could be addressed later)
- Remove the code example from the Fundamentals page, and link to the usage example instead
- Add snippets of code that show how to call a transaction with each API to the Fundamentals page

source/fundamentals/transactions.txt
```diff
@@ -63,40 +63,112 @@ MongoDB in the following sections of this guide:
...
+  is called when all of the operations in the function finish.
+
+These methods return the value that the provided function returns.
+
+For example, if a function you provide to the ``withTransaction()``
+method returns a ``Document`` type, then the ``withTransaction()``
+method returns that object. Furthermore, if you perform this
+transaction within the function you provide to the ``withSession()``
+method, then the ``withSession()`` method also returns the
+``Document`` object.
```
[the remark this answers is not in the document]
rustagir: Shortened the description and made the logic more clear

source/fundamentals/transactions.txt
```diff
@@ -273,28 +359,34 @@ you want to run on that session.
 
 The code and comments in the ``catch`` block demonstrate how you can identify
 the server transaction errors and where you can place your logic to handle
-them. Make sure to include the ``MongoError`` type from the driver in your
-code as shown in the following sample import statement:
+them. Make sure to import the ``MongoError`` type from the driver
+package as shown in the following sample import statement:
```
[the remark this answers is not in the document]
rustagir: I can take this snippet out but leave the information in the code patterns admonition for the core api

source/fundamentals/transactions.txt
rustagir: not relevant as I moved the example to its own page

source/fundamentals/transactions.txt
rustagir: Revising this whole section to be more specific

source/usage-examples/transaction.txt
rustagir: not applicable anymore

source/upgrade.txt
rustagir: https://github.com/mongodb/node-mongodb-native/releases/#:~:text=Previously%2C%20this%20function%20returned%20void%20this%20is%20a%20feature%20to%20align%20with%20the%20following%20breaking%20change. Uses the wording "returns void" -- so I'm not sure.

source/fundamentals/transactions.txt
```diff
@@ -45,313 +45,240 @@ you want to execute sequentially. We recommend you reuse
...
+Convenient Transaction API
+~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-.. code-block:: json
+The Convenient Transaction API provides the following methods to
+implement transactions:
 
-   { _id: 98765, orders: [] }
+- `withSession() <{+api+}/classes/MongoClient.html#withSession>`__: Runs
+  a given operation within a session. The API handles the creation and
```
ccho-mongodb, line 162: Suggestion:
I think it would be easier to understand what the "given operation" is by using the same terminology as in the next list item:
E.g. "Runs the callback passed to it within a session."

source/fundamentals/transactions.txt
```diff
@@ -45,313 +45,240 @@ you want to execute sequentially. We recommend you reuse
...
+- `withTransaction() <{+api+}/classes/ClientSession.html#withTransaction>`__:
+  Runs the callback passed to it within a transaction and calls the 
+  ``commitTransaction()`` method when the callback returns.
 
-The documents in the ``inventory`` collection in this example contain the
-following:
+These methods return the value that the callback returns. For example,
+if a callback you pass to the ``withTransaction()`` method returns a
+``Document`` type, then the ``withTransaction()`` method returns that
+object.
```
ccho-mongodb, line 171: Suggestion:

I think "that object" might be difficult to identify in this sentence. Maybe it could help to use a tangible return value.
E.g.

If the callback you pass to withTransaction() returns "hello, world", the value the withTransaction() method returns is also "hello world".

source/fundamentals/transactions.txt
```diff
@@ -45,313 +45,240 @@ you want to execute sequentially. We recommend you reuse
...
+   * - ``maxCommitTimeMS``
+     - Specifies the length of time that a commit action on a
+       transaction can run, in milliseconds.
 
-   For example, the following code generates an
-   ``MongoInvalidArgumentError`` error because it creates
-   a ``ClientSession`` instance from the ``client1`` client, but provides
-   this session to the ``client2`` client for a write operation:
+For a full list of options, see the API documentation for
+`TransactionOptions <{+api}/interfaces/TransactionOptions.html>`__.
```
ccho-mongodb, line 244: Issue:
The api source constant is malformed.

Suggestion:
I think `{+api+}` could fix the issue.

source/usage-examples/transaction-core.txt
```diff
@@ -0,0 +1,239 @@
...
+   :start-after: start placeOrder
+   :end-before: end placeOrder
+
+.. _nodejs-transaction-example-payment-result:
+
+Transaction Results
+~~~~~~~~~~~~~~~~~~~
+
+When your application completes the payment transaction, the collections
+in the database contains the customer and order updates.
```
ccho-mongodb, line 162: Issue:
I think there is mismatched verb agreement between "collections (in the database)" and "contains".

Suggestion:
"collections in the database contain"

ccho-mongodb (changes requested):
Leaving a few comments here on the Fundamentals and Upgrade page, will resume my review the Usage Example pages on Monday.

source/usage-examples/transaction-conv.txt
```diff
@@ -0,0 +1,188 @@
...
+
+Sample Orders and Transaction Results
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+When your application completes the payment transaction, the collections
+in the database contains the inventory and order updates. 
+If an error interrupts the transaction, none of the changes appear in
+your database.
+
+Suppose a customer sends the following order:
```
ccho-mongodb, line 130: Suggestion:
I think it would be easier to scan if the reader could identify this quickly a "transaction completed" example and the other code snippet was identifiable as a "transaction rolled back" example. E.g.

"The following sample order shows a successful order in which sufficient inventory exists and the transaction successfully completes:"

"The following sample order shows an unsuccessful order in which insufficient inventory exists and the transaction ends and rolls back the changes:"

source/usage-examples/transaction-conv.txt
```diff
@@ -0,0 +1,188 @@
...
+is ``29``. The ``orders`` collection contains the record of the
+purchase.
+
+Suppose a different customer sends the following order:
+
+.. literalinclude:: /code-snippets/transactions/txn-convenient.js
+   :language: javascript
+   :dedent:
+   :copyable: false
+   :start-after: start-order-fail
```
ccho-mongodb, line 167: Added https://jira.mongodb.org/browse/DOCSP-33010 to modify this Vale rule.

ccho-mongodb (changes requested):

source/fundamentals/transactions.txt
```diff
@@ -45,313 +45,240 @@ you want to execute sequentially. We recommend you reuse
...
-   * - ``customers``
-     - update
-     - Append the order id to associate it with the customer
+The following code demonstrates how to perform a transaction by using
+the Core API:
 
-   * - ``inventory``
-     - update
-     - Subtract quantity of ordered items
+.. literalinclude:: /code-snippets/transactions/txn-simple.js
```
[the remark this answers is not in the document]
rustagir: renamed

source/usage-examples/transaction-conv.txt
```diff
@@ -0,0 +1,188 @@
...
+
+.. seealso::
+   
+   To learn more about the performing transactions in the
+   {+driver-short+}, see the :ref:`nodejs-transactions` guide.
+
+   To learn how to use the Core API to perform transactions, see the
+   :ref:`node-usage-core-txn` usage example.
+
+Example
```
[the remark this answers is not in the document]
rustagir: That note describes how to load sample data, which is not used in either of the usage examples. I can create a note that is more applicable to these examples to include, though.

source/usage-examples/transaction-conv.txt
```diff
@@ -0,0 +1,188 @@
...
+-------
+
+Consider a situation in which a customer purchases items from your store.
+To record the purchase, your application must update
+your inventory and record the order information.
```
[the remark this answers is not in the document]
rustagir: I removed some of the information under the Example heading, but If the real issue here is that the examples are too complicated for Usage examples, then perhaps we could make another ticket to simplify them.

source/usage-examples/transaction-conv.txt
```diff
@@ -0,0 +1,188 @@
...
+
+The code example in this section demonstrates how to use the Convenient
+Transaction API to perform a multi-document transaction in a session. In
+this example, the transaction reflects the actions needed when a
+customer purchases items from your store.
+
+This example code performs a transaction through the following actions:
+
+1. Calls the ``withSession()`` method on the client to implicitly create
+   the session and include the transaction within this session.
```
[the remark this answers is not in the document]
rustagir: changed wording

source/usage-examples/transaction-core.txt
```diff
@@ -0,0 +1,233 @@
...
+
+   You can use this example to connect to an instance of MongoDB
+   and interact with a database that contains sample data. To learn more
+   about connecting to your MongoDB instance, see the :doc:`Usage Examples
+   guide </usage-examples>`.
+
+.. literalinclude:: /code-snippets/transactions/txn-core.js
+   :language: javascript
+   :linenos:
+   :emphasize-lines: 3-7,10,13,24,38,51,55,58-67,71,74
```
ccho-mongodb, line 148: Issue:
These seem to be off by a line in certain places (blank and nonfunctional lines are highlighted). 

Suggestion:
I think it could be good to only highlight items that need specific callouts, if any. The convenient transaction API usage example in this PR doesn't include any emphasize-lines attribute.

ccho-mongodb (approved):
Looks pretty good! A couple issues and suggestions to fix. 

Although there are structural/presentation issues that I did not go into detail, but since we want to get this content out as soon as possible, we can revisit ensuring the Usage Examples content is presented consistently throughout the section.

source/code-snippets/transactions/txn-api-examples.js
```diff
@@ -0,0 +1,53 @@
...
+// begin-core
+async function coreTest(client) {
+  const session = client.startSession();
+  try {
+    session.startTransaction();
+
+    const myColl = client.db("testdb").collection("coll");
+    await myColl.insertOne({ x: 1, y: 6 }, { session });
+
+    // ... perform other operations
```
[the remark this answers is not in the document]
rustagir: will change both examples in that file!

source/usage-examples/transaction-conv.txt
```diff
@@ -0,0 +1,180 @@
...
+   :widths: 25 25 50
+   :header-rows: 1
+
+   * - Collection
+     - Operation
+     - Description of the Change
+
+   * - ``orders``
+     - insert
+     - Insert a document that describes the order
```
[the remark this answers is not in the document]
rustagir: changed tables in both examples

source/usage-examples/transaction-core.txt
rustagir: I'll just remove the highlighting from the UEs and leave in the transaction examples.

PR #754, opened 2023-08-30
DOCSP-32718: deletemany UE code comments

rustagir:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-32718>
https://preview-mongodbrustagir.gatsbyjs.io/node/DOCSP-32718-deletemany-comments/usage-examples/deleteMany/

### DIFF
--- a/source/code-snippets/usage-examples/deleteMany.js
+++ b/source/code-snippets/usage-examples/deleteMany.js
@@ -11,22 +11,20 @@ const client = new MongoClient(uri);
 
 async function run() {
   try {
-    // Access the movies collection from the sample_mflix database.
     const database = client.db("sample_mflix");
     const movies = database.collection("movies");
 
-    // Create a filter for movies with a title containing the string "Santa".
+    /* Delete all documents that match the specified regular
+    expression in the title field from the "movies" collection */
     const query = { title: { $regex: "Santa" } };
-
-    // Delete all documents that match the preceding query filter.
     const result = await movies.deleteMany(query);
 
-    // Print the number of deleted documents.
+    // Print the number of deleted documents
     console.log("Deleted " + result.deletedCount + " documents");
   } finally {
-    // Close the client after the operation completes.
+    // Close the client after the operation completes
     await client.close();
   }
 }
-// Run the program and handle any errors that occur during execution.
+// Run the program and print any thrown exceptions
 run().catch(console.dir);
--- a/source/code-snippets/usage-examples/deleteMany.js
+++ b/source/code-snippets/usage-examples/deleteMany.js
@@ -1,21 +1,32 @@
+/* Delete multiple documents */
+
+// Import the MongoClient type from the mongodb package.
 import { MongoClient } from "mongodb";
 
 // Replace the uri string with your MongoDB deployment's connection string.
 const uri = "<connection string uri>";
 
+// Create a new client and connect to MongoDB.
 const client = new MongoClient(uri);
 
 async function run() {
   try {
+    // Access the movies collection from the sample_mflix database.
     const database = client.db("sample_mflix");
     const movies = database.collection("movies");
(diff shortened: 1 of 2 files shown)


source/code-snippets/usage-examples/deleteMany.js
```diff
@@ -1,21 +1,32 @@
...
 
 // Replace the uri string with your MongoDB deployment's connection string.
 const uri = "<connection string uri>";
 
+// Create a new client and connect to MongoDB.
 const client = new MongoClient(uri);
 
 async function run() {
   try {
+    // Access the movies collection from the sample_mflix database.
```
ccho-mongodb, line 14: Suggestion:

Lines 15-16 just get a reference to the collection. I think "access" means interact with it, which is started in line 22 when the deleteMany operation is called on the collection.

source/code-snippets/usage-examples/deleteMany.ts
```diff
@@ -1,22 +1,34 @@
...
     const result = await movies.deleteMany({ title: { $regex: "Santa" } });
+    
+    // Print the number of deleted documents.
     console.log("Deleted " + result.deletedCount + " documents");
   } finally {
+    // Close the client after the operation completes.
     await client.close();
   }
 }
+// Run the program and handle any errors that occur during execution.
```
ccho-mongodb, line 33: Suggestion:
I think this only prints the exception rather than handles them. I think "handle [an] error" usually refers to the control logic (rethrow, close resources, etc.). I think printing could technically be handling the error, but probably not the first thought. I think this could be more accurate and concise:

"Run the program and print any thrown exceptions."

source/code-snippets/usage-examples/deleteMany.ts
```diff
@@ -1,22 +1,34 @@
...
   title: string;
 }
 
 async function run() {
   try {
+    // Access the movies collection from the sample_mflix database.
     const database = client.db("sample_mflix");
     const movies = database.collection<Movie>("movies");
+
+    // Delete all documents where the title contains the string "Santa".
```
ccho-mongodb, line 23: Suggestion:
Instead of "where" (location), I think it could be good to use "in which" or to describe the matching.

While I think it can be helpful to include values, since this is a regular expression, it might get awkward to try to reproduce this language for a more complex regular expression. I think it could be more general since the value is apparent and self-explanatory in the code.

I think it could be helpful to skip the comment "Access the movies collection..." and reference the collection in this comment instead since this is closer to the usage of the collection. E.g.

"Delete all documents that match the specified regular expression in the title field from the "movies" collection."

source/code-snippets/usage-examples/deleteMany.ts
```diff
@@ -1,22 +1,34 @@
...
+// Import the MongoClient type from the mongodb package.
 import { MongoClient } from "mongodb";
 
 // Replace the uri string with your MongoDB deployment's connection string.
 const uri = "<connection string uri>";
 
+// Create a new client and connect to MongoDB.
 const client = new MongoClient(uri);
 
+// Define the Movie interface to model the data.
```
ccho-mongodb, line 12: Suggestion:

I'm open to any discussion, but I think some of the information is self-apparent. I think "Create a sample interface" contains enough information.


Also, this code isn't used in this example, so I think it could be good to remove (feel free to ticket if it's not trivial to fix and check).

source/code-snippets/usage-examples/deleteMany.ts
ccho-mongodb: I think this is also inconsistent with the other comment in the JS file even though the code is the same:
"// Delete all documents that match the preceding query filter."

ccho-mongodb (approved):
LGTM, a few suggestions.

PR #803, opened 2023-09-15
DOCSP-33103: v6.1 release

rustagir:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-33103>
Staging - https://preview-mongodbrustagir.gatsbyjs.io/node/DOCSP-33103-6.1-release/
- https://preview-mongodbrustagir.gatsbyjs.io/node/DOCSP-33103-6.1-release/compatibility/
- https://preview-mongodbrustagir.gatsbyjs.io/node/DOCSP-33103-6.1-release/whats-new/#what-s-new-in-6.1

## Self-Review Checklist


source/whats-new.txt
```diff
@@ -43,6 +44,32 @@ Learn what's new in:
...
+The {+driver-short+} v6.1 release includes the following features:
+
+- Raises the ``bson`` package version to 6.1.0 to expose the
+  ``Decimal128.fromStringWithRounding()`` method. To learn more, see the
+  `v6.1.0 bson release notes <https://github.com/mongodb/js-bson/releases/tag/v6.1.0>`__.
+
+- Detects environment variables for region settings when
+  authentication by using the ``MONGODB-AWS`` authentication mechanism.
+  The driver sets region options only if both of the following
+  environment variables are present:
```
ccho-mongodb, line 61: Suggestion:

I think it could be good to identify who is responsible for setting the environment variables. I also think it could be more clear to mention they need to be set rather than "present". The latter doesn't seem like a common way to describe environment variables since they are either set or not set.

Maybe a sentence with a different structure could present it more clearly:

```suggestion
   To instruct the driver to pass your region options, you must set both of the following environment variables:
```

I wonder if maybe this information might be easier to find if incorporated in the auth mechanism content, and replace this explanation with a link to that content.

source/whats-new.txt
```diff
@@ -224,6 +251,9 @@ New features of the 5.5 {+driver-short+} release include:
   later if you want to create an encrypted collection by using the
   Queryable Encryption feature.
 
+To learn more about this release, see the `v5.5.0 Release Highlights
+<https://github.com/mongodb/node-mongodb-native/releases/tag/v5.5.0>`__.
+
```
ccho-mongodb, line 256: Suggestion:
No action necessary, but older changes might be better in a separate PR to make it easier to backport to relevant versions. I think this page is one that the team should segment by major version, so it's probably not worthwhile backporting.

ccho-mongodb (approved):
LGTM, left a couple suggestions.

source/whats-new.txt
rustagir: Creating a ticket for this

PR #894, opened 2024-06-11
DOCSP-40099: Add note about client.connect()

mcmorisi:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-40099>
Staging - <https://preview-mongodbmcmorisi.gatsbyjs.io/node/DOCSP-40099-client-connect/fundamentals/connection/connect/#atlas-connection-example>

## Self-Review Checklist

### DIFF
--- a/source/fundamentals/connection/connect.txt
+++ b/source/fundamentals/connection/connect.txt
@@ -95,6 +95,14 @@ verify that the connection is successful:
 .. literalinclude:: /code-snippets/connection/stable-api-connect.js
    :language: javascript
 
+.. note:: Calling ``MongoClient.connect()``
+
+   As of v4.7, the {+driver-short+} will automatically call the ``MongoClient.connect()``
+   method when using the client to interact to your MongoDB instance, **except** when
+   you use the client to call bulk operations. We recommend manually calling
+   ``MongoClient.connect()`` if you want granular control and insight over
+   your application's connections.
+
 To learn more about the Stable
 API feature, see the :ref:`{+stable-api+} page <nodejs-stable-api>`.
 
--- a/source/fundamentals/connection/connect.txt
+++ b/source/fundamentals/connection/connect.txt
@@ -95,10 +95,10 @@ verify that the connection is successful:
 .. literalinclude:: /code-snippets/connection/stable-api-connect.js
    :language: javascript
 
-.. note:: Calling ``MongoClient.connect()``
+.. note::
 
    As of v4.7, the {+driver-short+} will automatically call the ``MongoClient.connect()``
-   method when using the client to interact to your MongoDB instance, **except** when
+   method when using the client to interact with your MongoDB instance, **except** when
    you use the client to call bulk operations. We recommend manually calling
    ``MongoClient.connect()`` if you want granular control and insight over
    your application's connections.


source/fundamentals/connection/connect.txt
```diff
@@ -95,6 +95,14 @@ verify that the connection is successful:
 .. literalinclude:: /code-snippets/connection/stable-api-connect.js
    :language: javascript
 
+.. note::
+
+   As of v4.7, the {+driver-short+} will automatically call the ``MongoClient.connect()``
```
mongoKart, line 100: S: specify driver version (as opposed to server version)

```suggestion
   As of {+driver-short+} v4.7, the driver will automatically call the ``MongoClient.connect()``
```

mongoKart, line 100: I: present tense

```suggestion
   As of v4.7, the {+driver-short+} automatically calls the ``MongoClient.connect()``
```

source/fundamentals/connection/connect.txt
```diff
@@ -95,6 +95,14 @@ verify that the connection is successful:
...
+   method when using the client to interact with your MongoDB instance, **except** when
```
mongoKart, line 101: I: the prerequisite of "using the client to interact with..." isn't clear to me. what does the user have to do to get the driver to call this automatically?

mongoKart, line 101: I: there was a whole thing about a year ago about server vs. deployment vs. instance. i think the upshot was:

- MongoDB Server = the product
- MongoDB deployment = a running instance of MongoDB
- MongoDB instance = not preferred

```suggestion
   method when using the client to interact with your MongoDB deployment, **except** when
```

mongoKart, line 101: S: the sentence is a little long, and the "except" clause seems pretty important. better as its own sentence?

source/fundamentals/connection/connect.txt
```diff
@@ -95,6 +95,14 @@ verify that the connection is successful:
...
+   you use the client to call bulk operations. We recommend manually calling
+   ``MongoClient.connect()`` if you want granular control and insight over
+   your application's connections.
```
mongoKart, line 104: S: i think it would be useful to be more specific here; "granular control and insight" is correct, but might not be helpful to a newer user. (you could also add info about bulk writes here)

```suggestion
   you use the client to call bulk operations. Call the ``MongoClient.connect()`` method
   explicitly if you are performing bulk operations or if you want to verify that the
   connection was successful.
```

mongoKart (changes requested):
Some small changes and suggestions, but nice job overall

source/fundamentals/connection/connect.txt
```diff
@@ -95,6 +95,13 @@ verify that the connection is successful:
 .. literalinclude:: /code-snippets/connection/stable-api-connect.js
    :language: javascript
 
+.. note::
+
+   As of {+driver-short+} v4.7, the driver automatically calls the ``MongoClient.connect()``
+   method when using the client to perform CRUD operations on your MongoDB deployment.
+   Call the ``MongoClient.connect()`` method explicitly to verify that the connection is
+   successful.
+
```
mcmorisi, line 104: @nbbeeken in the #node-driver-docs Slack channel posted proof that this automatic connection still works with bulk ops.

"Call the ``MongoClient.connect()`` method explicitly _if you would like to_ verify that the connection is successful" gets flagged by Vale for the italicized portion.

mongoKart: How about: "...if you want to manually verify that the connection..."? As worded, it sounds like we're directing users to do it, which isn't necessary.

source/fundamentals/connection/connect.txt
```diff
@@ -97,11 +97,10 @@ verify that the connection is successful:
 
 .. note::
 
-   As of v4.7, the {+driver-short+} will automatically call the ``MongoClient.connect()``
-   method when using the client to interact with your MongoDB instance, **except** when
-   you use the client to call bulk operations. We recommend manually calling
-   ``MongoClient.connect()`` if you want granular control and insight over
-   your application's connections.
+   As of {+driver-short+} v4.7, the driver automatically calls the ``MongoClient.connect()``
```
mongoKart, line 100: S: this is presumably getting backported to all active docs versions, and v4.7 is in legacy. Do we need to say when this behavior started?

```suggestion
   The {+driver-short+} automatically calls the ``MongoClient.connect()``
```

mongoKart (approved):
A couple of suggestions I'm happy to talk through, but LGTM otherwise!

source/fundamentals/connection/connect.txt
mcmorisi: Well spotted, happy to remove.

PR #908, opened 2024-10-09
DOCSP-43242: Improve UTF-8 validation documentation to clarify validation occurs on decoded data only

mayaraman19:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-43242>
Staging - <https://deploy-preview-908--docs-node.netlify.app/fundamentals/bson/utf8-validation/>

## Self-Review Checklist

- [x] Are the [facets and meta keywords](https://wiki.corp.mongodb.com/display/DE/Docs+Taxonomy) accurate?

### DIFF
--- a/source/fundamentals/bson/utf8-validation.txt
+++ b/source/fundamentals/bson/utf8-validation.txt
@@ -25,9 +25,10 @@ processing overhead since it needs to check the data.
 If you *disable* validation, your application avoids the validation processing
 overhead, but cannot guarantee consistent presentation of invalid UTF-8 data.
 
-The driver enables UTF-8 validation on documents coming from MongoDB by default. 
-It checks incoming documents for any characters that are not encoded in a 
-valid UTF-8 format when it transfers data from MongoDB to your application.
+By default, the driver enables UTF-8 validation on BSON strings in documents 
+from MongoDB. It checks incoming documents for any characters that are not 
+encoded in a valid UTF-8 format when it transfers data from MongoDB to your 
+application.
 
 .. note::
 
--- a/source/fundamentals/bson/utf8-validation.txt
+++ b/source/fundamentals/bson/utf8-validation.txt
@@ -25,10 +25,9 @@ processing overhead since it needs to check the data.
 If you *disable* validation, your application avoids the validation processing
 overhead, but cannot guarantee consistent presentation of invalid UTF-8 data.
 
-By default, the driver enables UTF-8 validation on BSON strings in documents
-from MongoDB. It checks incoming strings for any characters that are not 
-encoded in a valid UTF-8 format when it transfers data from MongoDB to your 
-application.
+By default, the driver enables UTF-8 validation on data from MongoDB. 
+It checks incoming strings for any characters that are not encoded in a 
+valid UTF-8 format when it transfers data from MongoDB to your application.
 
 .. note::
 
--- a/source/fundamentals/bson/utf8-validation.txt
+++ b/source/fundamentals/bson/utf8-validation.txt
@@ -31,7 +31,7 @@ valid UTF-8 format when it parses data sent from MongoDB to your application.
 
 .. note::
 
-   The current version of the {+driver-short+} automatically substitutes invalid
+   This version of the {+driver-short+} automatically substitutes invalid
    `lone surrogates <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String#utf-16_characters_unicode_code_points_and_grapheme_clusters>`__
    with the `replacement character <https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toWellFormed>`__ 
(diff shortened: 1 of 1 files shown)


source/fundamentals/bson/utf8-validation.txt
```diff
@@ -25,9 +25,9 @@ processing overhead since it needs to check the data.
 If you *disable* validation, your application avoids the validation processing
 overhead, but cannot guarantee consistent presentation of invalid UTF-8 data.
 
-The driver enables UTF-8 validation by default. It checks documents for any
-characters that are not encoded in a valid UTF-8 format when it transfers data
-between your application and MongoDB.
+The driver enables UTF-8 validation on documents coming from MongoDB by default.
```
jordan-smith721, line 28: Based on the jira ticket, I think it's important to specify that this is for BSON strings. Reworded the sentence in my suggestion a bit to make it flow better with the added information

```suggestion
By default, the driver enables UTF-8 validation on documents from MongoDB servers that contain BSON string values.
```

jordan-smith721 (approved):
One suggestion but otherwise LGTM!

source/fundamentals/bson/utf8-validation.txt
mayaraman19: @jordan-smith721 Thanks for the suggestion on flow! I was torn on adding the part about BSON strings since the preceding part of the page doesn't mention them, and (I think) it's implied throughout the page that in this context they are synonymous with documents.

source/fundamentals/bson/utf8-validation.txt
```diff
@@ -25,9 +25,10 @@ processing overhead since it needs to check the data.
 If you *disable* validation, your application avoids the validation processing
 overhead, but cannot guarantee consistent presentation of invalid UTF-8 data.
 
-The driver enables UTF-8 validation by default. It checks documents for any
-characters that are not encoded in a valid UTF-8 format when it transfers data
-between your application and MongoDB.
+By default, the driver enables UTF-8 validation on BSON strings in documents
+from MongoDB. It checks incoming strings for any characters that are not 
+encoded in a valid UTF-8 format when it transfers data from MongoDB to your 
+application.
```
nbbeeken, line 31: The section below is outside of the diff but I think we should take another look at improving its clarity if that's alright:

> The current version of the {+driver-short+} automatically substitutes
   invalid UTF-8 characters with alternate valid UTF-8 ones before
   validation when you send data to MongoDB. Therefore, the validation
   only throws an error when the setting is enabled and the driver
   receives invalid UTF-8 document data from MongoDB.

Specifically this bit:

> automatically substitutes invalid UTF-8 characters with alternate valid UTF-8 ones before validation when you send data to MongoDB.

I do not think we need to write a whole lesson on UTF and all the glossary of terms that come with it we can lean on the documentation here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String#utf-16_characters_unicode_code_points_and_grapheme_clusters

The important bit from that is strings in JS are made up of UTF-16 code points and you sometimes need more than one code point to represent a character, when multi-code-points (surrogate pairs) are accidentally split up, there is no way to represent that character anymore so you cannot translate it to UTF-8. There is a well-known filler character that is used in its place, U+FFFD that makes the output UTF-8 valid without removing the character entirely. 

We should say "substitutes lone surrogates with the replacement character", ref: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/isWellFormed

> with alternate valid UTF-8 ones

It is specifically the replacement character which is U+FFFD

mayaraman19: used ref https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/toWellFormed instead since this references U+FFFD and corrects invalid characters as opposed to strictly checking for them. Let me know if this is incorrect!

nbbeeken (approved):
LGTM

source/fundamentals/bson/utf8-validation.txt
```diff
@@ -25,15 +25,16 @@ processing overhead since it needs to check the data.
...
+By default, the driver enables UTF-8 validation on data from MongoDB. 
+It checks incoming documents for any characters that are not encoded in a 
+valid UTF-8 format when it parses data sent from MongoDB to your application.
 
 .. note::
 
-   The current version of the {+driver-short+} automatically substitutes
-   invalid UTF-8 characters with alternate valid UTF-8 ones before
-   validation when you send data to MongoDB. Therefore, the validation
+   The current version of the {+driver-short+} automatically substitutes invalid
```
jordan-smith721, line 34: S: Since we use the term "current" to refer to the currently active driver version, using it here could cause confusion once later versions are released. I think we can reword it so that we don't need to change it again when this version is a bit older
```suggestion
   This version of the {+driver-short+} automatically substitutes invalid
```

jordan-smith721 (approved):
One suggestion but otherwise LGTM!

PR #958, opened 2025-02-03
DOCSP-32752 Support for setting AWS region

lindseymoore:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-NNNNN>
Staging - <https://docs-mongodbcom-staging.corp.mongodb.com/drivers/docsworker-xlarge/NNNNN/>

## Self-Review Checklist

### DIFF
--- a/source/fundamentals/authentication/mechanisms.txt
+++ b/source/fundamentals/authentication/mechanisms.txt
@@ -235,6 +235,14 @@ The driver checks for your credentials in the following sources in order:
       .. literalinclude:: /code-snippets/authentication/aws-env-variable.js
          :language: javascript
 
+      .. note:: AWS Set Region
+
+         AWS recommends using Regional AWS STS endpoints instead of the global 
+         endpoint to reduce latency, build-in redundancy, and increase session token validity.
+         Set the following environment variables: AWS_STS_REGIONAL_ENDPOINTS MUST be set to either 'legacy' or 'regional', 
+         and AWS_REGION must be set.
+
+
 .. important:: Retrieval of AWS Credentials
    
    Starting in version 4.11, when you install the optional


source/fundamentals/authentication/mechanisms.txt
```diff
@@ -234,6 +234,23 @@ The driver checks for your credentials in the following sources in order:
...
+         AWS recommends using regional AWS STS endpoints instead of global 
+         endpoints to reduce latency, build-in redundancy, and increase session token validity.
+
+         To set the AWS region, set the following environment variables: 
+         
+         - ``AWS_STS_REGIONAL_ENDPOINTS``: Set to ``regional`` to enable regional endpoints. 
+   
+         - ``AWS_REGION``: Set to an AWS region, like ``us-east-2``. For a list of available AWS regions,
+           see the `AWS Regions <https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.RegionsAndAvailabilityZones.html#Concepts.RegionsAndAvailabilityZones.Regions>`__ 
+           guide in the AWS documentation.
```
rachel-mack, line 248:
```suggestion
         - ``AWS_REGION``: Set to an AWS region, such as ``us-east-2``. For a list of available AWS regions,
           see the `AWS Regions <https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.RegionsAndAvailabilityZones.html#Concepts.RegionsAndAvailabilityZones.Regions>`__ 
           guide in the AWS documentation. 
```

source/fundamentals/authentication/mechanisms.txt
```diff
@@ -234,6 +234,23 @@ The driver checks for your credentials in the following sources in order:
       .. literalinclude:: /code-snippets/authentication/aws-env-variable.js
          :language: javascript
 
+      .. note:: How to Set the AWS Region for Regional Endpoints
```
rachel-mack, line 237: I would move this to line 230 and not make it a note. I would also include a code block:
```bash
export AWS_STS_REGIONAL_ENDPOINTS=regional //enable regional endpoints
export AWS_REGION=us-east-2 //select your AWS region
```
Looking at the JIRA ticket, it looks like we should specify that the default region is us-east-1.

rachel-mack (changes requested):
I think this should be incorporated a little differently:

source/fundamentals/authentication/mechanisms.txt
lindseymoore: The restructuring makes sense, thanks! I'll ask the technical reviewer to confirm the default region and whether global endpoints are available for this feature (in which case I would delete the first sentence about why to use regional endpoints).

lindseymoore:
For the technical reviewer, can you confirm the following:
- Is ``us-east-1`` the default region if both region environment variables aren't set?
- Does this authentication only use regional endpoints? In which case, I will delete the first sentence talking about the benefits of using regional endpoints.
Thank you!

nbbeeken (approved):
the warning LGTM!

Yes us-east-1 is the default if neither are set. 

Auth does not _always_ use your region's endpoint (then us-east-1 would not be the default, your current region would be). So the first sentence should be kept because it is an improvement to enable this :) ty again

rachel-mack (approved):

PR #979, opened 2025-02-04
DOCSP-46884 Abort signal support

stephmarie17:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-46884>
Staging - https://deploy-preview-979--docs-node.netlify.app/fundamentals/crud/read-operations/cursor/

## Self-Review Checklist

### DIFF
--- a/source/code-snippets/crud/cursor.js
+++ b/source/code-snippets/crud/cursor.js
@@ -74,24 +74,19 @@ async function abort(myColl) {
   const controller = new AbortController();
   const { signal } = controller;
 
-  // Aborts operation after 20 seconds
-  setTimeout(() => controller.abort(), 20000);
-
-  const cursor = myColl.find({}, { signal });
+  process.on('SIGINT', () => controller.abort(new Error('^C pressed')));
 
   try {
-    while (await cursor.hasNext()) {
-      const doc = await cursor.next();
+    const cursor = myColl.find({}, { signal });
+    for await (const doc of cursor) {
       console.log(doc);
     }
   } catch (error) {
-    if (error.name === 'AbortError') {
-      console.error('Cursor operation was aborted');
+    if (error === signal.reason) {
+      console.error('Operation aborted:', error);
     } else {
-      console.error('An error occurred:', error);
+      console.error('Unexpected error:', error);
     }
-  } finally {
-    await cursor.close();
   }
   // end abort cursor example
 }
--- a/source/code-snippets/crud/cursor.js
+++ b/source/code-snippets/crud/cursor.js
@@ -68,6 +68,17 @@ async function close(myColl) {
   await cursor.close();
   // end close cursor example
 }
+// Abort in-progress operations
+async function abort(myColl) {
+  // start abort cursor example
+  const abort = new AbortController();
+  const { signal } = abort;
+
+  await myColl.find({}, { signal });
+
+  abort.abort();
+  // end abort cursor example
+}
 
 async function run() {
   try {
--- a/source/code-snippets/crud/cursor.js
+++ b/source/code-snippets/crud/cursor.js
@@ -71,12 +71,28 @@ async function close(myColl) {
 // Abort in-progress operations
 async function abort(myColl) {
   // start abort cursor example
-  const abort = new AbortController();
-  const { signal } = abort;
+  const controller = new AbortController();
+  const { signal } = controller;
 
-  await myColl.find({}, { signal });
+  // Aborts operation after 20 seconds
+  setTimeout(() => controller.abort(), 20000);
 
-  abort.abort();
+  const cursor = myColl.find({}, { signal });
+
+  try {
+    while (await cursor.hasNext()) {
+      const doc = await cursor.next();
+      console.log(doc);
+    }
+  } catch (error) {
+    if (error.name === 'AbortError') {
+      console.error('Cursor operation was aborted');
+    } else {
+      console.error('An error occurred:', error);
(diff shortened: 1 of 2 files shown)


source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -165,3 +165,29 @@ and the {+mdb-server+}:
    :language: javascript
    :start-after: start close cursor example
    :end-before: end close cursor example
+
+Abort
+~~~~~
+
+You can manage a cursor's resources and cancel cursor operations if they're no
+longer needed using an `abort signal <{+api+}/types/Abortable.html>`__.  You can
+pass the ``signal`` command to the following methods:
```
rachel-mack, line 174: I would rephrase this to be little more direct. The resource management is more of the "why" and I think we should lead with the "what"
```suggestion
You can cancel cursor operations and close your connection by using an `abort signal <{+api+}/types/Abortable.html>`__.  This can help you manage your resources by releasing the memory and network resources used by the cursor if they're no longer needed. 

You can pass the ``signal`` command to the following methods:
```

rachel-mack (changes requested):

source/code-snippets/crud/cursor.js
```diff
@@ -68,6 +68,17 @@ async function close(myColl) {
   await cursor.close();
   // end close cursor example
 }
+// Abort in-progress operations
+async function abort(myColl) {
+  // start abort cursor example
+  const abort = new AbortController();
+  const { signal } = abort;
```
rachel-mack, line 75: Just took a look at the JIRA ticket and wondering if this is the proper syntax?
```suggestion
  const signal = abort.signal;
```

nbbeeken: It's the same result either way but code samples almost always use the [destructuring syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment) when demonstrating abort signal use

source/code-snippets/crud/cursor.js
```diff
@@ -68,6 +68,17 @@ async function close(myColl) {
...
+
+  await myColl.find({}, { signal });
```
nbbeeken, line 77: find returns a cursor, you have to await something like toArray() or next()

stephmarie17: Updated code example based on the following comment as well 👍

source/code-snippets/crud/cursor.js
```diff
@@ -68,6 +68,17 @@ async function close(myColl) {
...
+
+  abort.abort();
```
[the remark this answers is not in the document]
stephmarie17: Thanks for that callout @nbbeeken - I updated the code example to try and think of a simple example that is a different scenario than the one in the API docs just to avoid repeating that content. LMK if that looks good to you! I tested it (although in testing changed the timeout to be very short so it would fire the signal) and it worked as expected.

nbbeeken (changes requested):

source/code-snippets/crud/cursor.js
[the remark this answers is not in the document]
nbbeeken: Glad it worked! see my other comment, unfortunately, timers are the easiest example but precisely not the one we want to recommend 😅

source/code-snippets/crud/cursor.js
```diff
@@ -68,6 +68,33 @@ async function close(myColl) {
...
   // end close cursor example
 }
+// Abort in-progress operations
+async function abort(myColl) {
+  // start abort cursor example
+  const controller = new AbortController();
+  const { signal } = controller;
+
+  // Aborts operation after 20 seconds
+  setTimeout(() => controller.abort(), 20000);
```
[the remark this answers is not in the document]
nbbeeken: It's hard to have an easy one without more boilerplate. If there was an http server then it might be tied to the http request closing:

```js
server.on('/endpoint', (req, res) => {
  req.on('close', () => controller.abort());
})
``` 

but that's a lot to sort of approximately get correct that it may not translate when trying to employ it (like what is `server` if we don't show it being created etc.)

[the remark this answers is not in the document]
stephmarie17: I took a look at what an example like this would require and yeah I think it might not translate super well without all of the boilerplate for context. For now, I repurposed example from the API docs to use a cursor. It seems like a sufficient example to copy and paste for testing out the feature, lmk what you think and if any other technical updates are needed for the PR. Ty for your help @nbbeeken

nbbeeken (approved):

source/fundamentals/crud/read-operations/cursor.txt
```diff
@@ -165,3 +165,36 @@ and the {+mdb-server+}:
...
+
+.. literalinclude:: /code-snippets/crud/cursor.js
+   :language: javascript
+   :start-after: start abort cursor example
+   :end-before: end abort cursor example
+
+.. note::
+   
+   This feature is experimental. Aborting a signal closes a connection, which 
+   might cause unnecessary connection reestablishment.
```
rachel-mack, line 200: I would move this note up to line 176.

rachel-mack (approved):
One suggestion:

source/fundamentals/crud/read-operations/cursor.txt
stephmarie17: Was thinking of moving this higher, will update!

PR #1069, opened 2025-04-04
DOCSP-48390 Document Data Formats: EJSON

lindseymoore:
# Pull Request Info

JIRA - <https://jira.mongodb.org/browse/DOCSP-NNNNN>

## Self-Review Checklist

- [ ] Are the page titles greater than 20 characters long and [SEO relevant](https://docs.google.com/spreadsheets/d/1Wkt0-5z04KmcMNscN5bjUKnzwWAtMq9VESp-Lz6r2o8/edit?usp=sharing)?

### DIFF
--- a/source/data-formats/extended-json.txt
+++ b/source/data-formats/extended-json.txt
@@ -1,0 +1,206 @@
+. _node-extended-json:
+
+=============
+Extended JSON
+=============
+
+.. contents:: On this page
+   :local:
+   :backlinks: none
+   :depth: 2
+   :class: singlecol
+
+.. facet::
+   :name: genre
+   :values: reference
+ 
+.. meta::
+   :keywords: code examples, bson, relaxed, canonical, legacy
+
+Overview
+--------
+
+JSON is a data format that represents the values of objects, arrays, numbers,
+strings, booleans, and nulls. The **Extended JSON** format defines a reserved
+set of keys prefixed with "``$``" to represent field type information that
+directly corresponds to each type in BSON, the format that MongoDB uses to
+store data.
+
+Extended JSON Formats
+---------------------
+
+MongoDB Extended JSON features different string formats to represent BSON data.
+Each of the different formats conform to the JSON RFC
+and meet specific use cases. The **extended** format, also known as the
+**canonical** format, features specific representations for every BSON type
+for bidirectional conversion without loss of information. The **Relaxed mode**
+format is more concise and closer to ordinary JSON, but does not represent
+all the type information such as the specific byte size of number fields.
+
+See the following table to see a description of each format:
+
+.. list-table::
+   :header-rows: 1
+   :stub-columns: 1
+   :widths: 10 40
+
+   * - Name
+     - Description
+
+   * - **Extended**
+     - | Also known as the *canonical* format, this JSON representation avoids loss of
+         BSON type information.
+       | This format prioritizes type preservation at the loss of human-readability and
+         interoperability with older formats.
+
+   * - **Relaxed Mode**
+     - | JSON representation that describes BSON documents with some type information loss.
+       | This format prioritizes human-readability and interoperability at the loss of
+         certain type information.
+
+.. note::
+
+   The driver parses the ``$uuid`` Extended JSON type from a string to a
+   ``BsonBinary`` object of binary subtype 4. For more information about ``$uuid`` field
+   parsing, see the
+   :spec:`special rules for parsing $uuid fields </extended-json.rst#special-rules-for-parsing-uuid-fields>`
+   section in the extended JSON specification.
+
+To learn more about JSON, BSON, and Extended JSON, see
+`our article about JSON and BSON <https://www.mongodb.com/resources/basics/json-and-bson>`__
+and :manual:`Extended JSON </reference/mongodb-extended-json/>` in the {+mdb-server+} manual.
+
+.. _extended_json_example_section:
+
+Extended JSON Examples
+~~~~~~~~~~~~~~~~~~~~~~
+
+The following examples show a document containing an ObjectId, date, and long
+number field represented in each Extended JSON format. Click the tab that
+corresponds to the format of the example you want to see:
+
+.. tabs::
+
+   .. tab:: Extended
+      :tabid: extended-format
+
+      .. code-block:: json
+
+         {
+           "_id": { "$oid": "573a1391f29313caabcd9637" },
+           "createdAt": { "$date": { "$numberLong": "1601499609" }},
+           "numViews": { "$numberLong": "36520312" }
+         }
+
+   .. tab:: Relaxed Mode
+      :tabid: relaxed-mode-format
+
+      .. code-block:: json
+
+         {
+           "_id": { "$oid": "573a1391f29313caabcd9637" },
+           "createdAt": { "$date": "2020-09-30T18:22:51.648Z" },
+           "numViews": 36520312
+         }
+
+Read Extended JSON
+------------------
+
+You can read an Extended JSON string into the JavaScript value or object described 
+by the string by using the ``EJSON.parse()`` method. 
+
+The following example shows how you can read an Extended JSON string into a
+JavaScript value or object by using the ``parse()`` method:
+
+.. io-code-block::
+
+   .. input::
+      :language: javascript
+
+      import { BSON } from 'mongodb';
+      const EJSON = BSON.EJSON;
+
+      const ejson_str = '[{"foo":[1,2]},{"bar":{"hello":"world"}},{"code":{"$scope":{},"$code":"function x() { return 1; }"}}]'
+
+      const doc = EJSON.parse(ejson_str);
+      console.log(doc)
+
+   .. output::
+      :language: none
+      :visible: false
+
+      [
+         {'foo': [1, 2]},
+         {'bar': {'hello': 'world'}},
+         {'code': Code('function x() { return 1; }', {})}
+      ]
+
+Write Extended JSON
+-------------------
+
+You can write an Extended JSON string from a BSON document object by using the
+``EJSON.stringify()`` method.
+
+The following example outputs an Extended JSON string in the Relaxed format:
+
+.. io-code-block::
+
+   .. input::
+      :language: js
+
+      const { EJSON } = require('bson');
+
+      doc = [
+         {'foo': [1, 2]},
+         {'bar': {'hello': 'world'}},
+         {'code': Code('function x() { return 1; }', {})}
+      ]
+
+      const ejson_str = EJSON.stringify(doc);
+      console.log(ejson_str);
+
+   .. output::
+      :language: none
+      :visible: false
+
+      [{"foo":[1,2]},{"bar":{"hello":"world"}},{"code":{"$scope":{},"$code":"function x() { return 1; }"}}]
+
+By default, the ``stringify()`` method returns the Extended JSON string in the 
+Relaxed format. To specify the Canonical format, set the ``relaxed`` 
+option to ``false``.
+
+The following example shows how to output Extended JSON in the Canonical format:
+
+.. io-code-block::
+
+   .. input::
+      :language: js
+
+      const { EJSON } = require('bson');
+
+      doc = [
+         {'foo': [1, 2]},
+         {'bar': {'hello': 'world'}},
+         {'code': Code('function x() { return 1; }', {})}
+      ]
+
+      const ejson_str = EJSON.stringify(doc, { relaxed: false });
+      print(ejson_str)
+
+   .. output::
+      :language: none
+      :visible: false
+
+      [{"foo":[{"$numberInt":"1"},{"$numberInt":"2"}]},{"bar":{"hello":"world"}},{"code":{"$scope":{},"$code":"function x() { return 1; }"}}]
+
+Additional Information
+----------------------
+
+The resources in the following sections provide more information about working
+with Extended JSON.
+
+API Documentation
+~~~~~~~~~~~~~~~~~
+
+To learn more about any of the methods or types discussed in this
+guide, see the `EJSON <https://github.com/mongodb/js-bson#ejson>`__ API documentation.
--- a/source/data-formats/bson.txt
+++ b/source/data-formats/bson.txt
@@ -138,4 +138,3 @@ API Documentation
 
 To learn more about any of the methods or types discussed in this
 guide, see the `BSON <{+api+}/modules/BSON.html>`__ API documentation.
-


source/data-formats/extended-json.txt
```diff
@@ -0,0 +1,203 @@
...
+store data.
+
+Extended JSON Formats
+---------------------
+
+MongoDB Extended JSON features different string formats to represent BSON data.
+Each of the different formats conform to the JSON RFC
+and meet specific use cases. The **Extended** format, also known as the
+**Canonical** format, features specific representations for every BSON type
+for bidirectional conversion without loss of information. The **Relaxed Mode**
```
stephmarie17, line 36: [q] Do you know if the different format names are meant to be capitalized? I think it makes sense to do so, but I've seen a mix in different docs.

stephmarie17 (approved):
LGTM with a question!

source/data-formats/extended-json.txt
lindseymoore: It seems to be capitalized in the Server Docs, so I tried to be consistent: https://www.mongodb.com/docs/manual/reference/mongodb-extended-json/

source/data-formats/extended-json.txt
```diff
@@ -0,0 +1,203 @@
...
+format is more concise and closer to ordinary JSON, but does not represent
+all the type information such as the specific byte size of number fields.
```
nbbeeken, line 38:
```suggestion
all the type information such as the specific bit width of numeric fields.
```

Because I believe what you are referencing is $numberLong, $numberInt being represented as plain numbers in relaxed mode. Size and number magnitude are easy to confuse so the phrase bit width is used to be specific about the amount of precision

source/data-formats/extended-json.txt
```diff
@@ -0,0 +1,203 @@
...
+      :language: js
+
+      const { Binary, Code, EJSON } = require('bson');
+
+      doc = [
+         {"foo": [1, 2]},
+         {"bar": {"hello": "world"}},
+         {"code": new Code("function x() { return 1; }", {})},
+         {"bin": Binary.createFromBase64("AQIDBA==", 0)}
+      ]
```
nbbeeken, line 118:
```js
      const doc = [
        { foo: [1, 2] },
        { bar: { hello: "world" } },
        { code: new Code("function x() { return 1; }", {}) },
        { bin: Binary.createFromBase64("AQIDBA==", 0) },
      ];
```

Just making this look more JavaScript-y, there's nothing wrong with quoted key syntax but most programmers won't take the time to do it unless there's a key with a symbol in it that can't be written into code literally (like keys with dots or dashes).

source/data-formats/extended-json.txt
```diff
@@ -0,0 +1,203 @@
...
+``EJSON.stringify()`` method.
+
+The following example outputs an Extended JSON string in the Relaxed format:
+
+.. io-code-block::
+
+   .. input::
+      :language: js
+
+      const { Binary, Code, EJSON } = require('bson');
```
nbbeeken, line 111:
```suggestion
      import { Binary, Code, EJSON } from 'bson';
```

Modernizing 🙂

source/data-formats/extended-json.txt
```diff
@@ -0,0 +1,203 @@
...
+
+      doc = [
+         {"foo": [1, 2]},
+         {"bar": {"hello": "world"}},
+         {"code": new Code("function x() { return 1; }", {})},
+         {"bin": Binary.createFromBase64("AQIDBA==", 0)}
+      ]
+
+      const ejson_str = EJSON.stringify(doc);
+      console.log(ejson_str);
```
nbbeeken, line 121:
```suggestion
      const ejsonStr = EJSON.stringify(doc);
      console.log(ejsonStr);
```

camel is the general JS convention (not wrong, just sticks out)

source/data-formats/extended-json.txt
```diff
@@ -0,0 +1,203 @@
...
+   .. input::
+      :language: js
+
+      const { Binary, Code, EJSON } = require('bson');
+
+      doc = [
+         {"foo": [1, 2]},
+         {"bar": {"hello": "world"}},
+         {"code": new Code("function x() { return 1; }", {})},
+         {"bin": Binary.createFromBase64("AQIDBA==", 0)}
```
nbbeeken, line 146: Same mods as above. 

Maybe we should replace the binary example with a Date one? we have the "$numberInt"s showing up here, but the difference between Date relaxed true/false is more significant, it demonstrates how one is readable and the other is intended more for transmission/storage.

source/data-formats/extended-json.txt
```diff
@@ -0,0 +1,203 @@
...
+      ]
+
+      const ejson_str = EJSON.stringify(doc, { relaxed: false });
+      print(ejson_str)
+
+   .. output::
+      :language: none
+      :visible: false
+
+      [{"foo":[{"$numberInt":"1"},{"$numberInt":"2"}]},{"bar":{"hello":"world"}},{"code":{"$scope":{},"$code":"function x() { return 1; }"}},{"bin":{"$binary":{"base64":"AQIDBA==","subType":"00"}}}]
```
nbbeeken, line 156: Should we have a more formatted result here (and above)?

```js
EJSON.stringify(doc, undefined, 2, {relaxed: false})
```
The "2" here will indent everything

source/data-formats/extended-json.txt
```diff
@@ -0,0 +1,203 @@
...
+The following example shows how you can read an Extended JSON string into a
+JavaScript value or object by using the ``parse()`` method:
+
+.. io-code-block::
+
+   .. input::
+      :language: javascript
+
+      import { BSON } from 'mongodb';
+      const EJSON = BSON.EJSON;
```
nbbeeken, line 173:
```suggestion
      import { EJSON } from 'bson';
```

To be consistent with the examples above. OR change the examples above to import from the driver (then this code snippet is correct as is)

source/data-formats/extended-json.txt
```diff
@@ -0,0 +1,203 @@
...
+
+      const ejson_str = '[{"foo":[1,2]},{"bar":{"hello":"world"}},{"code":{"$code":"function x() { return 1; }","$scope":{}}},{"bin":{"$binary":{"base64":"AQIDBA==","subType":"00"}}}]'
+
+      const doc = EJSON.parse(ejson_str);
```
nbbeeken, line 177:
```suggestion
      const ejsonStr = '[{"foo":[1,2]},{"bar":{"hello":"world"}},{"code":{"$code":"function x() { return 1; }","$scope":{}}},{"bin":{"$binary":{"base64":"AQIDBA==","subType":"00"}}}]'

      const doc = EJSON.parse(ejsonStr);
```

source/data-formats/extended-json.txt
nbbeeken:
```suggestion
      const doc = {
        foo: [1, 2],
        bar: { hello: "world" },
        code: new Code("function x() { return 1; }", {}),
        bin: Binary.createFromBase64("AQIDBA==", 0),
    };
```
_Edit:_

Node's EJSON stringifier supports arrays as a top-level object to have parity with the native `JSON.stringify`. In BSON however, and for other drivers, there is no such thing as a top-level array, we should stick to document examples as the top-level argument.

source/data-formats/extended-json.txt
nbbeeken: Optional, if you do update the output to be formatted and want to also do that here (not necessary) you can put multiline strings into JS with a backtick quote:

```
const ejsonStr = `
{
  "foo": [
    { "$numberInt": "1" },
    { "$numberInt": "2" }
  ],
  "bar": { "hello": "world" },
  "code": {
    "$code": "function x() { return 1; }",
    "$scope": {}
  },
  "bin": { "$binary": { "base64": "AQIDBA==", "subType": "00" } }
}
`;
```

nbbeeken (changes requested):

source/data-formats/extended-json.txt
nbbeeken: Ah I just caught the other page where we say we recommend importing from the driver. I agree with that, so let's switch the BSON imports above to be driver ones.

lindseymoore: Will use the driver imports!

source/data-formats/extended-json.txt
lindseymoore: Replaced with a Date object. Let me know what you think!

source/data-formats/extended-json.txt
lindseymoore: `[
  {
    "foo": [
      1,
      2
    ]
  },
  {
    "bar": {
      "hello": "world"
    }
  },
  {
    "code": {
      "$code": "function x() { return 1; }",
      "$scope": {}
    }
  },
  {
    "date": {
      "$date": "2024-07-20T14:30:00Z"
    }
  }
]`
This formatting looks a bit off to me. Fine with leaving as is!

source/data-formats/extended-json.txt
lindseymoore: Unfortunately, the back tick messes with the the Markdown formatting in the file. Will have to leave as is.

lindseymoore: Should the brackets surrounding the string be removed?

nbbeeken: Seems to render correctly:

image


but we still have a few small issue, are you referring to the square brackets in the output? because yes, it still shows the result being an array. Inside the string in code, no, the outer most braces are JSON syntax, you need those.

source/data-formats/extended-json.txt
```diff
@@ -0,0 +1,215 @@
...
+
+.. io-code-block::
+
+   .. input::
+      :language: javascript
+
+      import { BSON } from 'mongodb';
+      const EJSON = BSON.EJSON;
+
+      const ejsonStr = const ejsonStr = `{
```
nbbeeken, line 176:
```suggestion
      const ejsonStr = `{
```

source/data-formats/extended-json.txt
```diff
@@ -0,0 +1,215 @@
...
+   .. output::
+      :language: none
+      :visible: false
+
+      [
+        { foo: [ 1, 2 ] },
+        { bar: { hello: 'world' } },
+        { code: new Code('function x() { return 1; }', {}) },
+        { bin: Binary.createFromBase64('AQIDBA==', 0) }
+      ]
```
nbbeeken, line 201:
```suggestion
      {
         foo: [1, 2],
         bar: { hello: "world" },
         code: new Code("function x() { return 1; }", {}),
         bin: Binary.createFromBase64('AQIDBA==', 0),
      }
```

source/data-formats/extended-json.txt
```diff
@@ -0,0 +1,215 @@
...
+      :language: js
+
+      const { Binary, Code, EJSON } = require('bson');
+
+      const doc = {
+         { foo: [1, 2] },
+         { bar: { hello: "world" } },
+         { code: new Code("function x() { return 1; }", {}) },
+         { date: new Date(2024, 6, 20, 10, 30, 0) },
+      };
```
nbbeeken, line 148: This is not valid JS syntax, we've switched to an object here which means the keys need to be defined inside the outermost curly braces here.
```suggestion
      const doc = {
         foo: [1, 2],
         bar: { hello: "world" },
         code: new Code("function x() { return 1; }", {}),
         date: new Date(2024, 6, 20, 10, 30, 0),
      };
```

source/data-formats/extended-json.txt
```diff
@@ -0,0 +1,215 @@
...
+
+      import { Binary, Code, BSON } from 'mongodb';
+      const EJSON = BSON.EJSON;
+
+      const doc = {
+         { foo: [1, 2] },
+         { bar: { hello: "world" } },
+         { code: new Code("function x() { return 1; }", {}) },
+         { date: new Date(2024, 6, 20, 10, 30, 0) },
+      };
```
nbbeeken, line 119:
```suggestion
      const doc = {
         foo: [1, 2],
         bar: { hello: "world" },
         code: new Code("function x() { return 1; }", {}),
         date: new Date(2024, 6, 20, 10, 30, 0),
      };
```

source/data-formats/extended-json.txt
lindseymoore: Ahh thanks for catching this, forgot to push my last commit!

source/data-formats/extended-json.txt
lindseymoore: The output actually does look like the original here. Is it ok to keep as is?

source/data-formats/extended-json.txt
```diff
@@ -0,0 +1,219 @@
...
+
+      const ejsonStr = EJSON.stringify(doc);
+      console.log(ejsonStr);
+
+   .. output::
+      :language: none
+      :visible: false
+
+      [{"foo":[1,2]},{"bar":{"hello":"world"}},{"code":{"$code":"function x() { return 1; }","$scope":{}}},{"date":{"$date":"2024-07-20T14:30:00Z"}}]
```
nbbeeken, line 128:
```suggestion
      {"foo":[1,2],"bar":{"hello":"world"},"code":{"$code":"function x() { return 1; }","$scope":{}},"date":{"$date":"2024-07-20T14:30:00Z"}}
```

This output is still an array

source/data-formats/extended-json.txt
```diff
@@ -0,0 +1,219 @@
...
+option to ``false``.
+
+The following example shows how to output Extended JSON in the Canonical format:
+
+.. io-code-block::
+
+   .. input::
+      :language: js
+
+      const { Binary, Code, EJSON } = require('bson');
```
nbbeeken, line 141: I thought we wanted to use the driver import consistently?

source/data-formats/extended-json.txt
```diff
@@ -0,0 +1,219 @@
...
+      };
+
+      const ejsonStr = EJSON.stringify(doc, { relaxed: false });
+      print(ejsonStr)
+
+   .. output::
+      :language: none
+      :visible: false
+
+      [{"foo":[{"$numberInt":"1"},{"$numberInt":"2"}]},{"bar":{"hello":"world"}},{"code":{"$code":"function x() { return 1; }","$scope":{}}},{"date":{"$date":{"$numberLong":"1721485800000"}}}]
```
nbbeeken, line 157:
```suggestion
      {"foo":[{"$numberInt":"1"},{"$numberInt":"2"}],"bar":{"hello":"world"},"code":{"$code":"function x() { return 1; }","$scope":{}},"date":{"$date":{"$numberLong":"1721485800000"}}}
```

Output is an array but input is an object

source/data-formats/extended-json.txt
```diff
@@ -0,0 +1,219 @@
...
+            { "$numberInt": "1" },
+            { "$numberInt": "2" }
+         ],
+         "bar": { "hello": "world" },
+         "code": {
+            "$code": "function x() { return 1; }",
+            "$scope": {}
+         },
+         "bin": { "$binary": { "base64": "AQIDBA==", "subType": "00" } }
+         }`;
```
nbbeeken, line 190:
```suggestion
       }`;
```

Doesn't functionally make a difference but is the proper indentation level for this closing brace

source/data-formats/extended-json.txt
nbbeeken: I updated the change ^ I didn't mean to change the quotes, but the braces mean object, whereas brackets mean array. We switched the input to an object, so this should change too

nbbeeken (changes requested):

source/data-formats/extended-json.txt
lindseymoore: switched to driver import, thanks!

nbbeeken (approved):

